PR #11, opened 2023-01-21
WIP: Single node execution support

glitch003:


packages/lit-node-client/src/lib/lit-node-client.ts
```diff
@@ -80,15 +80,18 @@ import {
...
 import { computeAddress } from '@ethersproject/transactions';
 import { SiweMessage } from 'lit-siwe';
-import { joinSignature } from 'ethers/lib/utils';
+import { joinSignature, sha256 } from 'ethers/lib/utils';
 import {
   checkAndSignAuthMessage,
   getSessionKeyUri,
   parseResource,
 } from '@lit-protocol/auth-browser';
+import { importer } from 'ipfs-unixfs-importer';
```
glitch003, line 89: @Ansonhkg this seems to be producing an error when i try to run the tests.  Can you run `yarn test:packages` and see if the error is something you're familiar with?

packages/lit-node-client/src/lib/lit-node-client.ts
```diff
@@ -234,6 +237,7 @@ export class LitNodeClient {
       authSig: params.authSig,
       jsParams: convertLitActionsParams(params.jsParams),
       requestId: Math.random().toString(16).slice(2),
+      singleNode: params.singleNode ?? false,
```
glitch003, line 240: not sure if this is the right place to put the default of false?

packages/lit-node-client/src/lib/lit-node-client.ts
```diff
@@ -1027,6 +1031,87 @@ export class LitNodeClient {
...
+    });
+    reqBody.authSig = sigToPassToNode;
+
+    let executionResponse = await this.getJsExecutionShares(url, {
+      ...reqBody,
+    });
+
+    const successPromises: SuccessNodePromises = {
+      success: true,
+      values: [executionResponse as any],
```
glitch003, line 1109: I added "as any" here to make this compile but i think it's wrong.  I think `values` should have type NodeShare[] but it wasn't clear to me how you transform the NodeCommandResponse into the NodeShare type inside `handleNodePromises`.

packages/lit-node-client/src/lib/lit-node-client.ts
```diff
@@ -1344,28 +1430,33 @@ export class LitNodeClient {
...
-    const nodePromises = this.getNodePromises((url: string) => {
-      // -- choose the right signature
-      let sigToPassToNode = this.getAuthSigOrSessionAuthSig({
-        authSig,
-        sessionSigs,
-        url,
-      });
-      reqBody.authSig = sigToPassToNode;
+    let res;
+    if (singleNode) {
```
glitch003, line 1434: feel free to restructure this.  but i figured both single and multi execution can return an array of response promises so it makes sense to put the if statement here

packages/lit-node-client/src/lib/lit-node-client.ts
glitch003: thinking about this more...  i think singleNode and a boolean is wrong.  i think instead we should let them specify how many nodes it should run on.  there are def cases where someone might wanna run something on just 3 nodes instead of 1

packages/lit-node-client/src/lib/lit-node-client.ts
Ansonhkg: It seems like Jest isn't transforming those modules for us, but the libraries we are using ships a minified version of their code, so we can just use the in the jest tests and skip transforming all together.

`jest.config.ts`

```
  moduleNameMapper: {
    '^ipfs-unixfs-importer':
      'node_modules/ipfs-unixfs-importer/dist/index.min.js',
    '^blockstore-core': 'node_modules/blockstore-core/dist/index.min.js',
  },
```

packages/lit-node-client/src/lib/lit-node-client.ts
Ansonhkg: I've removed `singleNode:boolean` and added `targerNodeRange?: number` instead:
```
targetNodeRange: params.targetNodeRange ?? 0,
```

then we can do something like:
```js
  // -- only run on a single node
  if (targetNodeRange && targetNodeRange === 1) {
    res = await this.runOnSingleNode(params);
  
  // -- run on multiple nodes
  } else if (targetNodeRange && targetNodeRange > 1) {
    throw new Error('this.runOnMultipleNodes(params) not implemented yet');
  
  // -- run on all nodes (default)
  } else {
    ```

**Comment by Ansonhkg** [comment]
test passed but we are now having some polyfill issues on react:

```
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "crypto": require.resolve("crypto-browserify") }'
        - install 'crypto-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "crypto": false }

ERROR in ../../dist/packages/lit-node-client/src/lib/lit-node-client.js 20:31-62
Module not found: Error: Package path . is not exported from package /Users/user/Projects/js-sdk/node_modules/ipfs-unixfs-importer (see exports field in /Users/user/Projects/js-sdk/node_modules/ipfs-unixfs-importer/package.json)

ERROR in ../../dist/packages/lit-node-client/src/lib/lit-node-client.js 21:17-50
Module not found: Error: Package path ./memory is not exported from package /Users/user/Projects/js-sdk/node_modules/blockstore-core (see exports field in /Users/user/Projects/js-sdk/node_modules/blockstore-core/package.json)

webpack compiled with 2 errors and 7 warnings (8f6057cb2f310f04)
No errors found.

```

File: packages/lit-node-client/src/lib/lit-node-client.ts (hunk 2 above)
Comment (glitch003) [reply packages/lit-node-client/src/lib/lit-node-client.ts]: 🔥

File: packages/lit-node-client/src/lib/lit-node-client.ts (hunk 1 above)
Comment (glitch003) [reply packages/lit-node-client/src/lib/lit-node-client.ts:89]: nice find

**Comment by glitch003** [comment]
is the crypto polyfill needed because of this?  https://github.com/LIT-Protocol/js-sdk/blob/feature/single-execution/packages/lit-node-client/src/lib/lit-node-client.ts#L1083

hmm.  we could use the IPFS CID directly, instead of hashing it with sha256.  we would just need to decode it and turn it into a number.

**Comment by Ansonhkg** [comment]
> is the crypto polyfill needed because of this? https://github.com/LIT-Protocol/js-sdk/blob/feature/single-execution/packages/lit-node-client/src/lib/lit-node-client.ts#L1083
> 
> hmm. we could use the IPFS CID directly, instead of hashing it with sha256. we would just need to decode it and turn it into a number.

It's because of `ipfs-unixfs-importer` and `blockstore-core`, had to add the paths to `tsconfig.base.json` as well:

```json
      "ipfs-unixfs-importer": [
        "node_modules/ipfs-unixfs-importer/dist/index.min.js"
      ],
      "blockstore-core": ["node_modules/blockstore-core/dist/index.min.js"]
```

Ansonhkg:
Commit [2nd pass: works on react but not nodejs](https://github.com/LIT-Protocol/js-sdk/pull/11/commits/42e59df971f674f205170188a6557df92c0a41dc) now works with targeted nodes

you can try on:
```
yarn build:target lit-node-client && yarn nx run react:serve
```

and visit:
```
http://localhost:4003/lit-node-client-page
```

but not tests failed cus doesn't work on nodeks :/, going to fix that now

Ansonhkg:
Got it sorted! for some reasons the structure of node_modules has been changed (not sure if it's just me), the @walletconnect directory was initially being referenced in `node_modules/@lit-protocol/lit-node-client/node_modules/@wallet_connect`, but now it's been flatten and is pointing to `node_modules/@wallet_connect` instead. The one in the root folder "i think" is from other dependencies is using version `1.7.8` which I haven't figured out which yet, and it has overwritten our `2.1.1` version, the solution is to match the version to `1.7.8`. Long term I think we should extract `@wallet_connect` out and let devs to install their own callback instead for `checkAndSignAuthMessage` , or they can install our `authBrowser` and add to our litNodeClient constructor

PR #12, opened 2023-01-23
Update package build process to setup local development support

hwrdtm:
# What

In order to allow for developers to test against local changes made to `packages/` , we need the following:

- `dist/` directory inside each package with typescript typings and JS implementations
- each package's `package.json` having `main` and `typings` fields populated to reference files in the `dist/` directory above.

This PR represents the minimal change that is needed to enable this while continuing to work with Nx (tl;DR: Nx >> Lerna). Specifically, this PR:

- Introduces a new npm script `build:setupLocalDev` that can be run like `yarn build:setupLocalDev constants` **only after a package's typescript code has been compiled**. This npm script is opt-in **by default** for all developers. This npm script achieves the above requirements by creating a symlink in each package to the root `dist/package/<packageName>` directory, and updating each package's `package.json` accordingly. 
  - ⚠️ Note that this last part is the drawback of this PR's approach - since the npm script is automatically opted-in for all developers, running `yarn build:packages` will introduce unstaged changes. Happy to discuss improvements to this point.
- Update README to help instruct developers how to benefit from this.

# Alternative Solutions

## ❌ (Going Back To) Lerna

While I don't think turning Nx off (`useNx: false`) in order to use Lerna to build the packages is the direction we should be heading (since Nwrl / Nx has taken over Lerna and Nx is like Lerna but supercharged), here is the branch where I used Lerna to build each package into their own dist so that our local client applications can use the local changes that we make to any of the packages in the packages dir: https://github.com/LIT-Protocol/js-sdk/tree/hwrdtm/test-local-dev. As an example, I made changes to `apps/demo-encrypt-decrypt-react/` to show how you might import the package dependency locally and test with updates.

## ❌ Root `dist/` To Package `dist/`

I then looked into how we can work with Nx to output `dist/` artifacts inside each package (ie. `packages/constants/dist/`) **instead** of placing it in the root `dist/` but decided against it because a simple search of `dist` / `dist/packages` shows way too many callsites, which is trouble to tackle in a short amount of time. Hence, I landed on this PR as the minimal amount of changes needed to enable this functionality while still allow us to continue reaping benefits with the Nx tooling.

# Testing

1. Run `yarn build:packages`
2. `npm link` inside `packages/lit-node-client`
3. `npm link @lit-protocol/lit-node-client --force` inside client application.

# Misc

cc @Adarsh-Kumar28 @glitch003


glitch003:
@Ansonhkg can we merge this?  cool with you?

PR #13, opened 2023-01-25
Trying to remove extra logging

glitch003:
So I removed almost all the `console.log`'s and replaced them with `log` but there is still a problem.  

These 3 lines get logged even though debug is set to false in the LitNodeClient config, because the config hasn't been loaded yet, I think.  

```[Lit-JS-SDK] Oh hey you are running in Node.js!
[Lit-JS-SDK] ✅ [ECDSA SDK] wasmECDSA loaded. 2 functions available. Run 'wasmECDSA' in the console to see them.
[Lit-JS-SDK] ✅ [BLS SDK] wasmExports loaded. 310 functions available. Run 'wasmExports' in the console to see them.
```

two options i can think of...  write to an array / cache and then print after config is loaded.  the other option is just try to load the config first before anything else.

@Ansonhkg can you fix this?

### DIFF
--- a/packages/encryption/src/lib/params-validators.ts
+++ b/packages/encryption/src/lib/params-validators.ts
@@ -136,7 +136,7 @@ export const paramsValidators = {
     )
       return false;
 
-    console.log('authSig:', authSig);
+    log('authSig:', authSig);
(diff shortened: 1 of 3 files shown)


glitch003:
i'm taking a crack at this now

PR #17, opened 2023-01-27
fix: custom script to npm publish version or tag

Ansonhkg:
## Problem

as it stands right now, we build (with the old version) then update to the new version, which doesn't get published. 

## Solution

remove the `lerna publish` function and use a custom script to deploy based on the `lerna.json` version instead. This will both update the individual packages and also `dist/packages/*`. Note that this will still build the packages first because we don't want to bump the version every time we test local build

Updated scripts:

```json
    "publish:packages": "yarn node ./tools/scripts/pub.mjs --prod",
    "publish:dev": "yarn node ./tools/scripts/pub.mjs --tag dev",
    "publish:test": "yarn node ./tools/scripts/pub.mjs --tag test",
```


tools/scripts/pub.mjs
```diff
@@ -10,30 +10,149 @@ import {
...
+      try{
+        const pkg2 = await readJsonFile(
+          `${dir.replace('dist/', '')}/package.json`
+        );
+        pkg2.version = lernaVersion;
+
+        // write the package.json file
+        await writeJsonFile(`${dir.replace('dist/', '')}/package.json`, pkg2);
+      }catch(e){
+        // swallow
```
glitch003, line 109: is there an error you expect to get often here but don't wanna log?  

usually i will try and handle / swallow that specific case and then log or throw if there's any other error.  

because otherwise this can be a source of annoying bugs where, for example, writing the json file fails but you never know about it.

Ansonhkg: Yea, basically the `dirs` array also includes the udm packages eg. 

```
- @lit-protocol/constants
- @lit-protocol/constants-vanilla
```

but `@lit-protocol/constants-vanilla` is not a valid package in the `./packages/*` directory so it will also fail but that's not a problem.

I will add a guard that if the package includes the `vanilla` keyword it will swallow, else log the error, i think that would be better

tools/scripts/pub.mjs
```diff
@@ -97,16 +98,23 @@ await question('Are you sure you want to publish to? (y/n)', {
...
         // write the package.json file
         await writeJsonFile(`${dir.replace('dist/', '')}/package.json`, pkg2);
-      }catch(e){
-        // swallow
+      } catch (e) {
+
+        const path = `${dir.replace('dist/', '')}/package.json`;
+
+        // swallow error if it's not a vanilla package
+        if(!path.includes('vanilla')){
```
Ansonhkg, line 114: I will add a guard that if the package includes the `vanilla` keyword it will swallow, else log the error, i think that would be better

glitch003: perfect

glitch003 (approved):

PR #18, opened 2023-01-31
Make session signatures compatible with SIWE ReCap

hwrdtm:
# What

- This PR makes the `getSessionSigs` function of `@lit-protocol/lit-node-client` compatible with [EIP-5573 SIWE ReCap](https://ethereum-magicians.org/t/eip-5573-siwe-recap/10627).
- Also, the specific logic to do with how the capability objects are represented are abstracted away into a new package, `@lit-protocol/auth` so that the concerns are better separated from `@lit-protocol/lit-node-client` and are more localized. We know for sure that the ReCap standard will have at least one upcoming change as they become more compatible with UCANs.
- Refactored `GetSessionSigProps` to be in `@lit-protocol/lit-node-client` to avoid cyclical dependency tree and also better separation of concerns as, that belongs to the function call specific to the `LitNodeClient` anyways.

# Testing

Unit tests are implemented.

- `yarn test:packages`


glitch003:
@hwrdtm this looks good.  but is this live on the node yet?  we have to handle it there too, right?  or is there a PR?

hwrdtm:
Closing in favor of #88

PR #26, opened 2023-02-21
feat: Update Lit Connect Modal to include Coinbase wallet

Adarsh-Kumar28:
# What

Updated the lit-connect-modal versions in the root package.json to include Coinbase Wallet in the modal.

# Why

A team requested this feature.

# How

Simple package & yarn.lock update

# Note

Haven't updated the apps. They still have the old versions of the Modal

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "g": "^2.0.1",
     "ipfs-unixfs-importer": "^12.0.0",
     "jszip": "^3.10.1",
-    "lit-connect-modal": "0.1.8",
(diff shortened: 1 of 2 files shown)


jasonsprouse:
The wallet loads in applications if metamask isn't installed or activated and the coinbase wallet is. I think the additional sort functions may be throwing off the babel-loader and don't think it's necessary to be explicitly defined as it is here.

https://github.com/LIT-Protocol/lit-connect-modal/blob/master/src/helpers/walletList.js#L22-L35

PR #28, opened 2023-02-23
Example PR showing 20 files changed when I only changed 2

glitch003:
This PR is just an example to show how we need to change the dev workflow.  

i made a single line change in two files and github is showing 9,700 lines changed.

is it possible to build the packages in a "dev" mode that doesn't update the readme, doesn't bump the versions, etc?
i think the dev flow can look similar to how i've done it with the js-sdk where the user can make changes etc, and build the packages to test them, but the things that get built are .gitignored (you have this already).  and you can build as many times as you want and nothing changes in git.

and then after the PR is approved or you're satisfied with your changes or whatever, there's a separate "publish" script / command that actually bumps the version, updates all the package.json's, etc.


glitch003:
ok actually the reason it shows so many lines changed is because VSCode formatted the ecdsa-sdk and bls-sdk files when i saved them.  so the 9,700 number is actually my fault and not due to this workflow.  but still, i did only change 2 files and i see 20 files changed and we should see if we can make that better

PR #31, opened 2023-03-07
Fix/filtered out empty shares

Ansonhkg:
# What

at the moment the empty shares are not being filtered out in the response

# Solution

filter out the empty shares before prior combining ECDSA shares, and get its local X & Y values from the first index of the valid shares.


Ansonhkg:
Commit to be reviewed => https://github.com/LIT-Protocol/js-sdk/pull/31/commits/900cd682d526178cea0a745717b6bc63501753e4

packages/crypto/src/lib/crypto.ts
```diff
@@ -191,17 +191,50 @@ export const combineBlsShares = (
...
 
   // the public key can come from any node - it obviously will be identical from each node
   // const publicKey = sigShares[0].publicKey;
   // const dataSigned = '0x' + sigShares[0].dataSigned;
-  const validShares = sigShares.map((s: any) => s.shareHex);
-  const shares = JSON.stringify(validShares);
+  // filter out empty shares
+  const validShares = sigShares.reduce((acc, val) => {
+    if (val.shareHex !== '') {
+      acc.push(val.shareHex);
```
hwrdtm, line 203: this should be 
```suggestion
      acc.push(val);
```

otherwise it is breaking right now.

Also doesn't seem like we have any active cypress tests that test this / the `litNodeClient.executeJs` call?

PR #44, opened 2023-03-13
Feat/separate lit node client node code & add version bump tool

Ansonhkg:
# What 1 - Node code separation

The new TS-SDK is built with being isomorphic in mind, that means sometimes we need to include code that might not be needed in the environment. eg. walletconnect in nodejs

# Proposed Solution

Separate the LitNodeClient node code on its own repo, then the current LitNodeClient extends from it so no breaking changes for existing customers that decides to upgrade the `@lit-protocol/lit-node-client` package

- The new structure will look like this:
    - **LitNodeClientNodeJs** - This is purely for Node.js.
    - **LitNodeClient** - This is isomorphic, so both browser and Node.js code are included. This is the default repo, but if the user decides to, they could use NodeClient instead.

## Usage

This proposed solution is published to dev tag and can be tested in version `2.1.63`

```
// ONLY FOR NODE.JS
yarn add @lit-protocol/lit-node-client-nodejs@2.1.63

// import * as LitJsSdk from "@lit-protocol/node-client";
// const client = new LitJsSdk.LitNodeClientNodeJs();

OR

// ISOMORPHIC
yarn add @lit-protocol/lit-node-client@2.1.63

// import * as LitJsSdk from "@lit-protocol/lit-node-client";
// const client = new LitJsSdk.LitNodeClient();

```

Tested manually and e2e works

# What 2 - SDK tool - version bump

This PR also added a version bump tool which bumps the version in `lerna.json` and `version.ts`

# Solution

```
yarn v // to check version
yarn bump // for patch
yarn bump:minor // for minor
yarn bump:major //for major
```


Ansonhkg:
Will be changing the name from `@lit-protocol/node-client` to `@lit-protocol/lit-node-client-nodejs` cus `node` can mean two things here and don't want to be confusing

PR #45, opened 2023-03-14
fix: types

Ansonhkg:
# What

- Wrong types on symmetric key
- Remove `undefined` return type

# Solution

- Add new `SymmetricKey` type
- Add new `AcceptedFileType` type to accept `File | Blob` type
- Remove `undefined` and replace it with `throw Error` instead

# Test

Published to `dev` tag

```
@lit-protocol/lit-node-client@2.1.65
```


packages/encryption/src/lib/encryption.ts
```diff
@@ -283,7 +285,12 @@ export const decryptZip = async (
     },
   });
 
-  if (!paramsIsSafe) return;
+  if (!paramsIsSafe){
+    throwError({
```
Adarsh-Kumar28, line 289: I think this should be the default approach going forward instead of returning `undefined`. I've seen the later at multiple places we should updates those too prob in future

packages/encryption/src/lib/encryption.ts
```diff
@@ -383,7 +390,10 @@ export const encryptFileAndZipWithMetadata = async ({
     },
   });
 
-  if (!paramsIsSafe) return;
+  if (!paramsIsSafe) return throwError({
+    message: `authSig, accessControlConditions, evmContractConditions, solRpcConditions, unifiedAccessControlConditions, chain, file, litNodeClient, and readme must be provided`,
```
Adarsh-Kumar28, line 394: It's a better dev experience if we rather throw a more specific error- which param is unsafe

packages/encryption/src/lib/encryption.ts
```diff
@@ -693,7 +715,10 @@ export const verifyJwt = ({ jwt }: VerifyJWTProps): IJWT | undefined => {
       functionName: 'verifyJwt',
     })
   )
-    return;
+    return throwError({
+      message: 'jwt must be a string',
```
Adarsh-Kumar28, line 719: `string` -> `String`

Adarsh-Kumar28 (commented):
Looks good, left a couple comments

packages/encryption/src/lib/encryption.ts
Adarsh-Kumar28: We can add this in a separate PR

PR #51, opened 2023-03-18
lit-node-client-nodejs/Readme: Fix forgotten to change import alias.

uji:


Ansonhkg:
Thanks for pointing this out!

PR #54, opened 2023-03-22
added sessionSigs and validators where needed

glitch003:


packages/encryption/src/lib/params-validators.ts
Adarsh-Kumar28, line 1: Should we add a check for the case when both the `authSig` & `sessionSigs` are provided?

packages/misc/src/lib/misc.ts
```diff
@@ -204,14 +202,13 @@ export const checkType = ({
  * @returns { boolean }
```
Adarsh-Kumar28, line 202: In the comments above:

`* @param { object } authSig` -> `* @param { JsonAuthSig } authSig`

Adarsh-Kumar28 (commented):
Looks good, left a comment. There are a few places where the sessionSigs are still missing. I'll add a commit to this PR addressing that.

packages/encryption/src/lib/params-validators.ts
Adarsh-Kumar28: Addressed this: [feat: Validate if only either authSig or sessionSigs are provided](https://github.com/LIT-Protocol/js-sdk/pull/54/commits/757aa0732f17ab8797936f06d6d2fce560c493cb)

Ansonhkg:
added a little tool to check if the built package dependencies matches the package version. This **does not** affect the current workflow. 

```
yarn build:verify
```

Ansonhkg:
@Adarsh-Kumar28 we can merge this

PR #56, opened 2023-03-24
Feat/cosmos signer

Ansonhkg:
# What

This module defines the PKPCosmosWallet class, which is a modified version of the DirectSecp256k1HdWallet class from "@cosmjs/proto-signing". The class extends the PKPBaseWallet class and implements the OfflineDirectSigner interface, enabling it to use PKP for signing. The class handles the creation of a Cosmos wallet, management of account data, and signing transactions with the wallet.
The module exports the PKPCosmosWallet class, as well as the PKPCosmosWalletProp type definition used for initializing the class instances.

Source: https://github.com/cosmos/cosmjs/blob/4c8b278c1d988be3de415f767ce2f65ab3d40bd9/packages/proto-signing/src/directsecp256k1wallet.ts

# Test

```
yarn add @lit-protocol/pkp-cosmos@2.1.95

or 

yarn add @lit-protocol/pkp-cosmos@pkp
```

# Demo Repo
https://github.com/LIT-Protocol/js-sdk-pr-56


Ansonhkg:
Close as this is merged to #57

PR #57, opened 2023-03-24
[WIP] Minor release 2.2.0: PKP Client/library & Cosmos Signer

sarahzdev:
# Notable Changes

- [x] Merge #56  (feat/cosmos-signer) This merge only work in nodejs (see issue below)
- [x] The `@lit-protocol/pkp-ethers` repository no longer exports the `PKPWallet` class. Instead, it has been renamed to  `PKPEthersWallet` class. Additionally, the required property `provider` has been changed to `rpc`, which is now a state within PKPBase that both ethers and cosmos extends or implements from
- [x] Introduce `lit.config.json`: A configuration file that allows you to specify whether you want to send transactions in unit tests that incur gas fees.
- [x] Integrated [lit-pkp-sdk](https://github.com/LIT-Protocol/lit-pkp-sdk) to pkp-ethers
- [x] No longer need to pass RPC to the constructor in order to `signMessage` and `signTransaction` in PKPEthers
- [x] Export helper libraries such as Stargate for Cosmos & { ethers } for Ethereum
- [x] Introduce PKPClient

# Issues to be resolved

- [x] Fix `TypeError: unexpected type, use Uint8Array` while running `*.spec.ts` unit test #59 
- [x] PKPCosmos does not work in React applications due to issues with [@cosmjs/amino](https://github.com/cosmos/cosmjs/tree/main/packages/amino) and other [@cosmjs](https://github.com/cosmos/cosmjs/tree/main/packages) dependencies, as well as problems related to crypto polyfills. The @cosmjs/amino package appears to be a primary cause of these issues, as the error also occurs when it is installed separately in a React app. A temporary fix has been done in https://github.com/LIT-Protocol/js-sdk/pull/60
- [x] Added types to third-party-libs repo for re-exported functions 
- [x] Allow explicit definition of `litActionJsParams`
- [x] Added `webpack.config.js` to React test app to fix crypto/stream polyfill issues

# Dev flows
- [x] https://github.com/LIT-Protocol/js-sdk/pull/57/commits/fa1e5e5c5653a9aedceadb78025e5206f07ab532 (⚠️ Experimental) 
```
yarn dev:target <package-name> --deps
```
While developing a specific repo, you may need to rebuild all packages, which can be time-consuming. This command enables you to build a particular project and its dependencies, then setup a symlink of that repo immediately which would enable you to `npm link` that pacakge. Note that the package order is crucial, as dependent packages must be built and compiled first. This ensures that the latest changes in the dependent package are available for the package that relies on it.

[...]

PR #59, opened 2023-03-25
[WIP] Fix/spec uint8array

Ansonhkg:
# What

This should fix the error when running unit testing on packages that relies on the `uint8arrays` repo

```
TypeError: unexpected type, use Uint8Array

      1101 |   for (var i = 0; i < arguments.length; i++) {
      1102 |     if (!(arguments[i] instanceof Uint8Array))
    > 1103 |       throw new TypeError('unexpected type, use Uint8Array');
           |             ^
      1104 |   }
      1105 | }
      1106 |
```

# Solution

Instead of using the bundled version, we reconstruct the uint8arrayFromString and uint8arrayToString functions, which now support utf8, base16, and base64 encoding/decoding. Test cases have been created and have successfully passed. Check the results at https://cloud.nx.app/runs/Oq5A07AAfn


Ansonhkg:
Closing this as it's merged to #57

PR #60, opened 2023-03-26
Fix/pkp cosmos react

Ansonhkg:
# What

The `@cosmjs/amino` package is currently not supported on React, due to this line in  `pbdkf2.js`

```
/**
 * Returns the Node.js crypto module when available and `undefined`
 * otherwise.
 *
 * Detects an unimplemented fallback module from Webpack 5 and returns
 * `undefined` in that case.
 */
export async function getNodeCrypto(): Promise<any | undefined> {
  try {
    const nodeCrypto = await import("crypto");
    // We get `Object{default: Object{}}` as a fallback when using
    // `crypto: false` in Webpack 5, which we interprete as unavailable.
    if (typeof nodeCrypto === "object" && Object.keys(nodeCrypto).length <= 1) {
      return undefined;
    }
    return nodeCrypto;
  } catch {
    return undefined;
  }
}
```

# Solution

A temporary solution is to change this to try to use webcrypto if available (this was done in the bundled file)

```
      async function getNodeCrypto() {
        try {
          globalThis.crypto = __require("crypto").webcrypto;
        } catch (e) {
        }
      }
```


Ansonhkg:
Closing this as it's merged to https://github.com/LIT-Protocol/js-sdk/pull/57

PR #65, opened 2023-03-29
signSessionKey updates

hwrdtm:
# What

- Merge the webauthn SDK logic into `signSessionKey`
- Make `pkpPublicKey` an optional parameter to `signSessionKey`.

# Testing

Manually tested, it works.

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -2309,7 +2309,7 @@ export class LitNodeClientNodeJs {
   /** ============================== SESSION ============================== */
 
   /**
-   * Sign a session key using a PKP
+   * Sign a session public key using a PKP, which generates an authSig.
    * @returns {Object} An object containing the resulting signature.
    */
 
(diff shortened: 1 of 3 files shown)


packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2415,112 +2424,27 @@ export class LitNodeClientNodeJs {
...
-  ) => {
-    log('getVerifyWebAuthnAuthenticationKeyShares');
-    const urlWithPath = `${url}/web/auth/webauthn`;
-    return await this.sendCommandToNode({
-      url: urlWithPath,
-      data: body,
-      requestId,
-    });
-  };
+  generateAuthMethodForWebAuthn = (
```
sarahzdev, line 2427: May want to consider moving these utility methods outside of this class so one would not have to instantiate the class to use them

hwrdtm: Future PR 🤞🏻

Ansonhkg (approved):
Look good to me - e2e & unit tests unaffected

PR #73, opened 2023-04-05
Handle New / Old SDK Errors

hwrdtm:
# What 

Closes LIT-191. See issue for more details.

- This PR makes it so that old error codes (`storage_error`, `not_authorized`) from the node, as well as errors thrown from within the SDK (unrelated to node API calls) **are all unified into the new style of PascalCase codes, as well as the ability to parse new error response fields such as `errorKind`, `status` etc.**
  - If nodes return `errorCode: 'storage_error'`, SDK returns `errorCode: 'NodeStorageError'`
  - If nodes return `errorCode: 'NodeStorageError'`, SDK returns the same.

# Testing

Run `yarn build:packages`, then `yarn tools --dev --apps` to spin up `http://localhost:4002` which you can use for manual testing.

Manually tested using `http://localhost:4002/manual_tests.html` as an example, and the `Encrypt, set the conditions, then try to update the conditions with an unauthorized account` worked as expected:

Screenshot 2023-04-04 at 11 50 09 PM

### DIFF
--- a/apps/html/manual_tests.html
+++ b/apps/html/manual_tests.html
@@ -1290,7 +1290,7 @@
           console.log('updated the condition');
         } catch (e) {
           console.log('error getting symmetric key - this is expected', e);
-          if (isNodeStorageError(e.error_code)) {
+          if (isNodeStorageError(e.errorCode)) {
             document.getElementById(
               'status'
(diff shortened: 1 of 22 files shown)


Ansonhkg:
@saike Thanks for the PR! Merged and published to latest (2.1.111)!

PR #75, opened 2023-04-06
[wip] update PKPBase props

sarahzdev:
**Summary**
- Flatten `pkpWalletProp` so that the nested properties are now part of `PKPBase` (makes it easier to read, override, test, etc)
- Create and set session sigs and track the expiration
- In `pkp-ethers` `signMessage`, wrap rpc provider calls in a try catch (don't want to prevent running the Lit Action)


packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -40,20 +43,27 @@ const compressPubKey = (pubKey: string): string => {
  * A base class that can be shared between Ethers and Cosmos signers.
  */
 export class PKPBase<T = PKPBaseDefaultParams> {
-  pkpWalletProp: PKPBaseProp;
+  pkpPubKey: string;
```
Ansonhkg, line 46: the term `pkpPubKey` essentially refers to the `uncompressedPubKey`; i don't mind them both coexists. The rationale for this is that pkpPubKey is more Ethereum-centric, while uncompressed/compressed public keys are more neutral and can be applied to different blockchains

PR #77, opened 2023-04-07
Add support for Leap & Cypher wallets for Cosmos

josefleventon:
This PR replaces the previously hardcoded window.keplr with multiple wallet options for Cosmos (Keplr/Leap/Cypher) when using `checkAndSignAuthMessage`.

It does not interfere with the functionality of any other chain option except the Cosmos chains. It also defaults to Keplr when a specific wallet is not specified, meaning no changes to current implementation would be necessary.

Note: No `cypher` option is present as Cypher Wallet uses the `window.keplr` object.


Ansonhkg:
lgtm - tested and worked! Thanks for the PR! @josefleventon

PR #79, opened 2023-04-13
refactor: improve getSessionKey

sarahzdev:
**Summary**
- Be more explicit that one can pass in a serialized session key pair
- Enforce that `getSessionKey` returns a valid `SessionKeyPair`

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -295,7 +295,10 @@ export class LitNodeClientNodeJs {
         if (this.isSessionKeyPair(keyPair)) {
           return keyPair;
         } else {
-          throw new Error('Invalid session key pair provided');
+          return throwError({
+            message: 'Invalid session key pair provided',
+            error: LIT_ERROR.PARAMS_MISSING_ERROR,
+          });
         }
       } catch (err) {
         log(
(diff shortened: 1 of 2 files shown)


glitch003 (approved):
looks good, thanks @sarahzdev

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -281,53 +281,96 @@ export class LitNodeClientNodeJs {
...
-    if (sessionKey === '') {
-      // check if we already have a session key + signature for this chain
-      // let storedSessionKey;
-      let storedSessionKey: any;
-
-      // -- (TRY) to get it in the local storage
-      if (storedSessionKeyOrError.type === 'ERROR') {
-        console.warn(
-          `Storage key "${storageKey}" is missing. Not a problem. Contiune...`
+  getSessionKey = (serializedSessionKeyPair?: string): SessionKeyPair => {
```
hwrdtm, line 290: does it need to be serialized? is it more convenient to have clients pass in the raw keypair object instead? it's not like the keypair is something that needs to travel over the wire i would presume.....(and hence would need serialization if it does travel over the wire)

sarahzdev: @hwrdtm 

since `getSessionKey` is currently called in more client-facing methods like `getSessionSigs`, the key pair doesn't have to be serialized. i'm open to enforcing the type in `getSessionKey`, `getSessionSigs`, and `signSessionKey` but this would be a breaking change.

PR #85, opened 2023-04-20
Feature/lit 293 add registration for otp auth method

joshLong145:
# WHAT
- Adds first pass at integrating OTP
- Allows for an `AuthMethod` to be constructed from a successful OTP verification flow

# TODO
- Add default configuration for deployed OTP server
- Create an interface for communicating with the OTP server similar to the `relayer`
- Add tests with mocked OTP server 
- Refactor to use provider pattern when ready


packages/types/src/lib/interfaces.ts
```diff
@@ -1071,6 +1159,33 @@ export interface EthWalletProviderOptions {
   origin?: string;
 }
 
+export interface SignInWithOTPParams {
```
sarahzdev, line 1162: duplicate type def

packages/types/src/lib/interfaces.ts
```diff
@@ -915,6 +915,94 @@ export interface LitAuthClientOptions {
    * Lit Node Client
    */
   litNodeClient?: any;
+
+  litOtpConfig?: OtpProviderOptions
+}
+
+export interface SignInWithEthWalletParams {
```
sarahzdev, line 922: no longer need this (this is removed in master)

PR #67, opened 2023-03-30
feat: new pkg with pkp auth helpers

sarahzdev:
# lit-auth-client

The new `lit-auth-client` packages provides a `LitAuthClient` class and `BaseProvider` classes to streamline authentication for PKPs.

## Features
- Handle authentication with Google and Discord via [Lit's login server](https://github.com/LIT-Protocol/lit-login-server)
- Handle authentication with Eth wallet by generating an authSig
- Handle registration and authentication via WebAuthn
- Mint and fetch PKPs using [Lit's relay server](https://github.com/LIT-Protocol/relay-server)
- Wraps `getSessionSigs` to create session sigs for a given PKP and auth method

## Changes
- Created a new browser-only package `lit-auth-client`
- In `lit-auth-client`:
  - Added new classes and methods to support auth and relay server interactions
  - Added basic tests for utility methods
  - Merged in changes from #86 
- Updated `types` and `constants` packages to support the new types
- Updated `getSessionKey` method in `lit-node-client-nodejs` (merged in #79)

## Testing

The updated packages are currently published with the `dev` tag (e.g., `yarn add @lit-protocol/lit-auth-client@dev`).

Tested all functionality in this [Next.js TypeScript app](https://github.com/LIT-Protocol/pkp-social-auth-example/pull/1). See the deployed preview [here](https://pkp-social-auth-example-4jux05hev-lit-protocol.vercel.app/).

Also tested the same functionality as above in local React, React TS, and Next.js apps.

## Next Steps

- Handle social login flow in pop-up
- Add helper functions to mint and fetch PKPs via contract calls (wrapping methods from the `contracts-sdk`)
- Error class / type

PR #86, opened 2023-04-21
refactor: move auth, session, and relay logic into providers

sarahzdev:
**Summary**
- Refactor authentication logic for Google, Discord, EthWallet (now EthereumAccount) by moving it outside of `LitAuthClient` and into their respective provider classes
- Code related to handling authentication, session sigs, and relay server has been reorganized into the corresponding provider classes
- Added WebAuthn support

Notes
- Added `moduleResolution: "node16"` to handle dynamically import `@simplewebauthn/browser`, which is ESM-only (more context on the issue/workaround [here](https://github.com/microsoft/TypeScript/issues/43329)
- This flag breaks the spec so commented the spec for now
- Porting README content to the dev docs

TODO:
[x] Update spec test
[x] Update demo app
[x] Update readme
[ ] Merge into PR #67


packages/lit-auth-client/src/lib/lit-auth-client.ts
```diff
@@ -82,385 +65,102 @@ export class LitAuthClient {
...
-      signedMessage: toSign,
-      address: params.address,
-    };
-
-    const authMethod = {
-      authMethodType: AuthMethodType.EthWallet,
-      accessToken: JSON.stringify(authSig),
-    };
-    return authMethod;
+  initDiscordProvider(params: InitOAuthProviderOptions): DiscordProvider {
```
joshLong145, line 106: Perhaps this shouldn't be explicitly defined for each provider type.
maybe instead there is a single `initProvider<T extends ProviderOptions>(params: ProviderOptions<T>)`
T in the `ProviderOptions` will be a concrete interface type for the Provider Options. A bit cleaner than having an explicit init.

packages/lit-auth-client/src/lib/relay.ts
```diff
@@ -162,6 +162,34 @@ export class LitRelay implements IRelay {
...
   }
 
+  /**
+   * Generate options for registering a new credential to pass to the authenticator
+   *
+   * @param {string} [username] - Optional username to associate with the credential
+   *
+   * @returns {Promise<any>} Registration options for the browser to pass to the authenticator
+   */
+  public async generateRegistrationOptions(username?: string): Promise<any> {
```
joshLong145, line 172: Should this be defined on `IRelay`

packages/lit-auth-client/src/lib/providers/WebAuthnProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+      rpId,
+    };
+
+    // Authenticate with WebAuthn
+    const authenticationResponse = await startAuthentication(
+      authenticationOptions
+    );
+
+    // BUG: We need to make sure userHandle is base64url encoded. Deep copy the authentication response.
+    const actualAuthenticationResponse = JSON.parse(
```
joshLong145, line 106: Is this a bug, the spec does require the `userId` to be base64 encoded.

packages/lit-auth-client/src/lib/providers/WebAuthnProvider.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  /**
+   * Authenticate with WebAuthn
+   *
+   * @param {string} [rpcUrl] - RPC URL to use for getting latest block hash
+   *
+   * @returns {Promise<AuthMethod>} - Auth method object containing WebAuthn auth data
+   */
+  public async authenticate(): Promise<AuthMethod> {
```
joshLong145, line 81: should we add this method signature to `baseProvider`? might be useful for other auth methods in the future.

packages/lit-auth-client/src/lib/providers/DiscordProvider.ts
```diff
@@ -0,0 +1,87 @@
...
+    // Redirect to login url
+    window.location.assign(loginUrl);
+  }
+
+  /**
+   * Validate the URL parameters returned from Lit's login server and return the OAuth token
+   *
+   * @returns {AuthMethod} - Auth method object containing the OAuth token
+   */
+  public authenticateWithLit(): AuthMethod {
```
joshLong145, line 39: perhaps we uniform this to `authenticate` and add to `BaseProvider`

joshLong145 (changes requested):

packages/lit-auth-client/src/lib/lit-auth-client.ts
sarahzdev: @joshLong145 - updated, thanks!

packages/lit-auth-client/src/lib/providers/WebAuthnProvider.ts
sarahzdev: @joshLong145  - updated the comment and `userHandle` check slightly. looks like the `@simplewebauthn/browser` pkg we’re using returns the `userhandle` as a regular string seen [here](https://github.com/MasterKale/SimpleWebAuthn/blob/76c7a46b60b1093e9e71f38670175bee45512473/packages/typescript-types/src/index.ts#L145)

packages/lit-auth-client/src/lib/providers/WebAuthnProvider.ts
sarahzdev: @joshLong145 - added an abstract method `authenticate` to `BaseProvider`, ty!

packages/lit-auth-client/src/lib/providers/DiscordProvider.ts
sarahzdev: @joshLong145 - updated this to implement the `BaseProvider`'s `authenticate`

packages/lit-auth-client/src/lib/providers/WebAuthnProvider.ts
joshLong145: Oh I see, that's good to know.

joshLong145 (approved):

PR #103, opened 2023-05-06
Add explicit permanent type

bntzio:
The `saveSigningCondition` and `saveEncryptionKey` functions in the `lit-node-client-nodejs` package already handles the logic to add the correct boolean value (`permanent ? 1 : 0`) to the correspondent functions (`storeSigningConditionWithNode` and `storeEncryptionConditionWithNode` respectively).

This way, users can either specify a boolean value to the `permanant` or `permanent` parameters, or directly a `0` or `1`.

```ts
const encryptedSymmetricKey = await this.litNodeClient.saveEncryptionKey({
  solRpcConditions,
  symmetricKey,
  authSig,
  chain,
  permanent: false // <--- or directly 0
})
```

This is easier for users to understand which value they can pass as parameters, by making the `permanent` type more explicit, accepting either `0` or `1` only, if not the boolean.

Here's an actual example of the typescript lint error:

CleanShot 2023-05-06 at 11 11 07


Ansonhkg:
Thanks for the PR! Merged and published to version 2.1.161 (latest tag)!

PR #110, opened 2023-05-17
Add tslib@^2.3.0 as peer dependency of auth-helpers

hwrdtm:
# What

Without this PR, client apps that use older versions of `tslib` will encounter the following issue

image


packages/auth-helpers/package.json
```diff
@@ -17,6 +17,9 @@
     "access": "public",
     "directory": "../../dist/packages/auth-helpers"
   },
+  "peerDependencies": {
+    "tslib": "^2.3.0"
+  },
```
joshLong145, line 22: we might want to remove the `^` in the version if versions ABOVE 2.3.0 are not intended to be used.

joshLong145 (approved):

packages/auth-helpers/package.json
hwrdtm: This is consistent with the root package.json's dep of tslib being ^2.3.0, and, just tested it in a client app which installed 2.5.0 and it works!

PR #112, opened 2023-05-17
Add peerDependencies to dependencies regardless if any imports detected

hwrdtm:
# What

This PR **adds** `peerDependencies` to `dependencies` regardless if any imports / dependencies have been detected. 

Sometimes we may want to specify a peerDependency for a library that does not even get imported - for example, `auth-helpers` specify `tslib` peer dependency but there is no explicit import for that in code that is detected by us.

### DIFF
--- a/tools/scripts/map-deps-to-dist.mjs
+++ b/tools/scripts/map-deps-to-dist.mjs
@@ -48,21 +48,23 @@ await asyncForEach(packagesDir, async (packageDir) => {
 
   const dependencies = [...new Set(imports.flat())];
 
-  if (dependencies.length) {
-    const packageJsonPath = DIST_DIR + '/' + packageDir + '/package.json';
+  // Add peerDependencies to dependencies
+  const packageJsonPath = DIST_DIR + '/' + packageDir + '/package.json';
 
-    const packageJson = await readJsonFile(packageJsonPath);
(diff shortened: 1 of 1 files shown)


rahul-ramesh (approved):

tools/scripts/map-deps-to-dist.mjs
```diff
@@ -48,21 +48,23 @@ await asyncForEach(packagesDir, async (packageDir) => {
...
+  // Add peerDependencies to dependencies
+  const packageJsonPath = DIST_DIR + '/' + packageDir + '/package.json';
 
-    const packageJson = await readJsonFile(packageJsonPath);
+  const packageJson = await readJsonFile(packageJsonPath);
 
-    // await asyncForEach((dependencies), async (dep) => {
-    //     packageJson.dependencies[dep] = packageJson.version;
-    // })
+  // await asyncForEach((dependencies), async (dep) => {
```
rahul-ramesh, line 56: are we planning to use this code again soon?

glitch003 (approved):

tools/scripts/map-deps-to-dist.mjs
hwrdtm: not sure, defer to @Ansonhkg , but in any case we can do a future PR to clean that up.

PR #140, opened 2023-06-12
Fix/uint8array encoding base64

joshLong145:
# WHAT
- Fixes issues when encoding large `Uint8Array` to base64 strings 

# TESTING
- Addition of unit test for large array encoding and decoding.

### DIFF
--- a/packages/uint8arrays/src/lib/uint8arrays.spec.ts
+++ b/packages/uint8arrays/src/lib/uint8arrays.spec.ts
@@ -144,7 +144,7 @@ describe('conversion', () => {
 
       it('base64 large encoding should decode', () => {
         // generate a random base64urlpad string of length 1333 (which is equivalent to 1000 bytes when decoded)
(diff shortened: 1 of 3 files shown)


packages/uint8arrays/src/lib/uint8arrays.spec.ts
```diff
@@ -1,4 +1,6 @@
+import { base64ToUint8Array } from 'packages/uint8arrays/dist/src/lib/uint8arrays';
 import { uint8arrayFromString, uint8arrayToString } from './uint8arrays';
+import { base64StringToBlob } from '../../../misc-browser/src/lib/misc-browser';
```
Ansonhkg, line 3: This is triggering circular dependency error on build. 

1. change the import to `from '@lit-protocol-misc-browser`
2. In `project.json` add `"implicitDependencies": ["!misc-browser"],`

PR #144, opened 2023-06-14
Add manual test for sepolia

hwrdtm:
# What

Added manual test for Sepolia as we add support for it on the node side.


glitch003 (approved):

hwrdtm:
Tested manually, works.

PR #157, opened 2023-06-26
Feature/lit 787 add captcha response check to lit auth client for otp

joshLong145:
- Adds ReCaptcha to `lit-auth-client` for `OtpProvider`
   - Adds support for inject ReCaptcha scripts into the DOM with anchor tag injection, and script injection into provided `head` node.
- Adds ReCaptcha response to `sendOtp` request
- Allows for passing ReCaptcha `response` as init params to `lit-auth-client`

### DIFF
--- a/apps/demo-email-sms-auth/src/Otp.css
+++ b/apps/demo-email-sms-auth/src/Otp.css
@@ -9,6 +9,13 @@ span {
     justify-content: center;
 }
 
+.captcha {
+    display: flex;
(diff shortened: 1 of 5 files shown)


joshLong145:
> lgtm! could we update the `lit-auth-client` README with a section explaining what gets injected automatically and outlining its behavior in varying contexts, eg. browser vs. nodejs?

Yes, I'll add some updates to the README

PR #161, opened 2023-06-28
Add/cheqd-specific cosmos ACC properties

Eengineer1:
- Adds required cheqd-specific explicit propertieson ACC types.
  - Utilises `parameters` as `string[]`, rather than generic `params` as `any[]`. Details are implementation specific.
  - Brings `method` as optional at ACC definition layer for consistency.

Implementation Reference: https://github.com/LIT-Protocol/lit-assets/pull/244 .


Ansonhkg (approved):
lgtm

PR #182, opened 2023-07-06
Feature/lit 900 add function to js sdk to enable signing without lit actions

joshLong145:
# WHAT
- Removes the requirement of running a lit action when signing messages
- Creates a new `pkpSign` method on `LitNodeClientNodeJs`
- Support added to `pkip-base` to detect if default configuration is present and if so uses the `pkpSign` method over `executeJS`
    - `pkp-cosmos` `pkp-ethers`, have been upgraded to use the new utilities in `pkp-base`

# Testing
- Added   tests to `lit-node-client` and run all unit and ci tests to confirm new endpoints are working as expected with the existing features of effected packages.


glitch003 (approved):
looks good, thanks josh

PR #124, opened 2023-05-30
Feature/lit 462 remove verification steps from relayer

joshLong145:
# What
- Supports removing the authentication implementations present within the current relayer for allowing the Relayer server to be a pure pass through for minting and fetching PKP's.
- Adds `getRelayerRequest` method to `BaseProvider` for constructing new `RelayerRequest` for minting or querying PKP's.

PR #194, opened 2023-08-08
feat: support new mint and fetch endpoints

sarahzdev:
**Summary**
- Add helpers to each provider to generate the `authMethodId`
- Copy over @joshLong145 's work on parsing out public key from WebAuthn response from PR #124 
- Update mint and fetch methods to support the new relay server routes
- Update demo test app to showcase most of the auth methods (excluding Apple)

**Testing**
Run `demo-pkp-social-auth-next-ts` to test locally. Note that session sigs do not currently work for eth wallet and Stytch flows.


packages/lit-auth-client/src/lib/providers/WebAuthnProvider.ts
```diff
@@ -114,4 +163,20 @@ export default class WebAuthnProvider extends BaseProvider {
...
+  /**
+   * Get auth method id that can be used to look up and interact with
+   * PKPs associated with the given auth method
+   *
+   * @param {AuthMethod} authMethod - Auth method object
+   *
+   * @returns {Promise<string>} - Auth method id
+   */
+  public async getAuthMethodId(authMethod: AuthMethod): Promise<string> {
+    const credentialId = JSON.parse(authMethod.accessToken).rawId;
```
joshLong145, line 176: we should add a guard here incase the `accessToken` cant parse

packages/lit-auth-client/src/lib/relay.ts
```diff
@@ -36,24 +36,22 @@ export class LitRelay implements IRelay {
...
-      method: 'POST',
-      headers: {
-        'api-key': this.relayApiKey,
-        'Content-Type': 'application/json',
-      },
-      body: body,
-    });
+  public async mintPKP(body: string): Promise<IRelayMintResponse> {
+    const response = await fetch(
+      `${this.relayUrl}/mint-next-and-add-auth-methods`,
```
joshLong145, line 45: nit: the endpoint should be a constant definition

packages/lit-auth-client/src/lib/relay.ts
```diff
@@ -131,17 +129,12 @@ export class LitRelay implements IRelay {
...
    * @returns {Promise<IRelayFetchResponse>} Response from the relay server
    */
-  public async fetchPKPs(
-    authMethodType: AuthMethodType,
-    body: string
-  ): Promise<IRelayFetchResponse> {
-    const route = this._getFetchPKPsRoute(authMethodType);
-    const response = await fetch(`${this.relayUrl}${route}`, {
+  public async fetchPKPs(body: string): Promise<IRelayFetchResponse> {
+    const response = await fetch(`${this.relayUrl}/fetch-pkps-by-auth-method`, {
```
joshLong145, line 137: nit: endpoint should be a constant definition

packages/types/src/lib/interfaces.ts
```diff
@@ -763,6 +763,7 @@ export interface SessionKeyPair {
 export interface AuthMethod {
   authMethodType: number;
   accessToken: string;
+  pubkey?: string;
```
joshLong145, line 766: Should we separate the idea of `Auth Method` and data used to register auth methods?
I think it would help as the `AuthMethod` doesnt technically have this concept. in my previous PR i added:
```
export interface RelayerRequest {
  authMethodType: number;
  authMethodId: string;
  // used only in Webauthn credential registration
  authMethodPubKey?: string;
}
```
I think we should introduce it again as it is used in the new relay route.

joshLong145 (changes requested):
looks good, left a few comments

packages/types/src/lib/interfaces.ts
sarahzdev: sgtm, added `IRelayRequestData`

joshLong145 (approved):

Issue #195, opened 2023-08-10
Error with humanizeAccessControlConditions() when using ERC20 tokens


mattcasey:
**Before reporting, I have already *** 
- checked LIT Protocol [developer docs](https://developer.litprotocol.com/) and/or the [lit-js-sdk](https://lit-protocol.github.io/lit-js-sdk/api_docs_html/).
- searched for other [open issues](https://github.com/LIT-Protocol/lit-js-sdk/issues).

**Describe the bug**
When calling humanizeAccessControlConditions() with an ERC20 token, an exception is thrown.

**To Reproduce**
Code sample:
```
import { humanizeAccessControlConditions } from '@lit-protocol/lit-node-client';

function Component () {
  useEffect((async () => {
    const result = await humanizeAccessControlConditions({
      chains: ['goerli'],
      permanent: true,
      authSigTypes: ['ethereum'],
      unifiedAccessControlConditions: [
        {
          chain: 'goerli',
          method: 'balanceOf',
          parameters: [':userAddress'],
          conditionType: 'evmBasic',
          contractAddress: '0x5b8B8C9aD976aFCAd24fd6CF424294d372c190Ac',
          returnValueTest: {
            value: '100000000000000000000',
            comparator: '>='
          },
          standardContractType: 'ERC20'
        }
      ]
    });
  })();
  , []);
}
```

**Expected behavior**
Should return a human-friendly sentence of ERC20

**Environment:**
 - Lit SDK version: 2.2.47 (latest)

**Additional info**
This is the line that blows up: https://github.com/LIT-Protocol/js-sdk/blob/master/packages/misc/src/lib/misc.ts#L479

This is the error:
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'map')
    at new Interface (interface.js:57:23)
    at getInterface (index.js:625:1)
    at new BaseContract (index.js:481:52)
    at new Contract (index.js:899:8)
    at decimalPlaces (misc.js:348:1)
    at eval (humanizer.js:182:1)
    at Array.map (<anonymous>)
    at humanizeEvmBasicAccessControlConditions (humanizer.js:98:1)
    at eval (humanizer.js:399:1)
    at Array.map (<anonymous>)
    at humanizeUnifiedAccessControlConditions (humanizer.js:380:1)
    at humanizeAccessControlConditions (humanizer.js:469:1)
    at eval (TokenGateOption.tsx:47:36)
```

[...]

PR #196, opened 2023-08-11
Fix/humanize acc

joshLong145:
# WHAT
Fixes issue #195 
The `ERC20` Contract ABI is imported as a `Promise` instead of an `Object` type. Changing import statements from using `import()` to module imports results in `ERC20` resolving to an `Object` when imported.


sarahzdev (approved):

Ansonhkg (approved):
lgtm

PR #145, opened 2023-06-14
[V3] Upgrade ACC-based encryption + JWT signing

hwrdtm:
# What

Closes LIT-804.
Closes LIT-695.
Related to LIT-951.

⚠️ **This is a massive PR. Brace yourself with some 🍿  and 🍵 !!** ⚠️ 

This PR:
- Deserves a major version bump. We are completely overhauling the ACC-based encryption and JWT signing schemes with new schemes, and are bringing in a number of incompatible changes.
- **ID-based encryption**: This new scheme is also known as Timelock Encryption, or TimeCrypt, where there exists an Identity Parameter, and the signature over it is a decryption key over a ciphertext that has been generated with this Identity Parameter. In other words, signature shares - with the Identity Parameter as the signature payload - are combined into a signature, and this signature can be used to decrypt an associated ciphertext. Read more in the spec here: https://www.notion.so/litprotocol/Switch-from-Signcryption-to-BLS-signature-is-key-scheme-b75c21927ed64511b523c8a93b870a65. 
  - Note that the encryption step is non-interactive and happens entirely client-side.
- **ACC-based JWT Signing (V2)**: This new scheme simplifies the overall flow by baking into each dapp endpoint the required access control conditions directly. This means that there is no "store signing condition" step as dapp developers just write it into their dapp code and an additional check against the access control condition is performed there. When clients request for a signed JWT with claims of the specified access control condition, the signature of this JWT is only provided when the nodes determine that the user satisfy this access control condition. Read more in the spec here: https://www.notion.so/litprotocol/ACC-based-JWT-signing-V2-a4db1446bac54eea90b7ab08e5455246.
- The BLS WASM (`bls-sdk.ts`) is entirely replaced with new code. Unit tests are added to this package.
- **The SDK no longer supports the `jalapeno` network moving forward. Developers can use the old SDK versions to continue using the old encryption and JWT signing schemes**. `serrano` network is the new default.
- Updated `crypto` package to be a lightweight wrapper over `bls-sdk` package with minor conversion logic.
- Updated `encryption` package to replace the old encryption + JWT signing logic with the new schemes. This is becoming more a helper library over `lit-node-client-nodejs` and `crypto`.
- Refactored `params-validators` to use a plugin-based approach with various `*Validator`s.
- Updated all manual tests in `apps/html` to use the new ID-based encryption and JWT signing schemes. You will notice quite a few test cases are removed since we no longer need to store encryption keys, or store signing conditions respectively.
[...]

PR #107, opened 2023-05-11
Integrate/cait sith

joshLong145:
- Adds support for new ECDSA implementation for
     - k256
     - p256
- Updates `ecdsa-sdk.ts` in `packages/ecdsa-sdk` with updated wasm bundle
- Adds  post build step to `ecdsa-sdk` for correcting a tic generation error
- Update `SIGTYPE` to include new ecdsa types
- Adds key recovery tests to `packages/crypto`

*Note* waiting on an update to the wasm implementation before this can be merged, but can still be reviewed. Currently the sign ecdsa lit node client test fails as it uses the current nodes which are not upgraded to the new Cait Sith code and still return a  `signType` of `ECDSA`

PR #199, opened 2023-08-22
[version 3.0.0] Major version bump

joshLong145:
!! SDK VERSION 3.0.0 Release Branch !!
Changes from #145 
```
ACC-based JWT Signing (V2): This new scheme simplifies the overall flow by baking into each dapp endpoint the required 

access control conditions directly. This means that there is no "store signing condition" step as dapp developers just write it into their dapp code and an additional check against the access control condition is performed there. 

When clients request for a signed JWT with claims of the specified access control condition, the signature of this JWT is only provided when the nodes determine that the user satisfy this access control condition. Read more in the spec here: https://www.notion.so/litprotocol/ACC-based-JWT-signing-V2-a4db1446bac54eea90b7ab08e5455246.

The BLS WASM (bls-sdk.ts) is entirely replaced with new code. Unit tests are added to this package.
The SDK no longer supports the jalapeno network moving forward. Developers can use the old SDK versions to continue using the old encryption and JWT signing schemes. serrano network is the new default.

Updated crypto package to be a lightweight wrapper over bls-sdk package with minor conversion logic.
Updated encryption package to replace the old encryption + JWT signing logic with the new schemes. This is becoming more a helper library over lit-node-client-nodejs and crypto.

Refactored params-validators to use a plugin-based approach with various *Validators.
Updated all manual tests in apps/html to use the new ID-based encryption and JWT signing schemes. You will notice quite a few test cases are removed since we no longer need to store encryption keys, or store signing conditions respectively.
demo-encrypt-decrypt-react:

Turned this into a TS project.

Updated this demo to use the new ID-based encryption scheme - just a encryptString and decryptToString call.
Implement demo-locked-express-app to demo the new JWT signing flow. Instructions in its README.md.
Made use of generics in a number of places, eg. IEither, SuccessNodePromises and benchmark()
```

Changes from #107
```
 - Adds support for new ECDSA implementation for
     - k256
     - p256
- Updates `ecdsa-sdk.ts` in `packages/ecdsa-sdk` with updated wasm bundle
- Adds  post build step to `ecdsa-sdk` for correcting a tic generation error
- Update `SIGTYPE` to include new ecdsa types
- Adds key recovery tests to `packages/crypto`

```
- Adds `computePubKey` to `lit-core` which wraps an implementation in `crypto` for interfacing with a new wasm module for deriving HD public keys: for reference the implementation which is compiled to WASM is here: https://github.com/LIT-Protocol/lit-ecdsa-wasm-combine/blob/main/src/combiners/hd_ecdsa/mod.rs
- Addition of `claimKeyId` method on `lit-node-client-nodejs` for  deriving a key from an `authMethod`
   - Supports a new `MintCallBack` which is defined as `async (params: ClaimKeyResponse): Promise<ClaimKeyResponse>` which is called to route derived keys from a claim operation on chain.
- New `ClaimProcessor` abstraction for specifying arguments to a callback implementation for forwarding aggregated claim responses from nodes to the `Chronicle` chain which can be done by a `relayer` or `contract client` implementation within the callback.


packages/misc/src/lib/misc.ts
```diff
@@ -498,30 +501,37 @@ export const genRandomPath = (): string => {
...
+      : 'http://127.0.0.1:8081/auth/claim';
     const response = await fetch(relayUrl, {
-      method: "POST",
+      method: 'POST',
       body: JSON.stringify(params),
       headers: {
-        'api-key': "67e55044-10b1-426f-9247-bb680e5fe0c8_relayer",
+        'api-key': params.relayApiKey
+          ? params.relayApiKey
+          : '67e55044-10b1-426f-9247-bb680e5fe0c8_relayer',
```
Ansonhkg, line 517: was thinking we could use the nullish coalescing operator (??) but this is actually better, it supports falsy values including `''` and `0`

packages/types/src/lib/types.ts
```diff
@@ -119,4 +122,50 @@ export type ProviderOptions =
...
+ * Model for requesting a PKP to be claimed based on an {@link AuthMethod} identifier
+ * the {@link MintCallback} may be defined for custom processing of the {@link ClaimResult}
+ * which requires registering on chain, by default chain registering will be done by an external relay.
+ * @property {AuthMethod} authMethod to derive the key id from for claiming
+ * @property {MintCallback} mintCallback optional callback for custom on chain registration behavior
+ */
+export type ClaimRequest<T = ClaimProcessor> = {
+  authMethod: AuthMethod;
+  mintCallback?: MintCallback<T>, 
+} & (T extends 'relay' ? LitRelayConfig : {signer: ethers.Signer});
```
Ansonhkg, line 161: can `T extends RelayClaimProcessor` instead? would it work?

packages/types/src/lib/types.ts
```diff
@@ -119,4 +122,50 @@ export type ProviderOptions =
...
+
+/**
+ * Result from network claim proccessing, used in {@link MintCallback}
+ */
+export type ClaimResult<T = ClaimProcessor> = {
+  signatures: Signature[],
+  derivedKeyId: string,
+  authMethodType: AuthMethodType
+  pubkey: string, 
+} & (T extends 'relay' ? LitRelayConfig : {signer: ethers.Signer});
```
Ansonhkg, line 171: same as above

packages/types/src/lib/types.ts
```diff
@@ -119,4 +122,50 @@ export type ProviderOptions =
...
+/**
+ * Type aggregate for Claim proccessor types
+ */
+export type ClaimProcessor = RelayClaimProcessor | ClientClaimProcessor;
+
+/**
+ * Callback for processing claim requests.
+ * Processing can be done either by a relay server or a contract client
+ * for a drop in client the `contract-sdk` or  your own contract client with correct ABI's and contract addresses.
+ */
```
Ansonhkg, line 146: Revised comment:
```
/**
 * Callback function for processing claim requests.
 * 
 * This function can be used in two scenarios:
 * 1. When the claim is processed by a relay server.
 * 2. When the claim is processed by a contract client.
 * 
 * For contract clients, you can use the `contract-sdk` or implement your own client.
 * Ensure that your client has the correct ABI and contract addresses for successful processing.
 */
```

Ansonhkg (approved):

Ansonhkg:
Can we include this which is in the revamp branch atm.

packages/auth-browser/src/lib/chains/eth.ts
- https://github.com/LIT-Protocol/js-sdk/commit/de730e7ee73fe0183c6ac9ab4a51802cb68c04f5#diff-d013b1d6f980a3138dae1a272b673d0667786974ce29d2cdf43b1c673d135b7dR633-R647

Ansonhkg (changes requested):
Can we include this which is in the revamp branch atm.

packages/auth-browser/src/lib/chains/eth.ts
- https://github.com/LIT-Protocol/js-sdk/commit/de730e7ee73fe0183c6ac9ab4a51802cb68c04f5#diff-d013b1d6f980a3138dae1a272b673d0667786974ce29d2cdf43b1c673d135b7dR633-R647

packages/types/src/lib/types.ts
joshLong145: that does work. has been changed

Ansonhkg (approved):
All good so far. we need to update the CHANGELOG.md

joshLong145:
closing for #266

PR #207, opened 2023-09-04
feat: add handshake body checks on 'successful' handhshakes

Ansonhkg:


packages/core/src/lib/lit-core.ts
```diff
@@ -149,6 +149,21 @@ export class LitCore {
...
+            keys.subnetPubKey === 'ERR' ||
+            keys.networkPubKey === 'ERR' ||
+            keys.networkPubKeySet === 'ERR'
+          ) {
+            log('Error connecting to node ', url, keys);
+            throwError({
+              message: `Error connecting to node ${url}`,
+              errorKind: LIT_ERROR.INVALID_ARGUMENT_EXCEPTION.kind,
+              errorCode: LIT_ERROR.INVALID_ARGUMENT_EXCEPTION.name,
+            });
```
Ansonhkg, line 164: Actually probably not a good idea to throw an error as the other functions are still working, and might break other apps. Logging itself should be fine.

PR #213, opened 2023-09-08
Feat/manual claim test

glitch003:


rahul-ramesh (approved):

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2058,24 +2058,30 @@ export class LitNodeClientNodeJs extends LitCore {
       return this.getClaimKeyExecutionShares(url, nodeRequestParams, requestId);
     });
 
-    const responseData = await this.handleNodePromises(nodePromises);
+    const responseData = await this.handleNodePromises(
+      nodePromises,
+      this.connectedNodes.size // require from all connected nodes
```
glitch003, line 2063: we will fix the contracts to not need all nodes for derivation, and then we can change this.

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2088,15 +2094,17 @@ export class LitNodeClientNodeJs extends LitCore {
           relayUrl: (params as ClaimRequest<'relay'>).relayUrl,
           relayApiKey: (params as ClaimRequest<'relay'>).relayApiKey,
         });
+      } else {
+        mintTx = await defaultMintClaimCallback({
+          derivedKeyId,
+          authMethodType: params.authMethod.authMethodType,
+          signatures: nodeSignatures,
+          pubkey,
+          relayUrl: (params as ClaimRequest<'relay'>).relayUrl,
```
glitch003, line 2103: these changes were needed to support overriding the relayUrl and api key when no callback function is passed in

packages/lit-auth-client/src/lib/lit-auth-client.ts
```diff
@@ -74,10 +74,7 @@ export class LitAuthClient {
     }
 
     // Check if Lit node client is provided
-    if (
-      options?.litNodeClient &&
-      options.litNodeClient instanceof LitNodeClient
-    ) {
+    if (options?.litNodeClient) {
```
glitch003, line 77: i had to remove the instanceOf check, because it was always failing, even when i passed in the litNodeClient.  removing this fixed it.

PR #216, opened 2023-09-14
[🫓 Bun] Unit testing for release

Ansonhkg:


joshLong145:
This is cool, and the test times are definitely faster, can we make a `intigrate/bun` and keep it updated for when we want to make the switch?

PR #219, opened 2023-09-20
[Pending] Feature/lit 1487 js sdk fix unit testing for 300cayenne

Ansonhkg:
Updated: See https://github.com/LIT-Protocol/js-sdk/pull/221 integration branch where this branch is merged into https://github.com/LIT-Protocol/js-sdk/pull/199 

# What

- Migrating network call unit tests to e2e nodejs tests
- Added a script to always look for latest contract addresses before build https://github.com/LIT-Protocol/js-sdk/blob/737432ac9099afd578925da8a402d521b2de1978/package.json#L10
- Fixed `minNodeCount` in `PKPBase` https://github.com/LIT-Protocol/js-sdk/pull/219/commits/0734ff3f36d3a9621a408839dfbbc427957b1727
- Added allowJs to `tsconfig.spec.js` cus Jest is running in TS env. https://github.com/LIT-Protocol/js-sdk/pull/219/commits/9a61eb61f2d8a3c7eb2aa0d9033781387adf2fd6
- Added a try/catch to find decimalPlaces in the access control humanizer to prevent it throwing error https://github.com/LIT-Protocol/js-sdk/pull/219/commits/471394b46a42692b530db84362c986ee33f6d5df#diff-ac4447edd4b5dbebe38c9c02d1d5b28238f3654f5da6a53f190aa40964187af8R248-R255
- Fixed multiple places to use `runSign` instead of `runLitAction` in `pkp-ethers` `pkp-client`

# Where

### `./e2e-nodejs directory`
- https://github.com/LIT-Protocol/js-sdk/tree/feature/lit-1487-js-sdk-fix-unit-testing-for-300cayenne/e2e-nodejs

#### Commands
```
  💡 Usage: yarn test:e2e:node
  
  🌍 ENVs:
      DEBUG=true yarn test:e2e:node (Enables debug mode)
      REAL_TX=true yarn test:e2e:node (Enables real tx that costs gas)

  🚩 Flags:
      --filter=<keyword> (Filters files by keyword)
      --group=<group> (Group is directory prefix, e.g., --group=pkp-ethers uses group-pkp-ethers)
```

PR #221, opened 2023-09-22
[Merge Branch] Feature/lit 1498 js sdk merge e2e nodejs branch into v3

Ansonhkg:
# What

This is a merge branch where #219 is merged into #199 

# Tests

- [x] Unit Tests
- [x] E2E NodeJS Tests

------------
# What

- Migrating network call unit tests to e2e nodejs tests
- Added a script to always look for latest contract addresses before build https://github.com/LIT-Protocol/js-sdk/blob/737432ac9099afd578925da8a402d521b2de1978/package.json#L10
- Fixed `minNodeCount` in `PKPBase` https://github.com/LIT-Protocol/js-sdk/pull/219/commits/0734ff3f36d3a9621a408839dfbbc427957b1727
- Added allowJs to `tsconfig.spec.js` cus Jest is running in TS env. https://github.com/LIT-Protocol/js-sdk/pull/219/commits/9a61eb61f2d8a3c7eb2aa0d9033781387adf2fd6
- Added a try/catch to find decimalPlaces in the access control humanizer to prevent it throwing error https://github.com/LIT-Protocol/js-sdk/pull/219/commits/471394b46a42692b530db84362c986ee33f6d5df#diff-ac4447edd4b5dbebe38c9c02d1d5b28238f3654f5da6a53f190aa40964187af8R248-R255
- Fixed multiple places to use `runSign` instead of `runLitAction` in `pkp-ethers` `pkp-client`

# Where

### `./e2e-nodejs directory`
- https://github.com/LIT-Protocol/js-sdk/tree/feature/lit-1487-js-sdk-fix-unit-testing-for-300cayenne/e2e-nodejs

#### Commands
```
  💡 Usage: yarn test:e2e:node
  
  🌍 ENVs:
      DEBUG=true yarn test:e2e:node (Enables debug mode)
      REAL_TX=true yarn test:e2e:node (Enables real tx that costs gas)

  🚩 Flags:
      --filter=<keyword> (Filters files by keyword)
      --group=<group> (Group is directory prefix, e.g., --group=pkp-ethers uses group-pkp-ethers)
```


packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -103,7 +103,7 @@ export class PKPBase<T = PKPBaseDefaultParams> {
       minNodeCount:
         prop.bootstrapUrls && prop.litNetwork == 'custom'
           ? prop.minNodeCount
-          : 6,
+          : 2,
```
joshLong145, line 106: Can this be read from the network config instead of hard coded? we then dont have to update per network

tools/scripts/utils.mjs
```diff
@@ -566,6 +611,39 @@ export const log = Object.assign(
   }
 );
 
+export const testThis = async (test) => {
```
joshLong145, line 614: can we document this a bit more? its not inutative that this is wrapping the runtime tests.

tools/scripts/utils.mjs
```diff
@@ -566,6 +611,39 @@ export const log = Object.assign(
...
+  // calculate the time it takes to run the test
+  const start = Date.now();
+
+  const { status, message } = await test.fn();
```
joshLong145, line 618: should we add a `catch` to prevent unresolved rejections?

tools/scripts/utils.mjs
```diff
@@ -566,6 +611,39 @@ export const log = Object.assign(
...
+  let errorIsThrown = false;
+
+  try {
+    const end = Date.now();
+
+    const time = end - start;
+
+    if (status === 200) {
+      log.green(`\t${message} (${time}ms)`);
+      return true;
```
joshLong145, line 629: 200 might not always mean the request was sucessful. maybe we should also check for any errors in the object?

tools/scripts/utils.mjs
```diff
@@ -665,3 +743,84 @@ export async function checkEmptyDirectories(dirPath) {
...
+) {
+  const wallet = new ethers.Wallet(privateKey);
+
+  // expirtaion time in ISO 8601 format
+  const expirationTime = new Date(
+    Date.now() + 1000 * 60 * 60 * 24 * 7
+  ).toISOString();
+
+  const domain = 'localhost';
+  const origin = 'https://localhost/login';
```
joshLong145, line 773: Should we allow defining the domain and origin?

joshLong145 (changes requested):

tools/scripts/utils.mjs
Ansonhkg: sorry should've properly documented this - it will also return `200` cus in the test we are relying on the custom `success` or `fail` function, so it will either return 200, or 500. If not, the test is not correctly implemented:

```
export const success = (message) => {
  return {
    status: 200,
    message,
  };
};

export const fail = (message) => {
  return {
    status: 500,
    message,
  };
};
```

tools/scripts/utils.mjs
Ansonhkg: it's already in a try/catch block though?

tools/scripts/utils.mjs
Ansonhkg: Agreed

packages/pkp-base/src/lib/pkp-base.ts
Ansonhkg: https://github.com/LIT-Protocol/js-sdk/blob/952f7abecdafc451599c6734e6a6a3c2bb0b975b/packages/pkp-base/src/lib/pkp-base.ts#L107

tools/scripts/utils.mjs
Ansonhkg: https://github.com/LIT-Protocol/js-sdk/blob/952f7abecdafc451599c6734e6a6a3c2bb0b975b/tools/scripts/utils.mjs#L20-L32

tools/scripts/utils.mjs
Ansonhkg: https://github.com/LIT-Protocol/js-sdk/blob/952f7abecdafc451599c6734e6a6a3c2bb0b975b/tools/scripts/utils.mjs#L783-L784

PR #225, opened 2023-10-02
add check for node runtime and import from node buffer for Blob

joshLong145:
fixes an observed issue with `encryptWithSymmetricKey` which throws an error in NodeJS when attempting to use the `Blob` object as it is not available in global scope. 

refactors to check for if the env is `node` and then imports `node:buffer` to use the compatible `Blob` object


joshLong145:
Link to ticket: https://linear.app/litprotocol/issue/LIT-628/add-attestation-to-the-typescript-sdk-during-the-handshake-process

Adarsh-Kumar28 (dismissed):
LGTM

PR #230, opened 2023-10-09
Fix/controller auth method

joshLong145:
adds suport for auth method authentication in the `pkp-base` forwarding `controllerAuthMethod` to the `pkpSign` endpoints. changes the auth checks to use a `filter` to make sure only a single method is used when forwarding to the nodes.


Ansonhkg:
Makes sense to me!

PR #235, opened 2023-10-16
fix: Validate boolean Access Control Conditions

Adarsh-Kumar28:
# What

Access Control Validation not happening due to an internal error. Hence **all** the encryption & decryptions are failing.

# How

It's a logical issues in the param-validation

# Test

Tested it by linking it with apps/nodejs


packages/encryption/src/lib/utils.ts
```diff
@@ -16,29 +18,39 @@ export function isValidBooleanExpression(
...
           return false;
         }
         currentState = STATES.OPERATOR;
     }
   }
 
-  return currentState === STATES.CONDITION;
+  const res = currentState === STATES.CONDITION;
+  console.log('isValidBooleanExpression- ', res);
+  return res;
```
Ansonhkg: Can you remove all the console logs?

Ansonhkg (changes requested):
can you remove all the console logs?

packages/encryption/src/lib/utils.ts
egeyar: I think he removed them before you could request the changes :)

PR #240, opened 2023-10-20
Fix/non block get signatures

Ansonhkg:
make `getSignatures` not to throw error, it instead will return null so that it will still combine claims and return claims 

https://github.com/LIT-Protocol/js-sdk/blob/9a0d7bb7ccdb2f9347879b4b733b7f25362d7e84/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts#L1392-L1422

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -1098,20 +1098,24 @@ export class LitNodeClientNodeJs extends LitCore {
     const validatedSignedData = signedData.map((sigObj: any) => {
 
       // -- detect whatever signature is available
-      let signature;
+      let sigShare;
       for (let key in sigObj) {
         if (sigObj[key]) {
-          signature = sigObj[key];
+          sigShare = sigObj[key];
           break;
         }
+      }
+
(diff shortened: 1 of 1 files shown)


packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -1126,7 +1130,8 @@ export class LitNodeClientNodeJs extends LitCore {
     log("minimum required length:", this.config.minNodeCount);
 
     if (validatedSignedData.length < this.config.minNodeCount) {
-      throw new Error(`not enough nodes to get the signatures.  Expected ${this.config.minNodeCount}, got ${validatedSignedData.length}`);
+      log(`not enough nodes to get the signatures.  Expected ${this.config.minNodeCount}, got ${validatedSignedData.length}`);
```
Ansonhkg, line 1133: @joshLong145 so that it could continue this line https://github.com/LIT-Protocol/js-sdk/blob/9a0d7bb7ccdb2f9347879b4b733b7f25362d7e84/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts#L1394-L1420

joshLong145 (approved):
🔥

PR #242, opened 2023-10-26
Lit 1693/add explicit stytch authentication factors

joshLong145:
- Adds new `Stytch Auth Factor` type provider, and explicit `AuthMethodTypes` for the following authentication types
- email otp
- sms otp
- WhatsApp otp
- totp

Uses a single new provider class with a unique `FactorParsers` to parse the auth factor from the session token.


Ansonhkg:
lgtm

PR #244, opened 2023-10-31
Feature/lit 1799 fix req data mapping in js sdk

joshLong145:
- Addresses an issue introduced in commit 88a5300cf3e2f48bcede8f78407b997fff556640 which would cause `422` error on requests to node endpoints if `authMethod` parameters where defined as an empty array `[]` this pr corrects mistakes made to implementation. 
- Adds test cases for using the `authMethod` parameter for auth material, and also setting it to`[]` to exercise both cases.


glitch003 (approved):
looks great, thanks for the quick fix!

PR #245, opened 2023-11-03
Feature/lit 1815 pull node urls from chain in sdk instead of using hardcoded

Ansonhkg:
## Usage
- `yarn gen:internal-dev` 
> note: the above command is added to the `yarn build` pipeline 
- API: https://lit-general-worker.getlit.dev/internal-dev-contract-addresses

### Scripts
- [package.json](https://github.com/LIT-Protocol/js-sdk/blob/2799529766c95b1b4a2c7061724705916a3aeb6c/package.json#L42)
- [tools/scripts/gen-internal-dev.mjs](https://github.com/LIT-Protocol/js-sdk/blob/feature/lit-1815-pull-node-urls-from-chain-in-sdk-instead-of-using-hardcoded/tools/scripts/gen-internal-dev.mjs)

### Output
- [packages/constants/src/lib/constants/autogen_internal.ts](https://github.com/LIT-Protocol/js-sdk/blob/feature/lit-1815-pull-node-urls-from-chain-in-sdk-instead-of-using-hardcoded/packages/constants/src/lib/constants/autogen_internal.ts)

### Integrations
- [packages/constants/src/lib/constants/constants.ts](https://github.com/LIT-Protocol/js-sdk/blob/2799529766c95b1b4a2c7061724705916a3aeb6c/packages/constants/src/lib/constants/constants.ts#L666)
- [packages/core/src/lib/lit-core.ts](https://github.com/LIT-Protocol/js-sdk/blob/2799529766c95b1b4a2c7061724705916a3aeb6c/packages/core/src/lib/lit-core.ts#L73-L82)

## Additionally
- lit-general-worker PR: https://github.com/LIT-Protocol/lit-apps/pull/8/files

## Test

1. Run `yarn build`
2. Run `yarn test:e2e:node --filter=test-connection-internal-dev`

- [e2e-nodejs/group-connection/test-connection-internal-dev.mjs](https://github.com/LIT-Protocol/js-sdk/blob/feature/lit-1815-pull-node-urls-from-chain-in-sdk-instead-of-using-hardcoded/e2e-nodejs/group-connection/test-connection-internal-dev.mjs)

```
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://199.115.117.113:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://199.115.117.113:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://199.115.117.114:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://199.115.117.114:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://108.62.0.105:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://108.62.0.105:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://64.131.85.106:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://64.131.85.106:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://64.131.85.108:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://64.131.85.108:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://167.114.17.201:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://167.114.17.201:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://167.114.17.202:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://167.114.17.202:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://167.114.17.203:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://167.114.17.203:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://167.114.17.204:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://167.114.17.204:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] handshakeWithSgx http://167.114.17.205:443/web/handshake
[Lit-JS-SDK v3.0.16] sendCommandToNode with url http://167.114.17.205:443/web/handshake and data { clientPublicKey: 'test' }
[Lit-JS-SDK v3.0.16] ✅ [BLS SDK] wasmExports loaded. 12 functions available. Run 'wasmExports' in the console to see them.
[Lit-JS-SDK v3.0.16] ✅ [ECDSA SDK NodeJS] wasmECDSA loaded. 6 functions available. Run 'wasmECDSA' in the console to see them.
Something went wrong, internal id for request: lit_c043c891323d1. Please provide this identifier with any support requests. Error is request to http://199.115.117.113:443/web/handshake failed, reason: socket hang up - undefined
[Lit-JS-SDK v3.0.16] Error connecting to node  http://199.115.117.113:443 FetchError: request to http://199.115.117.113:443/web/handshake failed, reason: socket hang up
```


joshLong145:
The idea of using the staking contracts to get the current peer set is that we can dynamically get the `active peers` when standing up an instance of the client. 
there are two operations we need to do for resolving  the active peers for the bootstrapping

1) Query the validators in current epoch from the staking contract
2) Calculate the threshold as 2/3 of the validator count
3) Query the active validators in the current epoch to get the set of urls to use
4) determine if the active validator set is at least the threshold.
5) persist the active validator urls as the bootstrap urls for the instance of the client

hwrdtm:
> The idea of using the staking contracts to get the current peer set is that we can dynamically get the `active peers` when standing up an instance of the client. there are two operations we need to do for resolving the active peers for the bootstrapping
> 
> 1. Query the validators in current epoch from the staking contract
> 2. Calculate the threshold as 2/3 of the validator count
> 3. Query the active validators in the current epoch to get the set of urls to use
> 4. determine if the active validator set is at least the threshold.
> 5. persist the active validator urls as the bootstrap urls for the instance of the client

Do we also need to understand which nodes are kicked from the next node so that the SDK can exclude sending messages to them?

glitch003:
this looks great to me.  yeah @hwrdtm - i think you're right, we should remove the kicked validators?  we can use this function to get kicked validators and then filter them from the set.  https://github.com/LIT-Protocol/lit-assets/blob/431bed455e4bff990dd42a01ee21ecd4c8b60227/blockchain/contracts/contracts/lit-node/Staking/StakingViewsFacet.sol#L287

joshLong145:
> this looks great to me. yeah @hwrdtm - i think you're right, we should remove the kicked validators? we can use this function to get kicked validators and then filter them from the set. https://github.com/LIT-Protocol/lit-assets/blob/431bed455e4bff990dd42a01ee21ecd4c8b60227/blockchain/contracts/contracts/lit-node/Staking/StakingViewsFacet.sol#L287



> https://github.com/LIT-Protocol/lit-assets/blob/431bed455e4bff990dd42a01ee21ecd4c8b60227/blockchain/contracts/contracts/lit-node/Staking/StakingViewsFacet.sol#L287

Has been added to the linear task
https://linear.app/litprotocol/issue/LIT-1815/pull-node-urls-from-chain-in-sdk-instead-of-using-hardcoded-values

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -357,137 +358,225 @@ export class LitContracts {
...
+      console.log('❌ Active validator set does not meet the threshold');
+    }
+
+    // remove kicked validators in active validators
+    const cleanedActiveValidators = activeValidators.filter(
+      (av: any) => !kickedValidators.some((kv: any) => kv.nodeAddress === av.nodeAddress)
+    );
+
+    const networks = cleanedActiveValidators.map((item: any) => {
+      return `http://${intToIP(item.ip)}:${item.port}`;
```
joshLong145, line 574: i dont think we want to hardcode `http://` as prod nodes will run in `https` but im unsure how we will know as if we attempt to connect over http on any of the ports it will fail i think

packages/core/src/lib/lit-core.ts
```diff
@@ -136,17 +152,87 @@ export class LitCore {
...
-   * Connect to the LIT nodes
-   *
-   * @returns { Promise } A promise that resolves when the nodes are connected.
-   *
+   * @returns {Promise<void>} A promise that resolves when the listener is successfully set up.
+   */
+  listenForNewEpoch = async (): Promise<void> => {
+    const stakingContract = await LitContracts.getStakingContract(this.config.litNetwork as LitNetwork);
+
+    stakingContract.on("StateChanged", async () => {
```
joshLong145, line 206: I dont know if we want to reload config on every state change
```
    enum States {
        Active,
        NextValidatorSetLocked,
        ReadyForNextEpoch,
        Unlocked,
        Paused,
        Restore
    }
```
I think `nextValidatorSetLocked` will be a good state to check for as then we know the next epoch set is locked in. but this might mean that we should check if the `ValidatorsStructsInCurrentEpoch` are set before attempting to reload the config as we might still be performing dkg and not ready to switch them out.

packages/core/src/lib/lit-core.ts
```diff
@@ -203,8 +204,11 @@ export class LitCore {
   listenForNewEpoch = async (): Promise<void> => {
     const stakingContract = await LitContracts.getStakingContract(this.config.litNetwork as LitNetwork);
 
-    stakingContract.on("StateChanged", async () => {
-      await this.setNewConfig();
+    stakingContract.on("StateChanged", async (state: StakingStates) => {
+      log(`New state detected: "${state}"`);
+      if (state === StakingStates.NextValidatorSetLocked) {
+        await this.setNewConfig();
+      }
```
joshLong145, line 211: nice, @Adarsh-Kumar28 is there any other state transitions where we would want to re check the validators?

joshLong145:
Thanks for this work. closing as the implementations have been migrated to: #272

PR #246, opened 2023-11-03
fix encryptFileAndZipWithMetadata node

vmc-crossmint:
`encryptFileAndZipWithMetadata` was not working for the @lit-protocol/lit-node-client-nodejs version of the sdk.

Presumably the `zipBlob = await zip.generateAsync({ type: 'blob' })` works properly only in browser env.

- The function now returns a `nodebuffer` instead of `blob` when not on browser env.
- The file validation function now accepts a 'Uint8array' and not only the client side 'Blob','File' obj

Node v18.16, tested both on next.js and standalone script

### DIFF
--- a/packages/encryption/src/lib/encryption.ts
+++ b/packages/encryption/src/lib/encryption.ts
@@ -571,7 +571,12 @@ export const encryptFileAndZipWithMetadata = async (
 
   folder.file(file.name, uint8arrayFromString(ciphertext, 'base64'));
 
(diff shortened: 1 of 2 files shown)


glitch003:
this looks good to me.  @Ansonhkg is this okay to merge?

joshLong145:
@glitch003 looks good to me, will approve.

joshLong145 (approved):

PR #248, opened 2023-11-06
feat: Store blockhash from handshake + use it in SIWE

Adarsh-Kumar28:
# What

Stores `latest_blockhash` returned by the nodes in the handshake to the `litNodeClient` instance. This is used as the nonce when crafting the SIWE message for signing in both `AuthSig` & `SessionSigs`.

This PR is **backwards-compatible** i.e. devs can still choose not to provide the `nonce` in the updated signing functions and I've also taken care of the fact the node currently don't return `latest_blockhash`.

# How

Updated the following functions to accept the nonce which can be fetched from stored in the client using the `getLatestBlockhash()`:

- `checkAndSignAuthMessage`
- `checkAndSignEVMAuthMessage`
- `signAndSaveAuthMessage`
- `_signAndGetAuth`

```
const litNodeClient = new LitJsSdk.LitNodeClient({
  litNetwork: "cayenne",
});
await litNodeClient.connect();
const nonce = litNodeClient.getLatestBlockhash();
const authSig = await LitJsSdk.checkAndSignAuthMessage({ chain, nonce });
```

Since the SDK receives the `latest_blockhash` from all the nodes they could be different. Hence the final one to be stored is determined by using `mostCommonString()` on all the returned hashes. This is similar to all the other node returned values in the handshake like `hdRootPubkeys`.

**Note:** When the client is not provided in the above functions or if the `latest_blockhash` is undefined (since the corresponding node PR will be merged after this one) then a random nonce is generated by `generateNonce()`.

## LitAuthClient

Have also updated `authenticate()` to use the `latest_blockhash` for `EthWalletProvider` nonce.

## Expiry

Have verified that we use `Date.now()` in the expiry instead of `issueAt`

# Tests

Added manual tests for both the scenarios when the nonce is provided and when it isn't.

### DIFF
--- a/apps/html/manual_tests.html
+++ b/apps/html/manual_tests.html
@@ -21,22 +21,47 @@
 
       //var authSig = JSON.parse("{\"sig\":\"0x18a173d68d2f78cc5c13da0dfe36eec2a293285bee6d42547b9577bf26cdc985660ed3dddc4e75d422366cac07e8a9fc77669b10373bef9c7b8e4280252dfddf1b\",\"derivedVia\":\"web3.eth.personal.sign\",\"signedMessage\":\"I am creating an account to use LITs at 2021-08-04T20:14:04.918Z\",\"address\":\"0xdbd360f30097fb6d938dcc8b7b62854b36160b45\"}")
 
-      var CheckAndSignAuthMessageWorksWithoutBlockhash = async() => {
+      var CheckAndSignAuthMessageWithoutBlockhash = async() => {
         const authSig = await LitJsSdk.checkAndSignAuthMessage({
             chain: "ethereum",
         });
+        let nonceLine = authSig.signedMessage.match(/^Nonce: (.+)$/m);
+
+        if (nonceLine[1].length ===  17) {
+          document.getElementById(
+            'status'
+          ).innerText = "CheckAndSignAuthMessageWithoutBlockhash passed!";
+          return true;
+        } else {
+          document.getElementById('status').innerText = "CheckAndSignAuthMessageWithoutBlockhash failed!";
+          return false;
+        }
+
+        await LitJsSdk.disconnectWeb3();
       }
 
       // The nonce should be we the below blockhash
-      var CheckAndSignAuthMessageWorksWithBlockhash = async() => {
+      var CheckAndSignAuthMessageWithBlockhash = async() => {
         // Currently this will be undefined but later we should replace it with the hardcoded blockhash
-        const nonce = litNodeClient.getLatestBlockhash() 
-        || "0xfe88c94d860f01a17f961bf4bdfb6e0c6cd10d3fda5cc861e805ca1240c58553";
+        const TEST_BLOCKHASH = "0xfe88c94d860f01a17f961bf4bdfb6e0c6cd10d3fda5cc861e805ca1240c58553";
+        const nonce = litNodeClient.getLatestBlockhash() || TEST_BLOCKHASH;
 
         const authSig = await LitJsSdk.checkAndSignAuthMessage({
             chain: "ethereum",
             nonce,
         });
+
+        if (authSig.signedMessage.includes(TEST_BLOCKHASH)) {
+          document.getElementById(
+            'status'
+          ).innerText = "CheckAndSignAuthMessageWithBlockhash passed!";
+          return true;
+        } else {
+          document.getElementById('status').innerText = "CheckAndSignAuthMessageWithBlockhash failed!";
+          return false;
+        }
+
+        await LitJsSdk.disconnectWeb3();
       }
 
       var testEncryptingFileAndZipWithMetadata = async ({
@@ -1582,10 +1607,10 @@
     <div id="networkStatus">Connecting to Lit Protocol Network...</div>
     <br />
     <br />
-    <button onclick="CheckAndSignAuthMessageWorksWithoutBlockhash()">checkAndSignAuthMessage without blockhash nonce</button>
+    <button onclick="CheckAndSignAuthMessageWithoutBlockhash()">checkAndSignAuthMessage without blockhash</button>
     <br />
     <br />
-    <button onclick="CheckAndSignAuthMessageWorksWithBlockhash()">checkAndSignAuthMessage with blockhash nonce</button>
+    <button onclick="CheckAndSignAuthMessageWithBlockhash()">checkAndSignAuthMessage with blockhash</button>
     <br />
     <br />
     <button onclick="ETHEmptyWallet()">ETH (Empty Wallet)</button>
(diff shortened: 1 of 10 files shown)


packages/auth-browser/src/lib/auth-browser.ts
```diff
@@ -44,14 +46,23 @@ export const checkAndSignAuthMessage = ({
...
   if (chainInfo.vmType === VMTYPE.EVM) {
+    // TODO: Uncomment below after the new node changes are deployed
+    // if (!litNodeClient) {
+    //   throwError({
+    //     message: 'litNodeClient not provided for EVM signing',
+    //     errorKind: LIT_ERROR.LIT_NODE_CLIENT_NOT_PROVIDED.kind,
+    //     errorCode: LIT_ERROR.LIT_NODE_CLIENT_NOT_PROVIDED.name,
+    //   });
+    // }
+
```
joshLong145, line 57: Is there a task for this TODO?

packages/auth-browser/src/lib/chains/eth.ts
```diff
@@ -726,7 +730,8 @@ export const signAndSaveAuthMessage = async ({
   resources,
   expiration,
   uri,
-}: signAndSaveAuthParams): Promise<AuthSig> => {
+}: signAndSaveAuthParams,
```
joshLong145, line 733: not sure why we need to inject the client instance. cant the sdk and nodes independently query the block height? requiring a client instance for this method is a bit of a heavy lift and does not seem needed for simply querying the block height has this can be done with a scoped rpc client in `signAndSaveAuthMessage`

joshLong145 (changes requested):

packages/auth-browser/src/lib/auth-browser.ts
Adarsh-Kumar28: Good idea lemme create one on Linear and link it here

Adarsh-Kumar28:
Closing this PR for a new PR: https://github.com/LIT-Protocol/js-sdk/pull/260

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2273,6 +2280,7 @@ export class LitNodeClientNodeJs extends LitCore {
         params.resourceAbilityRequests.map((r) => r.resource)
       );
     let expiration = params.expiration || LitNodeClientNodeJs.getExpiration();
+    let nonce = this.latest_blockhash || generateNonce();
```
joshLong145, line 2283: isnt the point to not generate a local nonce? if the plan is to remove this once the nodes have merged support for this then we should have a task to follow up and throw an error if the `latest_blockhash` is undefined. Otherwise the work on the nodes can be bypassed.

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2103,6 +2109,7 @@ export class LitNodeClientNodeJs extends LitCore {
       chainId: params.chainId ?? 1,
       expirationTime: _expiration,
       resources: params.resources,
+      nonce: this.latest_blockhash || generateNonce(),
```
joshLong145, line 2112: where is `generateNonce` defined?

apps/html/manual_tests.html
```diff
@@ -21,6 +21,24 @@
...
+      // The nonce should be we the below blockhash
+      var CheckAndSignAuthMessageWorksWithBlockhash = async() => {
+        // Currently this will be undefined but later we should replace it with the hardcoded blockhash
+        const nonce = litNodeClient.getLatestBlockhash() 
+        || "0xfe88c94d860f01a17f961bf4bdfb6e0c6cd10d3fda5cc861e805ca1240c58553";
+
+        const authSig = await LitJsSdk.checkAndSignAuthMessage({
+            chain: "ethereum",
+            nonce,
+        });
```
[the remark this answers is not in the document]
Adarsh-Kumar28: Will update the document status based on the result

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Adarsh-Kumar28: Exactly, here's the ticket: https://linear.app/litprotocol/issue/LIT-1935/update-sdk-to-throw-error-if-blockhash-form-nodes-is-undefined

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Adarsh-Kumar28: From lit-siwe it's imported above

joshLong145:
I think we should allow a `Wallet` or `Signer` to be passed to `checkAndSignAuthMessage` so end users don't have to go out of the sdk to sign if not using our sdk in a browser where a wallet extension is installed. This would also allow us to test outside of the browser tests.
End users who are building in Nodejs wont have a browser extension and will need to have a way to get the `block hash` from the node handshake into their siwe messages or their experience will be much worse than users who are building web applications using a browser extension wallet to sign messages.

Adarsh-Kumar28:
> I think we should allow a `Wallet` or `Signer` to be passed to `checkAndSignAuthMessage` so end users don't have to go out of the sdk to sign if not using our sdk in a browser where a wallet extension is installed.

This would fundamentally change how `checkAndSignAuthMessage` is supposed to work. It's only meant for the browser.

Adarsh-Kumar28:
> End users who are building in Nodejs wont have a browser extension and will need to have a way to get the `block hash` from the node handshake into their siwe messages or their experience will be much worse than users who are building web applications using a browser extension wallet to sign messages.

Why would it be much worse? That's how they create an authSig by using hot wallet signing. Also to do anything meaningful with Lit you gotta connect to the network.

I've added an **E2E Node test** to demonstrate that the workflow for fetching the latest blockhash from the client remains the same irrespective of ones development environment

joshLong145:
> > End users who are building in Nodejs wont have a browser extension and will need to have a way to get the `block hash` from the node handshake into their siwe messages or their experience will be much worse than users who are building web applications using a browser extension wallet to sign messages.
> 
> Why would it be much worse? That's how they create an authSig by using hot wallet signing. Also to do anything meaningful with Lit you gotta connect to the network.
> 
> I've added an **E2E Node test** to demonstrate that the workflow for fetching the latest blockhash from the client remains the same irrespective of one's development environment

> > End users who are building in Nodejs wont have a browser extension and will need to have a way to get the `block hash` from the node handshake into their siwe messages or their experience will be much worse than users who are building web applications using a browser extension wallet to sign messages.
> 
> Why would it be much worse? That's how they create an authSig by using hot wallet signing. Also to do anything meaningful with Lit you gotta connect to the network.
> 
> I've added an **E2E Node test** to demonstrate that the workflow for fetching the latest blockhash from the client remains the same irrespective of one's development environment

```
> Why would it be much worse? That's how they create an authSig by using hot wallet signing. Also to do anything meaningful with Lit you gotta connect to the network.
```

I am not sure what you mean. singing an auth signature with a browser extension wallet is not the same as connecting to our network of nodes. As we support other authentication steps. If you mean from the perspective of needing an auth signature then this is true. but we are analogous to how that is generated. hence supporting both wallet extensions and programmatically generated wallets is required for the network. 

I think with what you added to `e2e-nodejs` this is sufficient for the programmatic wallet case. Could this be added to some docs or made visible to end users so they know when the feature is turned on how to generate proper nonce values?

Adarsh-Kumar28:
Sure we're gonna updates the docs once the node code is deployed and the nodes start sending the blockhashes in the handshake

joshLong145 (approved):

PR #249, opened 2023-11-07
Fix/executeJS array normalizing

joshLong145:
Fixes an issue with field normalizing to a generic `Array` type where `jsParams` might contain a different property than `toSign` for passing data for signature requests.
Will now loop through keys on the `jsParam` object checking for `Array` and `TypedArray` instance types to determine if the property should be normalized to a basic `Array` instance


glitch003:
@joshLong145 looks good!  does this work if jsParams is undefined because they didn't pass any?  we could add a default of {} if needed.

glitch003 (approved):
approved pending a test that it works when jsParams isn't passed in

joshLong145:
> approved pending a test that it works when jsParams isn't passed in

test has been added, `Object.keys` does actually throw an Error if it is given an undefined value. updated to use an empty object which then does not iterate. test case is added to cover this case

PR #250, opened 2023-11-07
Feature/lit 1765 js sdk provider not set in pkpethers

Ansonhkg:
fix & improve pkp-ethers to set custom provider and allow contract calls within


glitch003 (approved):
looks good

PR #253, opened 2023-11-09
mint by auth method

Ansonhkg:
Introduced an abstracted function to mint a pkp by auth method and allowing user to set permitted scopes with useful error reporting when scopes is not defined.


e2e-nodejs/group-contracts/test-contracts-write-mint-a-pkp-and-set-scope-1-2-easy.mjs
```diff
@@ -0,0 +1,72 @@
...
+
+  // ==================== LitContracts Setup ====================
+  const contractClient = new LitContracts({
+    signer: controllerWallet,
+  });
+
+  await contractClient.connect();
+
+  // ==================== Test Logic ====================
+  const mintInfo = await contractClient.mint({
```
joshLong145, line 27: Maybe we add a new namespace for higher level utility methods since we namepsace by contract name. We should call this method some other than mint as we are combining minting and permissions which is important for this function name to indicate

joshLong145 (approved):
🔥 thanks for adding this. 
Left a comment on the naming.

PR #257, opened 2023-11-14
make a change to claim and mint to add in tx options

VoR0220:
I came into this as a roadblock and not being able to get my claim through. I figure the tx object should still be available in the helper function and should overwrite the cost to mint, but could probably figure out a way to make cost to mint as the default in the object. Either way. Would appreciate a look and a merge.


Ansonhkg:
Awesome! This LGTM @joshLong145

joshLong145 (approved):
Looks good, with this change this allows for transaction data to be set by caller. this does mean if using `txOpts` you do have to call `mintCost` externally. but i think this is fine. Thank you for the contribution 🔥

PR #260, opened 2023-11-16
feat: Use latest eth blockhash for authsig nonce

Adarsh-Kumar28:
# What

Fetches the latest Eth blockhash from the Ethereum RPCs & use it as the nonce for AuthSig & sessionSig signing.

# How

Added 3 more Ethereum RPCs and a try mechanism to fetch the latest Eth blockhash. Once we get that we use it for the `nonce` field.

## LitAuthClient

Have also updated `authenticate()` to use the `latest_blockhash` for `EthWalletProvider` nonce.


packages/constants/src/lib/utils/utils.ts
```diff
@@ -28,3 +30,26 @@ export function ERight<T>(result: T): IEither<T> {
...
+    } catch (e: any) {
+      console.error(
+        'Error getting the latest Eth blockhash- ' + i + ', trying again: ',
+        e
+      );
+    }
+  }
+
+  throw new Error('Unable to get the latestBlockhash from any RPCs');
+}
```
Ansonhkg, line 55: 1. initialise provider inside loop to ensure each rpc url is used when attempting to fetch the latest block
2. start loop from index 0

Ansonhkg (commented):
demo/e2e test app required to be tested

packages/constants/src/lib/utils/utils.ts
Adarsh-Kumar28: The first RPC isn't working properly & always returns the same nonce

Adarsh-Kumar28: Fixed the provider new init. Also written a comment on the Ethereum RPCs

Adarsh-Kumar28:
> demo/e2e test app required to be tested
@Ansonhkg 

For testing that the nonce is indeed the latest Ethereum blockhash you may use: 
1. **checkAndSignAuthMessage:** demo-encrypt-decrypt-react
2. **signSessionKey:** demo-pkp-social-auth-next-ts
3. **getSessionSigs:** demo-pkp-session-sigs

joshLong145:
is there a reason we cannot add tests to `e2e` so we can run them in test runners? would be easier than having to build demo apps to run sanity tests.

packages/constants/src/lib/utils/utils.ts
```diff
@@ -28,3 +30,27 @@ export function ERight<T>(result: T): IEither<T> {
     result,
   };
 }
+
+export async function getLatestEthBlockhash(): Promise<string> {
+  // Not using the first rpc as it always returns the same nonce. Seems like the RPC just returns a dummy value
+  for (let i = 1; i < LIT_EVM_CHAINS['ethereum'].rpcUrls.length; i++) {
+    const provider = new ethers.providers.JsonRpcProvider(
```
joshLong145, line 37: nit: move the loop for the rpc provider into the try catch

packages/constants/src/lib/utils/utils.ts
Ansonhkg: re: The first RPC isn't working properly & always returns the same nonce

Can you change the RPC value in index 0 so that it won't return the same nonce. we need to start the loop index from 0

Ansonhkg: also can we move the `getLatestEthBlockhash` function to `auth-browser`. It's not a constant so it shouldn't be in the `constants` package

Ansonhkg (changes requested):
we also need e2e coverage for the `getLatestEthBlockhash` function, so that we know all rpcs are working, and change them if needed.

Adarsh-Kumar28:
Closing this PR for https://github.com/LIT-Protocol/js-sdk/pull/248

PR #265, opened 2023-11-20
Add/logger

joshLong145:
- Adds a new logging package with support for
    - Levels
    - Timing
    - Colors
    - categories
    - TimeStamps
    - Trace Id's
    - forwarding to aggregates for telemetry storage

Integrated into `core` by injecting the `log manager` and a single `logger` into `lit-core` by global injection

## Examples
Using the `LogManager` to create a new logger
```
const logger = LogManager.Instance.get('my-logger-category');
```

Using the `LogManager` to set logging levels for all loggers
```
LogManager.Instance.setLevel(LogLevel.TRACE);
```

Adding a `handler` for all loggers
```
LogManager.setHandler((log) => {
  console.log("new log ", log);
});
```

The above can be done on individual loggers for single instance modifications

## Next Steps 
- support log levels where we can turn `debug` option to a `log level` option for finer grade log support.
- Add forwarding to central logging aggregate


Ansonhkg (commented):
lgtm. we should make this a standard for logging across all our packages and PRs.

PR #274, opened 2023-12-01
Add/dynamic contract resolving jit test pkps

joshLong145:
# WHAT
- Adds dynamic resolving from the `https://lit-general-worker.getlit.dev` which allows for runtime switching of network contexts.
     - A follow up task after this merging is remove the autogen `data` files from the contract code gen as we no longer need `address` / `abi` context to be built at compile time.
 - Adds further test context to `00-setup.mjs` to allow for JIT generation of test data (auth signatures, pkp's, auth methods, etc).
 - Some clean up to logging
 - Slight tweaks to `connect` to detect if the network context is `cayenne` as we cannot pull validator information for internal test networks from chain.
 
 TODO:
 Once `mainnet` contract functionality is stable we can probably remove the worker proxy as we will only need to update contract addresses for our `ContractResolver` to pull the correct network, and abis should in theory be stable at this point.


glitch003 (approved):
lgtm!

PR #275, opened 2023-12-05
Ref/small fixes

joshLong145:
Follow up pr with fixes for dynamic network resolving
- Adds more JIT data for exercising test scenarios 
- PKP with auth method (type 1 `eth wallet signature`)

changes to session signature tests to use JIT pkp with auth method


Ansonhkg:
lgtm

PR #288, opened 2023-12-19
fix e2e tests for blockhash

joshLong145:
- Fixes an issue with authentication signatures when `latestBlockhash` is enabled.


joshLong145:
Closing and moving to pr: #288

PR #292, opened 2023-12-19
feat: Always use the blockhash as the nonce instead of gen randomly

Adarsh-Kumar28:
# Description

Remove generating the nonce randomly & use only the onces returned by the nodes in the handshake as Cayenne nodes were updated

## Type of change

# How Has This Been Tested?

I've updated the E2E node tests to test creating a sessionSig using hot-wallet signing. If fails if the nonce isn't a valid Eth blockhash.

# Checklist:

- [X] I have made corresponding changes to the documentation
- [X] Any dependent changes have been merged and published in downstream modules


packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2386,7 +2386,7 @@ export class LitNodeClientNodeJs extends LitCore {
           params.resourceAbilityRequests.map((r) => r.resource)
         );
     let expiration = params.expiration || LitNodeClientNodeJs.getExpiration();
-    let nonce = this.latestBlockhash || generateNonce();
+    let nonce = this.latestBlockhash!;
```
joshLong145, line 2389: can we add a check here that `nonce` is defined?  This situation shouldnt happen as we set it in the `handshake` but just in case we shouldnt sign the message unless it is defined in this scope.

packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
```diff
@@ -92,7 +92,7 @@ export default class EthWalletProvider extends BaseProvider {
     } else {
       authSig = await checkAndSignAuthMessage({
         chain,
-        nonce: this.litNodeClient.latestBlockhash || generateNonce(),
+        nonce: this.litNodeClient.latestBlockhash!,
```
joshLong145, line 95: Same here, we should check that `latestBlockhash` is defined before  attempting to sign the siwe

e2e-nodejs/group-hotwallet-with-blockhash/test-hotwallet-with-blockhash.mjs
```diff
@@ -14,58 +27,88 @@ export async function main() {
...
+        ability: LitAbility.AccessControlConditionDecryption,
+      },
+    ],
+    authNeededCallback,
+  });
 
-  if (!authSig.signedMessage.includes(TEST_BLOCKHASH)) {
-    return fail("authSig doesn't contain the blockhash");
-  }
+  console.log(sessionSigs);
```
joshLong145, line 111: do we do anything with the `sessionSigs` beside log it? we can remove this if we are not doing assertions.

joshLong145 (changes requested):

joshLong145:
@Adarsh-Kumar28 
It looks like your pr brings in the changes made in #288 can you merge this pr into yours so it can be closed?

e2e-nodejs/group-hotwallet-with-blockhash/test-hotwallet-with-blockhash.mjs
Adarsh-Kumar28: No but the idea is to check whether the sessionSigs is created correctly

joshLong145: Right, can we add some assertions thats the `latestBlockhash` is included in the objects contained in the `sessionSigsMap`

joshLong145 (changes requested):

packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
```diff
@@ -90,9 +90,17 @@ export default class EthWalletProvider extends BaseProvider {
         address: address,
       };
     } else {
+      if (!this.litNodeClient.latestBlockhash) {
+        throwError({
+          message: 'Eth Blockhash is undefined.',
```
joshLong145, line 95: can you make this  more descriptive of why the block hash is undefined? since this is implicitly loaded from `handshake` the end user doesn't have a lot of options.

packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
```diff
@@ -74,7 +74,7 @@ export default class EthWalletProvider extends BaseProvider {
         version: '1',
         chainId,
         expirationTime: expiration,
-        nonce: this.litNodeClient.latestBlockhash || generateNonce(),
+        nonce: this.litNodeClient.latestBlockhash!,
```
joshLong145, line 77: can we  first check if this is defined? otherwise the `nonce` will be `undefined`

packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
Adarsh-Kumar28: Added a message to reconnect

joshLong145 (approved):

PR #311, opened 2024-01-19
Feature/lit 2183 error with checkandsignauthmessage

joshLong145:
# Description

fixes issue: [circular reference with checkAndSignAuthMessage](https://linear.app/litprotocol/issue/LIT-2183/error-with-checkandsignauthmessage)
caused from a complex `provider` object being handed to the logging implementation which threw an exception when attempting to serialize circular references.
A new test has been added to exercise serializing circular objects where we can observe the properties which cause circular referencing are removed when stored.

## Type of change

# How Has This Been Tested?
A new test has been added to `packages/logger/logger.spec.ts`

# Checklist:

- [] I have made corresponding changes to the documentation


packages/auth-browser/src/lib/chains/eth.ts
```diff
@@ -389,7 +389,7 @@ export const connectWeb3 = async ({
 
   const provider = await dialog.getWalletProvider();
 
-  log('got provider', provider);
+  log('got provider');
```
Ansonhkg, line 392: why did we remove the `provider` var in the log?

packages/logger/src/lib/logger.ts
```diff
@@ -89,12 +89,20 @@ function _resolveLoggingHandler(level: LogLevel): any {
...
+    obj,
+    (key, value) =>
+      typeof value === 'object' && value !== null
+        ? cache?.includes(value)
+          ? undefined // Duplicate reference found, discard key
+          : cache?.push(value) && value // Store value in our collection
+        : value,
+    indent
+  );
+  cache = null;
```
Ansonhkg, line 104: can we add some comments showing what this is doing?

Ansonhkg (changes requested):

packages/auth-browser/src/lib/chains/eth.ts
joshLong145: the provider is a complex object with multiple circular references within. logging it seems unnecessary so i removed it. Since it is of type `{unknown}` we cannot know what properties are on this object so i removed the logging of it all together.

Ansonhkg (approved):
lgtm

PR #316, opened 2024-01-20
feature/lit-2303-js-sdk-authmethod-is-a-sessionsig-if-so-parse-one-of-the

Ansonhkg:
# Description

@Adarsh-Kumar28: "
There seems to be an issue with using sessionSigs for in the contractClient.mintWithAuth() function
It throws an error: address.toLowerCase is not a function
I investigated it and the bug is that it expects the authMethod to be an authSig: address = JSON.parse(authMethod.accessToken).address;
https://github.com/LIT-Protocol/js-sdk/blob/master/packages/contracts-sdk/src/lib/auth-utils.ts#L37
The fix is check whether the AuthMethod is a sessionSig if so parse one of the elements of it's array"

## Type of change

# How Has This Been Tested?

- [x] Unit test

# Checklist:

### DIFF
--- a/packages/contracts-sdk/src/lib/auth-utils.ts
+++ b/packages/contracts-sdk/src/lib/auth-utils.ts
@@ -34,14 +34,16 @@ export async function getAuthIdByAuthMethod(authMethod: any): Promise<string> {
 
 // check if the given object is a session sigs object
 export const isSessionSigs = (obj: any) => {
-
   // the "top-level" object must have a key that starts with "https://"
-  const hasUrlKey = Object.keys(obj).every((key) => key.includes('https://'));
-
   // one of the objects inside the session sigs must have the "algo" key
-  const hasAlgoKey = Object.values(obj).every((sig: any) => sig['algo']);
+  // -- check if key is a url and if the object has an algo key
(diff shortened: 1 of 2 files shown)


packages/contracts-sdk/src/lib/auth-utils.ts
```diff
@@ -30,15 +31,46 @@ export async function getAuthIdByAuthMethod(authMethod: any): Promise<string> {
...
-function getEthAuthMethodId(authMethod: any): string {
+
+// check if the given object is a session sigs object
+export const isSessionSigs = (obj: any) => {
+
+  // the "top-level" object must have a key that starts with "https://"
+  const hasUrlKey = Object.keys(obj).every((key) => key.includes('https://'));
+
+  // one of the objects inside the session sigs must have the "algo" key
+  const hasAlgoKey = Object.values(obj).every((sig: any) => sig['algo']);
```
joshLong145, line 42: nit: I think we could do this with a single `for in` loop
```
for (const key in obj) {
  if (!key.includes('https://') || !obj[key]['algo']) { 
     return false;
  }
}
```

joshLong145 (approved):

PR #320, opened 2024-01-22
Feature/lit 2096 add json schema runtime assertions for access control

FedericoAmura:
# Description

This PR adds schema validation to access control conditions on PKP. The added schemas are also used to define the TS classes that are used on several places, so PR also adds scripts to generate those classes (if schemas are modified) and tests to validate we have those in sync

## Type of change

# How Has This Been Tested?

- [x] Success e2e tests on ACC when encrypting
- [x] Failing e2e tests on ACC when encrypting
- [x] Modifying JSON schemas manually and running sync validation unit test

# Checklist:


packages/encryption/src/lib/params-validators.ts
```diff
@@ -532,43 +534,148 @@ class AccessControlConditionsValidator implements ParamsValidator {
...
 
     return ERight(undefined);
   }
+
+  // Filters an array of things that can themselves be nested arrays
+  // Result is a flat array of all the things that pass the filter
+  private flattenAndFilter(
+    values: any[],
+    filter: (value: any) => boolean
+  ): any[] {
```
joshLong145, line 670: can this be strongly typed?

packages/encryption/src/lib/params-validators.ts
```diff
@@ -532,43 +534,148 @@ class AccessControlConditionsValidator implements ParamsValidator {
...
+    }
 
     return ERight(undefined);
   }
+
+  // Filters an array of things that can themselves be nested arrays
+  // Result is a flat array of all the things that pass the filter
+  private flattenAndFilter(
+    values: any[],
+    filter: (value: any) => boolean
```
joshLong145, line 669: can `value` be strongly typed?

packages/encryption/src/lib/schemas/LPACC_ATOM.json
```diff
@@ -0,0 +1,44 @@
+{
+  "$id": "https://litprotocol.com/schemas/LPACC_EVM_ATOM.json",
```
joshLong145, line 2: This will not work as a schema hosting platform. can we instead host schemas in a separate repo and link explicitly to that through github.

packages/encryption/src/lib/schemas/index.spec.ts
```diff
@@ -0,0 +1,73 @@
+import * as fs from 'fs';
```
joshLong145, line 1: see other comments about moving implementation for schema validation outside of this package.

packages/encryption/src/lib/schemas/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { LIT_ERROR } from '@lit-protocol/constants';
+import { throwError } from '@lit-protocol/misc';
+import { ConditionType } from '@lit-protocol/types';
+
+export const SCHEMAS: { [K in ConditionType]: JSONSchemaType<any> } = {
+  cosmos: require('./LPACC_ATOM.json'),
+  evmBasic: require('./LPACC_EVM_BASIC.json'),
+  evmContract: require('./LPACC_EVM_CONTRACT.json'),
+  solRpc: require('./LPACC_SOL.json'),
```
joshLong145, line 11: nit: can we load these async? using require seems a bit dangerous. Perhaps we use `import` and `Promise.all` to concurrently load schemas instead.

packages/encryption/src/lib/schemas/index.ts
```diff
@@ -0,0 +1,28 @@
...
+};
+
+export const getSchema = (
+  conditionType: ConditionType
+): JSONSchemaType<any> => {
```
joshLong145, line 16: do we need a generic here if we are just saying `any` maybe we can restrict to the known schema types.

packages/encryption/tools.mjs
```diff
@@ -0,0 +1,29 @@
+import { exit } from 'process';
```
joshLong145, line 1: see comments on moving implementations out of this package.

packages/misc/src/lib/misc.ts
```diff
@@ -410,6 +411,48 @@ export const checkType = ({
...
+ * @returns { Boolean } true/false
+ */
+export const checkSchema = (
+  value: any,
+  schema: JSONSchemaType<any>,
+  paramName: string,
+  functionName: string,
+  throwOnError = true
+) => {
+  const ajv = new Ajv();
```
joshLong145, line 433: nit: this looks like a function that is called many times. Can we lazy load `ajv` such that we do not keep creating new instances each time? the implementation looks to be non mutative.

joshLong145 (commented):
Overall the implementation looks nice!
I think we should move the validation logic out of `packages/encryption` and into a seperate package. We can call it `packages/access-control-validation`

For the actual schema definitions. I think we should make a separate repo for these and keep the build scripts for the type generation in this new repo. This will allow us to independetly version schema files separately from our types package. 

@Ansonhkg what do you think about the above?

Ansonhkg:
> Overall the implementation looks nice! I think we should move the validation logic out of `packages/encryption` and into a separate package. We can call it `packages/access-control-validation`
> 
> For the actual schema definitions. I think we should make a separate repo for these and keep the build scripts for the type generation in this new repo. This will allow us to independently version schema files separately from our types package.
> 
> @Ansonhkg what do you think about the above?

I'm aligned with @joshLong145 as we've discussed on slack huddle

packages/encryption/README.md
```diff
@@ -16,3 +16,11 @@ yarn add @lit-protocol/encryption
...
 </script>
 ```
+
+### Update TS interfaces
+
+After updating the JSON schema definitions for any access control conditions, run the following command to update the TS interfaces:
+
+```bash
+node ./packages/encryption/tools.mjs --gen
+```
```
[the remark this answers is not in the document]
Ansonhkg: +1

packages/encryption/gen-interfaces.mjs
```diff
@@ -0,0 +1,37 @@
+import { exit } from 'process';
```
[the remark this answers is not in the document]
Ansonhkg: fyi. @FedericoAmura we have all the gen scripts in `./tools/scripts/`

packages/encryption/src/lib/schemas/LPACC_ATOM.json
Ansonhkg: we can also use this `$id` for versioning eg. `LPACC_EVM_ATOM_V1.json`

packages/encryption/src/lib/params-validators.ts
FedericoAmura: Typed this with a generic so we can reuse the function later

packages/encryption/gen-interfaces.mjs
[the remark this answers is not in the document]
FedericoAmura: I see now, thanks (except for [`gen-code.mjs`](https://github.com/LIT-Protocol/js-sdk/blob/master/packages/contracts-sdk/gen-code.mjs) which is the one that I used  for reference...)

FedericoAmura:
> > Overall the implementation looks nice! I think we should move the validation logic out of `packages/encryption` and into a separate package. We can call it `packages/access-control-validation`
> > For the actual schema definitions. I think we should make a separate repo for these and keep the build scripts for the type generation in this new repo. This will allow us to independently version schema files separately from our types package.
> > @Ansonhkg what do you think about the above?
> 
> I'm aligned with @joshLong145 as we've discussed on slack huddle

Setting this as draft again so it holds until we have the new package for schemas and update that here

packages/encryption/src/lib/params-validators.ts
FedericoAmura: It was removed now

packages/access-control-conditions/src/lib/humanizer.ts
```diff
@@ -48,7 +48,7 @@ export const formatAtom = (amount: number): string => {
  * @returns { string } humanized version of the comparator
  */
 export const humanizeComparator = (comparator: string): string | undefined => {
-  let list: any = {
+  const list: { [key: string]: string } = {
```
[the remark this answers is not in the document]
FedericoAmura: There is a similar type in humanizer, but this type is really just to make TS complain less as it is just saying that those are objects with keys and values that are indexes. It is almost like an `string-any`
I don't see much value in having it defined in types package

packages/encryption/src/lib/schemas/index.ts
FedericoAmura: Done on the other repo

packages/encryption/src/lib/schemas/index.ts
FedericoAmura: The `any` is just because Ajv forces it. We care that it is a JSONSchemaType but it does not accept our derived types because it expects some more fields

packages/misc/src/lib/misc.ts
FedericoAmura: Removed to a common place. Also made the schemas avoid compiling if Ajv has them already in the cache

FedericoAmura:
> Overall the implementation looks nice! I think we should move the validation logic out of `packages/encryption` and into a separate package. We can call it `packages/access-control-validation`
> 
> For the actual schema definitions. I think we should make a separate repo for these and keep the build scripts for the type generation in this new repo. This will allow us to independently version schema files separately from our types package.
> 
> @Ansonhkg what do you think about the above?

Moved the schemas and derived classes to their repo and the ACC validation is now similar to ACC hashing and "humanization" (?). All together in the access-control-conditions package

packages/access-control-conditions/src/lib/humanizer.ts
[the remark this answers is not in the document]
joshLong145: yes since it is just making the object indexable it is quite alright to inline

Ansonhkg (approved):
lgtm - a little tool I think would be useful is to check if we are on the latest version for `@lit-protocol/accs-validator` after running `yarn build` in the js-sdk, similar to the `yarn v` command

PR #326, opened 2024-01-26
Feat/manzano habanero final polish

Ansonhkg:
# Description

- This PR rename 90% of the RLI naming to Capacity Credits NFT
- Added the ability to add multiple resources to the authNeededCallback

## Type of change

# How Has This Been Tested?

## Code to run
```
DEBUG=true NETWORK=habanero MINT_NEW=true yarn test:e2e:nodejs --filter=test-rli-pkp-as-delegatee.mjs
```

- [x] e2e nodejs

# Checklist:


packages/types/src/lib/interfaces.ts
```diff
@@ -1605,17 +1607,11 @@ export interface RetryTolerance {
   maxRetryCount?: number;
 }
 
-export interface MintRLIPerDayContext {
+export interface MintCapacityCreditsPerDayContext {
```
Ansonhkg, line 1610: @joshLong145 fyi. I'm also thinking to support "PerMinuteContext" for more specified use case so it doesn't have to do the built-in day-to-minute conversion

eg. 

```
export interface MintCapacityCreditsPerMinuteContext{
  requestPerMinute: number;
}
```

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2492,25 +2492,24 @@ export class LitNodeClientNodeJs extends LitCore {
 
     let siweMessage;
 
-    if (params?.capability && params?.litResource) {
+    if (params?.resourceAbilityRequests) {
+      const resources = params.resourceAbilityRequests.map((r) => r.resource);
```
Ansonhkg, line 2496: @joshLong145 this would allow users to only pass the `resourceAbilityRequests` once to the `getSessionSigs` function and also allowing multiple resources instead of a single resource previously

joshLong145: Makes sense, do we need to order them such that the Capacity Increase request is always index `0`?

packages/types/src/lib/interfaces.ts
joshLong145: I think this makes sense, this would allow for more precise increases.

joshLong145 (approved):

PR #331, opened 2024-01-29
[LIT-2333] Add e2e test for initializing multiple clients

cairomassimo:
# Description

This PR enables the use of multiple instances of `LitNodeClient`, in order to connect to different networks in the same application, at the same time.

This was achieved by removing the global variables `litConfig`, `litNodeClient`, `logger` and `logManager` and their references.

## Type of change

# How Has This Been Tested?

- [ ] Added an e2e test which calls `.connect()` on multiple clients.

# Checklist:

### DIFF
--- a/e2e-nodejs/group-connection/test-connection-multi-network.mjs
+++ b/e2e-nodejs/group-connection/test-connection-multi-network.mjs
@@ -1,0 +1,30 @@
+import path from 'path';
+import { success, fail, testThis, log } from '../../tools/scripts/utils.mjs';
+import { LitNodeClient } from '@lit-protocol/lit-node-client';
+
+export async function main() {
+  const networks = ['cayenne', 'habanero', 'manzano'];
+
+  for (const network of networks) {
+    // ==================== Test Logic ====================
+    const client = new LitNodeClient({
+      litNetwork: network,
+      debug: globalThis.LitCI.debug,
+    });
+    log(`connecting to ${network.toUpperCase()}...`);
+    await client.connect();
+
+    // ==================== Post-Validation ====================
+    if (!client.ready) {
+      return fail('client not ready');
+    }
+    if (client.config.litNetwork !== network) {
+      return fail(`client not connected to ${network}`);
+    }
+  }
+
+  // ==================== Success ====================
+  return success(`Connected to ${networks.join(', ')}`);
+}
+
+await testThis({ name: path.basename(import.meta.url), fn: main });
--- a/packages/lit-auth-client/src/lib/lit-auth-client.ts
+++ b/packages/lit-auth-client/src/lib/lit-auth-client.ts
@@ -20,7 +20,7 @@ import WebAuthnProvider from './providers/WebAuthnProvider';
 import { StytchOtpProvider } from './providers/StytchOtpProvider';
 import AppleProvider from './providers/AppleProvider';
 import StytchAuthFactorOtpProvider from './providers/StytchAuthFactorOtp';
-import { bootstrapLogManager, getLoggerbyId, log } from '@lit-protocol/misc';
+import { log } from '@lit-protocol/misc';
 import { ethers } from 'ethers';
 
 /**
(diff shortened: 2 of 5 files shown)


e2e-nodejs/group-connection/test-connection-multi-network.mjs
```diff
@@ -0,0 +1,30 @@
...
+      debug: globalThis.LitCI.debug,
+    });
+    log(`connecting to ${network.toUpperCase()}...`);
+    await client.connect();
+
+    // ==================== Post-Validation ====================
+    if (!client.ready) {
+      return fail('client not ready');
+    }
+    if (client.config.litNetwork !== network) {
```
joshLong145, line 21: Could we add a check on `globalThis` to assert on the new global namespacing?

packages/lit-auth-client/src/lib/lit-auth-client.ts
```diff
@@ -60,7 +60,6 @@ export class LitAuthClient {
    */
   constructor(options?: LitAuthClientOptions) {
     this.providers = new Map();
-    bootstrapLogManager('auth-client');
```
joshLong145, line 63: why did we remove this line?

packages/misc/src/lib/misc.ts
```diff
@@ -27,10 +27,10 @@ import {
 } from '@lit-protocol/types';
 import { JsonRpcProvider } from '@ethersproject/providers';
 import { Contract } from '@ethersproject/contracts';
-import { LogLevel, LogManager } from '@lit-protocol/logger';
+import { LogManager } from '@lit-protocol/logger';
 import { version } from '@lit-protocol/constants';
 
-const logBuffer: Array<Array<any>> = [];
+const defaultLogger = LogManager.Instance.get('lit-sdk');
```
joshLong145, line 33: will this be a global logger? I would like to keep the package names as categories if possible. Open to discussion however.

packages/misc/src/lib/misc.ts
```diff
@@ -233,121 +211,19 @@ export const getLoggerbyId = (id: string) => {
...
-
-  // if there are there are logs in buffer, print them first and empty the buffer.
-  while (logBuffer.length > 0) {
-    const log = logBuffer.shift() ?? '';
-    globalThis?.logger &&
-      globalThis.logManager.get(globalThis.logger.category).error(...log);
-  }
-
-  globalThis?.logger &&
-    globalThis.logManager.get(globalThis.logger.category).error(...args);
```
joshLong145, line 350: Why did we remove all of these utilities?

joshLong145 (changes requested):
I see you removed methods which we where using in many packages. Can you outline the implications and behavior changes that are introduced with these utilities gone? #331 will also add many of these methods. back when merged. There does not seem to be a path forward for the above mentioned pr with the current implementation.

cairomassimo:
@joshLong145 I updated the PR so that it _only_ adds the e2e test for connecting to multiple networks. The test passes with the code as is.
We discussed about the changes on the logger in a call, but overall I think it makes sense to not apply those changes in this PR anyway.

To be clear, with the code as it is, creating multiple clients does mess up the global state (`globalThis.litConfig` and `globalThis.litNodeClient` get replaced). However:
1. this is the current behavior anyway, so the PR is not introducing a regression here,
2. I think we should recommend the SDK consumers to _not rely on the global state_ anyway, rather than introducing a namespacing convention, espcially since they have to instantiate the clients themselves anyway.

There is one more check to do: making sure that there are no conflicts with the usage of `localStorage` when instantiating more than one client. I couldn't find any, but it's better that someone with more experience with the codebase has a look at this specifically.

e2e-nodejs/group-connection/test-connection-multi-network.mjs
cairomassimo: I didn't implement the global namespacing, I think it's a risky approach (see also https://github.com/LIT-Protocol/js-sdk/pull/331#issuecomment-1916891635). There could be a use case for global namespaces I'm not aware of yet: in that case please let me know.

joshLong145:
@cairomassimo can we close this pr since we no longer need to make code changes?

cairomassimo:
> @cairomassimo can we close this pr since we no longer need to make code changes?

@joshLong145 It's not much, but we can merge it so we have the new e2e test case? I'm making it "ready for review" just in case

Ansonhkg (approved):
I'm ok with having this for our e2e tests

PR #332, opened 2024-01-29
Cleanup/logger

joshLong145:
# Description

Changes implementation of `package/logger` such that logs are no longer persisted under their `requestId` but rather their `category` allowing for logs of any request id to aggregate under the category it was logged under.
For functionality which spans multiple loggers we now search for a `requestId` as they `key` in each collection of category logs.

Implementations can be added in the future which further speeds up logging access. but at this time it is considerably better at managing its resources. 

## Type of change

# How Has This Been Tested?

Testing has been added to `logger.spec.ts` which now checks for cross categorical logging still aggregates requests from all categories.

- [x] logger.spec.ts 

# Checklist:


packages/logger/src/lib/logger.ts
```diff
@@ -397,49 +408,82 @@ export class LogManager {
...
+      instance.Config = this._config;
       return instance;
     }
 
-    const logger = Logger.createLogger(
+    this._loggers.set(
       category,
-      this._level ?? LogLevel.INFO,
-      id ?? ''
+      Logger.createLogger(category, this._level ?? LogLevel.INFO, '', true)
```
FedericoAmura, line 425: Repeated calls to `.get('sameCategory', 'someId')` will create a new logger each time and save those over and over to this set at `category`. Shouldn't we have a previous step where trying to get and return?

As I understand, this will recreate the LogManager default logger each time it has any log requested, even childs. This seems to be the `else` part of the `if` right below

joshLong145: wow yes this is totally true, I guarded the creation of the logger in an `else` case so now it only creates a new instance when required. Good catch 😄

joshLong145: @FedericoAmura 
I cleaned this up

FedericoAmura (approved):
LGTM now!

PR #335, opened 2024-01-30
Fix/default crypto bug

joshLong145:
- fixes a small typeo in the `default` case for a switch statement in `computeHDPubkey`
- adds helpers to `computeHDPubkey` to remove `0x` from hex values represented as strings if found in the input parameters.


Ansonhkg (approved):
lgtm

PR #337, opened 2024-01-31
make capacityDelegationAuthSig optional

joshLong145:
- Makes `capacityDelegationAuthSig` optional on `SessionSigsParams` as we do not need to delegate capacity credit as the `authSig` will be used in place of the `capacityDelegationAuthSig` which is a valid use case as developers may either
   - Delegate usage through `capacityDelegationAuthSig`
   - Provide proof of Capacity Credit through a `authSig`

### DIFF
--- a/packages/types/src/lib/interfaces.ts
+++ b/packages/types/src/lib/interfaces.ts
@@ -1063,8 +1063,11 @@ export interface GetSessionSigsProps {
   sessionKey?: any;
 
   // rateLimitAuthSig: AuthSig;
-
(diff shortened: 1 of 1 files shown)


Ansonhkg (commented):
can we rebase this to https://github.com/LIT-Protocol/js-sdk/pull/338 and merge it there?

Ansonhkg:
> can we rebase this to #338 and merge it there?

That PR is closed as no longer needed.

Ansonhkg (approved):

PR #341, opened 2024-02-05
Refactor `crypto` and other WASM-based modules

cairomassimo:
# Description

Re-implement the functionality of `@lit-protocol/crypto` package as follows (WIP!):

- [x] create a new `nx` workspace (currently in the `ng/` dir), with a simpler/standard build system,
- [x] create a unified Rust crate exposing an API to js,
- [x] merge all Rust crates, which expose functionality via wasm, into the unified one,
- [x] set up a mechanism to generate TypeScript types from Rust (beyond those created by wasm-bindgen, if needed),
- [ ] add a thin (type-safe) JS layer on top of Rust, dealing with data conversion (the less, the better),
- [x] migrate all tests for `@lit-protocol/crypto` to the new implementation,
- [ ] migrate all uses of/references to `@lit-protocol/crypto` to the new implementation.

## Design choices

- `esbuild` as bundler
  - supports embedding binary files out-of-the box, useful for wasm
  - fast
  - supported by jest via `esbuild-jest-transform`
- single javascript package (currently named `@lit-protocol/ng`, to be renamed)
  - tree-shaking should remove unused functionalities,
  - better split into modules, before splitting into packages,
  - split into packages later, when useful/needed (e.g. to reduce build time, to support different build pipelines, what not).

## Notes

The build does not generate `.d.ts` TypeScript declaration files yet.
Not adding this for now, since it could come for free if recent PR https://github.com/nrwl/nx/pull/21084 is merged (or https://github.com/nrwl/nx/issues/20688 fixed anyway).


cairomassimo:
To do:
- make the exported wasm function handle multiple curves dynamically (the api is identical, less code, no/less need for macros)
- see what happens if we bypass `nx` entirely
- set up a long-lived feature branch and/or split this PR into smaller ones for easier reviews (@joshLong145)

ng/wasm/Cargo.toml
```diff
@@ -0,0 +1,82 @@
...
+wasm-bindgen = { version = "0.2", features = ["serde-serialize"] }
+blsful = { version = "2", default-features = false, features = ["rust"] }
+base64_light = "0.1"
+getrandom = { version = "0.2", features = ["js"] }
+hex = "0.4"
+serde = "1.0"
+serde_json = "1.0"
+serde_bare = "0.5"
+serde-wasm-bindgen = "0.5"
+bls12_381_plus = { version = "=0.8.9" }
```
mikelodder7, line 22: Not needed since you're including blsful. Use blsful::inner_types and it gets you the same.

ng/wasm/src/bls.rs
```diff
@@ -0,0 +1,201 @@
...
+#[derive(Tsify, Deserialize)]
+#[tsify(from_wasm_abi)]
+pub enum BlsVariant {
+    Bls12381G1,
+    Bls12381G2,
+}
+
+struct Bls<C>(C);
+
+// We have to create this trait because G1/G2 implementations do not use a common trait for all the methods we need.
```
mikelodder7, line 26: They actually do. See https://github.com/mikelodder7/blsful/blob/main/src/traits/serdes.rs

ng/wasm/src/ecdsa.rs
```diff
@@ -0,0 +1,302 @@
...
+use serde::Deserialize;
+use serde_bytes::Bytes;
+use tsify::Tsify;
+use wasm_bindgen::{prelude::*, JsError};
+
+use crate::abi::{from_js, into_js, JsResult};
+
+#[derive(Tsify, Deserialize)]
+#[tsify(from_wasm_abi)]
+pub enum EcdsaVariant {
```
mikelodder7, line 22: Missing P384

ng/wasm/src/ecdsa.rs
```diff
@@ -0,0 +1,302 @@
...
+        let k = Self::derive_key_inner(id, public_keys)?;
+
+        let k = k.to_encoded_point(false);
+        let k = Bytes::new(k.as_bytes().as_ref());
+        let k = into_js(&k)?;
+
+        Ok(k)
+    }
+
+    fn derive_key_inner(
```
mikelodder7, line 125: Please use these traits instead https://github.com/LIT-Protocol/hd-keys-curves/blob/main/src/lib.rs#L23

ng/wasm/src/frost.rs
```diff
@@ -0,0 +1,420 @@
...
+use serde::Deserialize;
+use serde_bytes::Bytes;
+use tsify::Tsify;
+use wasm_bindgen::prelude::*;
+
+use crate::abi::{from_js, into_js, JsResult};
+
+#[derive(Tsify, Deserialize)]
+#[tsify(from_wasm_abi)]
+pub enum FrostVariant {
```
mikelodder7, line 20: Again why are we copying code from lit-frost? The point of a crate is that its reusable.

ng/wasm/src/ecdsa.rs
```diff
@@ -0,0 +1,302 @@
...
+        let mut powers = vec![C::Scalar::ONE; public_keys.len()];
+        powers[1] = scalar;
+        for i in 2..powers.len() {
+            powers[i] = powers[i - 1] * scalar;
+        }
+        let k = Self::sum_of_products_pippenger(&public_keys, &powers);
+        Ok(k)
+    }
+
+    fn sum_of_products_pippenger(
```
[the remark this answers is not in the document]
cairomassimo: This code was _moved_*, not copied, from https://github.com/LIT-Protocol/lit-ecdsa-wasm-combine/blob/main/src/combiners/hd_ecdsa/mod.rs#L53, I didn't know it was also available in another crate.

[*] the idea is having only one crate that exposes the crypto functionality to JS via WASM, and have it be part of the SDK. Ideally, this crate would not actually implement any cryptography itself, but only delegate to other crates making sure the API is JS-friendly and type-safe.

So indeed this piece of code lives much better in a different crate, I'll happily work on delegating this to `hd-keys-curves`!

ng/wasm/Cargo.toml
cairomassimo: Agree.

I added this to fix the crate to an older version, which would fix a compile error. IIRC, I got the error when using the `blsful` crate from Git, but it doesn't occur at all using versions published to crates.io. Removing.

ng/wasm/src/bls.rs
cairomassimo: Not sure if we can go `Uint8Array` -> `Vec<u8>` -> `{struct}` via serde and these traits. A `TryFrom<Vec<u8>>` would fit better, or we could change the overall approach to rely on serde for JS communication everywhere.
Happy to discuss this today!

PR #346, opened 2024-02-07
Feature/lit 2406 add support for handshaking with nodes after on chain events

joshLong145:
# Description

Addresses cases where the `Active Validator set` for the next epoch has changed and there are new nodes in the set which have not been `attested` 

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


Ansonhkg (commented):
It looks technically correct to me, but I'm worried if it would work 100% correctly as we don't have proper testing during the epoch transition. cc @glitch003

packages/core/src/lib/lit-core.ts
```diff
@@ -301,7 +301,41 @@ export class LitCore {
...
         if (state === StakingStates.NextValidatorSetLocked) {
+          log(
+            'State found to be new validator set locked, checking validator set'
+          );
+          let oldNodeUrls: string[] = [...this.config.bootstrapUrls].sort();
           await this.setNewConfig();
+          let currentNodeUrls: string[] = this.config.bootstrapUrls.sort();
+          let delta: string[] = currentNodeUrls.filter(
+            (item) => oldNodeUrls.indexOf(item) > -1
+          );
```
FedericoAmura, line 312: Just some details:
- there is no need to sort the arrays
- prefer `const` over `let`
- instead of using `Array.indexOf` we could use a Set, Map/object or `Array.includes`

packages/core/src/lib/lit-core.ts
```diff
@@ -301,7 +301,41 @@ export class LitCore {
...
+              The sdk should be able to understand its current execution enviorment and wait on an active 
+              network request to the previous epoch's node set before changing over.
+              
+            */
+            log(
+              'Active validator sets changed, new validators ',
+              delta,
+              'starting node connection'
+            );
+            await this._runHandshakeWithBootstrapUrls().catch(
```
FedericoAmura, line 330: Before this we should clear `this.connectedNodes` because `_runHandshakeWithBootstrapUrls` will try to re-add every handshaked node
The old ones will stay, the new ones will be added and the ones that didn't change will stay with the same value (there is no update on `Set.add`)

joshLong145: Good call, yes quite right. I have been used to using `indexOf` for a long time. switch to `includes` but kept the top level `filter`.

packages/core/src/lib/lit-core.ts
joshLong145: Good call, yes quite right. I have been used to using `indexOf` for a long time. switch to `includes` but kept the top level `filter`.

FedericoAmura (approved):
LGTM!

PR #349, opened 2024-02-09
Feature/lit 2407 js sdk add timestamp on each handshake to alert for when we

joshLong145:
# Description
Adds an expiration check for `latestBlockhash` as after `30 seconds` the hash is expired on the nodes and we should no longer generate new auth materials using the stale hash. `getLatestBlockhash` will now reconnect to nodes to fetch the latest hash. we need to connect to all nodes to assure the `latestBlockhash` returned is the same from all network participants.
Unfortunately this does break backwards compatibility as we now need to make `getLatestBlockhash` and async  implementation as it calls async implementation internally. Since we must wait for the hash to refresh the method signature has now been made `async`

## Type of change

# How Has This Been Tested?

New `e2e` test has been made to force the blockhash to expire and cause the system to refresh its sate.
- `e2e-nodejs/group-latest-blockhash-update/test-block-has-update-after-exp.mjs`

# Checklist:


packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -490,8 +490,29 @@ export class LitNodeClientNodeJs extends LitCore {
...
+    if (!this.ready) {
+      logError('Client not connected, remember to call connect');
+      throwError({
+        message: 'Client not connected',
+        errorKind: LIT_ERROR.LIT_NODE_CLIENT_NOT_READY_ERROR.kind,
+        errorCode: LIT_ERROR.LIT_NODE_CLIENT_NOT_READY_ERROR.code,
+      });
+    }
+
+    if (Date.now() - this.lastblockHashRetrieved! >= 30_000) {
```
Ansonhkg, line 508: can we make `30_000` as constant

joshLong145:
closing for a new implementation using a `setInterval` for syncing block hashes implicitly. This will allow us to keep the current api surface.

PR #350, opened 2024-02-09
Feature/lit 2407 js sdk add timestamp on each handshake to alert for when we

joshLong145:
# Description
Creates a background polling service for refreshing network connections ever `30 seconds` which allows to sync with the current latest block hash on the nodes. this operations completely refreshes the internal state of the network, as we have to `handshake` to every nodes during this process.

This is okay as we are confident in the network state to stay consistent across epochs. So we will not experience undefined behaviors in the event of pending network requests during the reconnection process.

## Type of change

# How Has This Been Tested?
Have created a new test to check the state of the `LitNodeClient` internal network state before and after a resync with the network to confirm **only** the block hash is updated.

- `e2e-nodejs/group-latest-blockhash-update/test-block-has-update-after-exp.mjs`

# Checklist:


packages/core/src/lib/lit-core.ts
```diff
@@ -82,6 +82,8 @@ export class LitCore {
   networkPubKeySet: string | null;
   hdRootPubkeys: string[] | null;
   latestBlockhash: string | null;
+  lastblockHashRetrieved: number | null;
```
Ansonhkg, line 85: `lastBlockHashRetrieved`

Ansonhkg (changes requested):

PR #358, opened 2024-02-14
implement new interface to unify session authentication in client

FedericoAmura:
# Description

This PR changes the way PKP packages handle `sessionSigs`. Now instead of receiving it directly they will obtain it from the client and its auth context. This way we are unifying that management and using them from a common storage instead of making the user handle it directly

## Type of change

# How Has This Been Tested?

- [x] Unit test to validate PKP wallet proper creation
- [x] e2e test to sign using the sessionSigs from the common storage

# Checklist:


packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -222,6 +200,14 @@ export class PKPBase<T = PKPBaseDefaultParams> {
     }
   }
 
+  private validateAuthContext() {
+    if (this.controllerAuthSig && this.authContext) {
+      throw new Error(
```
Ansonhkg, line 205: can we use `this.throwError` in this class instead? we probably need to clean up in other places as well in this file

packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -243,26 +228,29 @@ export class PKPBase<T = PKPBaseDefaultParams> {
...
-    }
+    this.validateAuthContext();
 
-    // If session sigs are provided, they must be an object
+    // If auth context is provided, it must have a client, props and auth methods
     if (
-      this.controllerSessionSigs &&
-      typeof this.controllerSessionSigs !== 'object'
+      this.authContext &&
+      typeof this.authContext?.client !== 'object' &&
```
Ansonhkg, line 236: `typeof` null will also be an object so we gotta check `this.authContext.client !== null` too I think 

image

packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -324,11 +321,12 @@ export class PKPBase<T = PKPBaseDefaultParams> {
       throw new Error('pkpPubKey (aka. uncompressPubKey) is required');
     }
 
-    if (this.controllerAuthSig && this.controllerSessionSigs) {
-      throw new Error(
-        'controllerAuthSig, controllerSessionSigs and controllerAuthMethod are defined, can only use one authorization type'
+    this.validateAuthContext();
+
```
Ansonhkg, line 325: shouldn't we have this code as well here?
```js
if (
  this.authContext &&
  typeof this.authContext?.client !== 'object' &&
  typeof this.authContext?.getSessionSigsProps !== 'function' &&
  !Array.isArray(this.authContext?.authMethods)
) {
  throw new Error('controllerSessionSigs must be an object');
}
```

Ansonhkg (changes requested):
We no longer accept `controllerSessionSigs`, `controllerAuthMethods` and `sessionSigsExpiration` as params, so i think is a breaking change, requires docs update and a migration guide. Also we need to update the e2e tests too if we are using these fields eg. `test-pkp-ethers-handler-sign-tx.mjs`

I think this should be a major bump

joshLong145 (commented):
looks good just one more additional comments

packages/pkp-base/src/lib/pkp-base.ts
FedericoAmura: Changed

packages/pkp-base/src/lib/pkp-base.ts
FedericoAmura: Changed

packages/pkp-base/src/lib/pkp-base.ts
FedericoAmura: You are right. Added

e2e-nodejs/group-pkp-session-sigs/test-pkp-sign-with-session-sigs-session-cache.mjs
```diff
@@ -0,0 +1,94 @@
...
+    pkpPubKey: globalThis.LitCI.AUTH_METHOD_PKP_INFO.publicKey,
+    rpc: LITCONFIG.CHRONICLE_RPC,
+    litNetwork: globalThis.LitCI.network,
+    authContext: {
+      client,
+      getSessionSigsProps: {
+        chain: 'ethereum',
+        expiration: new Date(Date.now() + 60_000 * 60).toISOString(),
+        resourceAbilityRequests: resourceAbilities,
+        sessionKey: sessionKeyPair,
```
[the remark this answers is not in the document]
FedericoAmura: It was loaded from storage. But it was obtained anyway when getting sessionSigs so I removed it to make that option explicit

FedericoAmura:
> We no longer accept `controllerSessionSigs`, `controllerAuthMethods` and `sessionSigsExpiration` as params, so i think is a breaking change, requires docs update and a migration guide. Also we need to update the e2e tests too if we are using these fields eg. `test-pkp-ethers-handler-sign-tx.mjs`
> 
> I think this should be a major bump

We are backwards compatible now =D

Ansonhkg (commented):
Niccee!! 

Can we also update the docs? I have created a branch here https://github.com/LIT-Protocol/docs/pull/226 you can continue what I have edited the file. It should reflect to this page:
https://developer.litprotocol.com/v3/sdk/wallets/walletconnect/#1-create-a-pkpclient

FedericoAmura:
> Niccee!!
> 
> Can we also update the docs? I have created a branch here [LIT-Protocol/docs#226](https://github.com/LIT-Protocol/docs/pull/226) you can continue what I have left there.

Updated!

packages/types/src/lib/interfaces.ts
```diff
@@ -1094,15 +1094,44 @@ export declare type AuthenticatorAttachment = 'cross-platform' | 'platform';
...
+    sessionKeyUri: any;
+    resourceAbilityRequests: Array<LitResourceAbilityRequest>;
+  }) => Promise<boolean>;
+  getSessionSigs: (params: GetSessionSigsProps) => Promise<SessionSigsMap>;
+  signSessionKey: (
+    params: SignSessionKeyProp
+  ) => Promise<SignSessionKeyResponse>;
+}
+
+export interface AuthenticationProps {
```
joshLong145, line 1117: this looks good but i think we also need `authNeededCallback` for if they wish to override the default implementation with out and explicit call to `getSessionSigs`

joshLong145 (commented):
Overall looks good, left a comment on `AuthenticationProps`

PR #366, opened 2024-02-18
ref: Remove ethers in logger

joshLong145:
# Description

replaces `ethers.utils.hashMessage` with a non secure 32 bit hash function which has a low collision chance but quick throughput. This does mean the hashing function is non secure, meaning it is possible to reverse hashes in reasonable time. However we do need a secure hashing implementation for the purposes of the logger. 

Test times are observed to be the same and in some cases **100 - 200ms faster**  measurement was done with `logger.spec.ts` tests with both hashing implementations. 

## Type of change

# How Has This Been Tested?
`logger.spec.ts` was run against the new hashing function

# Checklist:

### DIFF
--- a/packages/logger/src/lib/logger.ts
+++ b/packages/logger/src/lib/logger.ts
@@ -358,7 +358,7 @@ export class Logger {
     if (str.length === 0) return hash;
     for (i = 0; i < str.length; i++) {
       chr = str.charCodeAt(i);
-      hash = ((hash << 5) - hash) + chr;
+      hash = (hash << 5) - hash + chr;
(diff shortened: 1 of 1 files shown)


packages/logger/src/lib/logger.ts
```diff
@@ -1,6 +1,4 @@
 import { version } from '@lit-protocol/constants';
-import { hashMessage } from 'ethers/lib/utils';
-import { encode } from 'punycode';
```
FedericoAmura, line 3: With this, I think we can remove punycode from the project

PR #378, opened 2024-02-21
Add Google Bearer Token Flow

glitch003:
Update: AuthMethod 5 - "Google" is broken on the nodes too.  so, this PR is kind of pointless.

Original description:
Ok so apologies since this is kind of confusing.  

We support 2 google auth methods: 
5 - "Google" - Traditional Bearer Token Oauth.  The nodes resolve and verify the access token by sending it to google and using the response from Google
6 - "GoogleJwt" - JWT token auth.  The nodes resolve the data in the token by parsing the JWT.  The nodes verify the access token by retrieving the google certificates and checking against them.

Right now on Habanero and Manzano, 6 - "GoogleJwt" is broken, because the folder that should hold the google certificates doesn't exist and the node wasn't programmed to create it.  This is already fixed in the node, but those changes haven't been deployed yet.

5 - "Google" should still work though!  So, we can either deploy a whole network upgrade to fix auth method 6 - "GoogleJwt", or we can try and get users to use auth method 5 - "Google"

This PR is an attempt at adding support for auth method 5 - "Google".  We've only ever implemented auth method 6 - "GoogleJwt" in the SDK since it's more efficient and better for the nodes to use this auth method.  So therefore, when the user asks for an auth provider of type "google" it actually uses auth method 6 - "GoogleJwt".

I don't want to break this for existing users, so instead I added two new ProviderTypes: "googleJwt" which uses auth method 6 - "GoogleJwt" and"googleBearer" which uses auth method 5 - "Google".  ProviderType "google" still maps to 6 - "GoogleJwt" which retains backward compatibility, but is indeed confusing.


zach-is-my-name:
Not sure if what you mentioned above is the cause of this or more urgently will block me from implementing the advised remedy:  https://discord.com/channels/896185694857343026/1197596414541303848

lewinskimaciej:
Just letting you know, that this one's actually important to us, would love to help if needed to get this out sooner rather than later.

We're implementing Google integration right now and encountered this exact issue. It seems that v3(`3.2.5`) SDK does not support 5 - Google AuthMethodType properly, it always assumes GoogleJwt (6), it fails on the exact part you fixed in `GoogleProvider.ts` - trying to parse OAuth Access Token as a JWT in `authMethodId`. In `3.2.5` Google Provider Type doesn't work at all with OAuth token(starting with `ya29.(...)` due to this.

We use OAuth flow because we want to implement Social logins through our Supabase instance where we already have some users(signed via Google, Discord and Apple), instead of using the flow provided by the SDK and having to migrate existing users while also creating custom actions which is... well, complex and not yet fully understandable to us. 
It technically all works, we created custom signIn flow, where user signs into our Supabase via the same options available in Lit, we get the provider(google) token back from Supabase, the only gist is we're getting Google's OAuth access_token instead of a JWT. What we then do is we do the authentication part via Lit provided SDK providers, in this case - GoogleProvider.

So since GoogleProvider didn't work with OAuth token, I copied your entire GoogleProvider implementation from this PR to our solution for now and... it seems to work. Kind of, we're failing to sign a session for some reason now but I'm going to make another issue about this in a while 😃

Ansonhkg:
do we still need this @glitch003 ?

Ansonhkg:
Closing for now due to inactivity. cc @glitch003

PR #384, opened 2024-02-22
feat: make capacityTokenId optional

Ansonhkg:
# Description

Making `capacityTokenId` param optional to make delegating capacity credits easier

## Type of change

# How Has This Been Tested?

- [x] e2e

```
DEBUG=true NETWORK=manzano MINT_NEW=true yarn test:e2e:nodejs --filter=test-rli-from-lit-node-client-optional-capacity-token-id.mjs
```

# Checklist:


packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -255,7 +250,7 @@ export class LitNodeClientNodeJs
       litResource,
       LitAbility.RateLimitIncreaseAuth,
       {
-        nft_id: [capacityTokenId],
+        nft_id: capacityTokenId ? [capacityTokenId] : [],
```
glitch003, line 253: i think you should omit nft_id as a key entirely.  at least, looking at the node code, that's what is uses.  like if this key is present it tries to process it.  if it isn't present, then we look at all nfts the address owns.  so i think we should omit it?  

but your test is working?  did you try it on habanero where rate limiting is enabled?  because i don't think it should work as is.  like if you pass in nft_id: [] then on the node it should say "no allowed nft ids!" and the test should fail.  omitting the nft_id key entirely should make the node say "there's no restriction on nft_id, so we will consider all nfts owned by the wallet"

you can see this if statement in the node code here https://github.com/LIT-Protocol/lit-assets/blob/f1d512c1717b82ce1b10bab1827123b9e06f5ea8/rust/lit-node/src/rate_limiting/data.rs#L355

glitch003 (changes requested):

FedericoAmura:
Can we close this? Seems like it got replaced by other PRs

PR #385, opened 2024-02-22
remove old v0 errors from js sdk.  provide error details to the user

glitch003:
Users often get the error "There was an error getting the signing shares from the nodes".  We should aim to include any details we have from the node, and not just return this opaque error.  IMO the "After" message below is a lot better - it clearly indicates that there was a timeout.

We can still improve this a lot, but this is just a quick fix to ensure that the user can at least see what's coming out of the node.  In reality, this timeout error should be a specific error code, etc, but that might be something we need to fix on the node side (aka, the node isn't returning a proper error code in this case, and it's returning some unknown error, but at least that error has details that help with debugging)

Before:
```
"error": {
      "message": "There was an error getting the signing shares from the nodes.",
      "errorKind": "UnknownError",
      "errorCode": "unknown_error"
    }
```
After: 
```
"error": {
      "message": "There was an error getting the signing shares from the nodes",
      "errorCode": "unknown_error",
      "errorKind": "Timeout",
      "status": 500,
      "details": [
        "timeout limit reached timeout limit: 31000ms"
      ],
      "requestId": "a3073ee78572a"
    }
```

Or, in the case where we just didn't get enough shares:
```
"error": {
      "message": "The total number of valid signatures shares 2 does not meet the threshold of 4",
      "errorCode": "no_valid_shares",
      "errorKind": "Unexpected",
      "requestId": "d44599bd20edf"
    }
```


rahul-ramesh (approved):

packages/core/src/lib/lit-core.ts
```diff
@@ -944,13 +944,13 @@ export class LitCore {
       throwError({
         ...res.error,
         message:
-          res.error.message ||
+          `Message: ${res.error.message}.  Details: ${res.error.details}` ||
           'You are not authorized to access to this content',
```
FedericoAmura, line 948: Now the right-side of this new `||` will never be executed because the left one always resolves to a truthy string. Even when `res.error.message` is not defined

It seems that the not authorized message should be part of the `if` conditions right above, not a fallback here. I would make this change but also remove the right side of this `||`

tools/scripts/tools.mjs
```diff
@@ -528,7 +528,7 @@ async function buildFunc() {
       .map((item) => item.replace('apps/', ''))
       .join(',');
 
-    const command = `yarn nx run-many --target=build --exclude=${ignoreList}`;
+    const command = `yarn nx run-many --target=build --exclude=${ignoreList} --parallel=false`;
```
FedericoAmura, line 531: This is not related to this PR. Can we remove?

packages/misc/src/lib/misc.ts
```diff
@@ -181,20 +180,14 @@ export const isNodeClientErrorV0 = (
...
 
 export const isNodeErrorV1 = (
-  nodeError: NodeErrorV0 | NodeErrorV1
+  nodeError: NodeErrorV1
 ): nodeError is NodeErrorV1 => {
   return (
     nodeError.hasOwnProperty('errorCode') &&
     nodeError.hasOwnProperty('errorKind')
   );
 };
```
FedericoAmura, line 189: We can remove this function too

Screenshot from 2024-02-23 11-50-11

glitch003: good catch, removed it!

Ansonhkg (approved):

PR #395, opened 2024-03-07
Get generated stuff out of PRs to make them cleaner

glitch003:
I want to get all the generated code out of PR's, so that the PR only contains the code changes themselves.  

This is a first pass at this:

* Remove any version info from the generated readme.  Instead, we use NPM badges that just show the latest version.
* Remove "last modified" time from autogen_internal-dev.js - this info is in git already
* Run prettier after build, so that our checked in code and our generated JSON files all match


MaximusHaximus (approved):
:+1: less churn on standard builds is a good thing IMO :) LGTM

Ansonhkg (approved):
lgtm!

PR #400, opened 2024-03-15
LIT-2658 - Empty delegateeAddresses passed to createCapacityDelegationAuthSig() now means "everyone can use this"

MaximusHaximus:
# Description

Modifies logic in `createCapacityDelegationAuthSig()` so that callers can omit the `delegateeAddresses` and the behaviour will be that _anyone_ can use it rather than the current logic which adds an empty array,  making it so _nobody_ can use it :)

**Without** this fix, we actually created and set an empty array into the `delegate_to` property on the capability, which meant "Nobody can use this".
**With** this fix, we don't set `delegate_to` on the capability at all if no `delegateeAddresses` was provided as an argument.  This means "Anyone can use this".  It matches the behaviour [described in our docs](https://github.com/LIT-Protocol/docs/blob/9f6bccaad533bafa34b9e4020feca6b74186c57b/docs/sdk/capacity-credits.md?plain=1#L88).

- I ran `eslint -- fix` on the file, so there are incidental changes. Best reviewed commit-by-commit. Actual changes are here: https://github.com/LIT-Protocol/js-sdk/pull/400/commits/4940e490becbc2b8eb9821b69db32769a3f70eb2
- I "drive-by" fixed the usage of `any` type on the arguments passed to the node client, so that TS can do basic inference there now 🎉  :)

## Type of change

# How Has This Been Tested?

- [x] E2E tests

# Checklist:

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -122,8 +122,6 @@ interface CapacityCreditsRes {
   capacityDelegationAuthSig: AuthSig;
 }
 
-/** ---------- Main Export Class ---------- */
-
 export class LitNodeClientNodeJs
   extends LitCore
   implements LitClientSessionManager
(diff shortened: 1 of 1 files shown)


Ansonhkg (approved):
Thanks for picking this up! lgtm!

PR #402, opened 2024-03-15
BREAKING CHANGE: make getLatestBlockhash async

Ansonhkg:
# Description

"so i was thinking about how we get the poller out of the JS SDK.  the main issue is that `litNodeClient.getLatestBlockhash();` is not async.  if we make it async, then we can remove the poller, and just get the latest block hash when they call that function.  this is a breaking change, but def worth it IMO"

## Type of change

# How Has This Been Tested?

- [x] yarn test:e2e:nodejs --filter=test-latest-blockhash.mjs            

# Checklist:

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -17,6 +17,7 @@ import {
 } from '@lit-protocol/constants';
 
 import {
+  checkSevSnpAttestation,
   combineEcdsaShares,
   combineSignatureShares,
   encrypt,
@@ -59,6 +60,7 @@ import {
   GetSigningShareForDecryptionRequest,
   GetWalletSigProps,
   JsonExecutionRequest,
+  JsonHandshakeResponse,
   JsonPkpSignRequest,
   LitClientSessionManager,
   LitNodeClientConfig,
@@ -269,7 +271,7 @@ export class LitNodeClientNodeJs
       throw new Error('Failed to verify capabilities for resource');
     }
 
-    let nonce = this.getLatestBlockhash();
+    let nonce = await this.getLatestBlockhash();
 
     // -- get auth sig
     let siweMessage = new siwe.SiweMessage({
@@ -495,18 +497,66 @@ export class LitNodeClientNodeJs
    * will call refresh if the block hash is expired
    * @returns {Promise<string>} latest block hash from `handhsake` with the lit network.
    */
-  getLatestBlockhash = (): string => {
-    if (!this.ready) {
-      logError('Client not connected, remember to call connect');
+  getLatestBlockhash = async (): Promise<string> => {
+    const requestId = this.getRequestId();
+
+    this.connectedNodes = new Set();
+
+    if (this.config.bootstrapUrls.length <= 0) {
       throwError({
-        message: 'Client not connected',
-        errorKind: LIT_ERROR.LIT_NODE_CLIENT_NOT_READY_ERROR.kind,
-        errorCode: LIT_ERROR.LIT_NODE_CLIENT_NOT_READY_ERROR.code,
+        message: `Failed to get bootstrapUrls for network ${this.config.litNetwork}`,
+        errorKind: LIT_ERROR.INIT_ERROR.kind,
+        errorCode: LIT_ERROR.INIT_ERROR.name,
       });
     }
 
-    // we are confident in this value being non null so we return
-    return this.latestBlockhash!;
+    const respList = [];
+
+    for (const url of this.config.bootstrapUrls) {
+      const challenge = this.getRandomHexString(64);
+
+      try {
+        const resp = await this.handshakeWithNode(
+          { url, challenge },
+          requestId
+        );
+        respList.push(resp);
+      } catch (e) {
+        logError(`Error getting latestBlockhash from node ${url}`, e);
+      }
+    }
+
+    // -- if the list is empty, then return an error
+    if (respList.length === 0) {
+      throwError({
+        message: 'Failed to get latestBlockhash from any node',
+        errorKind: LIT_ERROR.INIT_ERROR.kind,
+        errorCode: LIT_ERROR.INIT_ERROR.name,
+      });
+    }
+
+    // -- pick the latestBlockhash the appears the most
+    const blockhashCounts = respList.reduce(
+      (acc: { [key: string]: number }, response) => {
+        const { latestBlockhash } = response;
+
+        // Check if latestBlockhash is defined
+        if (latestBlockhash) {
+          acc[latestBlockhash] = (acc[latestBlockhash] || 0) + 1;
+        }
+
+        return acc;
+      },
+      {}
+    );
+
+    // -- get the most frequent blockhash
+    const mostFrequestBlockhash = Object.keys(blockhashCounts).reduce(
+      (a: string, b: string) =>
+        blockhashCounts[a] > blockhashCounts[b] ? a : b
+    );
+
+    return mostFrequestBlockhash;
   };
 
   /**
--- a/e2e-nodejs/0_manual-tests/test-block-has-update-after-exp.mjs
+++ b/e2e-nodejs/0_manual-tests/test-block-has-update-after-exp.mjs
@@ -18,7 +18,7 @@ export async function main() {
   console.log(client.hdRootPubkeys);
   let subnetKey = '' + client.subnetPubKey;
   let rootKeys = client.hdRootPubkeys;
-  let blockhash = client.getLatestBlockhash();
+  let blockhash = await client.getLatestBlockhash();
   await new Promise((resolve, _reject) => {
     setTimeout(resolve, 35_000);
   });
--- a/e2e-nodejs/group-connection/test-latest-blockhash.mjs
+++ b/e2e-nodejs/group-connection/test-latest-blockhash.mjs
@@ -1,0 +1,32 @@
+import path from 'path';
+import { success, fail, testThis } from '../../tools/scripts/utils.mjs';
+import LITCONFIG from '../../lit.config.json' assert { type: 'json' };
+import { LitNodeClient } from '@lit-protocol/lit-node-client';
+
+const LIT_NETWORK = 'cayenne';
+
+export async function main() {
+  // ==================== Test Logic ====================
+  const client = new LitNodeClient({
+    litNetwork: globalThis.LitCI.network,
+    debug: globalThis.LitCI.debug,
+    checkNodeAttestation: globalThis.LitCI.sevAttestation,
+  });
+  await client.connect();
+
+  const latestBlockhash = await litNodeClient.getLatestBlockhash();
+
+  // ==================== Post-Validation ====================
+  if (!latestBlockhash) {
+    return fail('latest blockhash not found');
+  }
+
+  if (!latestBlockhash.startsWith('0x')) {
+    return fail('latest blockhash not in hex format');
+  }
+
+  // ==================== Success ====================
+  return success(`Latest blockhash found: ${latestBlockhash}`);
+}
+
+await testThis({ name: path.basename(import.meta.url), fn: main });
--- a/e2e-nodejs/group-hotwallet-with-blockhash/test-hotwallet-with-blockhash.mjs
+++ b/e2e-nodejs/group-hotwallet-with-blockhash/test-hotwallet-with-blockhash.mjs
@@ -62,7 +62,7 @@ export async function main() {
   });
   await litNodeClient.connect();
 
-  let nonce = litNodeClient.getLatestBlockhash();
+  let nonce = await litNodeClient.getLatestBlockhash();
   console.log('Eth blockhash nonce- ', nonce);
 
   if (!nonce) {
--- a/e2e-nodejs/group-latest-blockhash-update/test-block-has-update-after-exp.mjs
+++ b/e2e-nodejs/group-latest-blockhash-update/test-block-has-update-after-exp.mjs
@@ -18,7 +18,7 @@ export async function main() {
   console.log(client.hdRootPubkeys);
   let subnetKey = '' + client.subnetPubKey;
   let rootKeys = client.hdRootPubkeys;
-  let blockhash = client.getLatestBlockhash();
+  let blockhash = await client.getLatestBlockhash();
   await new Promise((resolve, _reject) => {
     setTimeout(resolve, 35_000);
   });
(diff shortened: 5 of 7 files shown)


packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -495,18 +497,66 @@ export class LitNodeClientNodeJs
...
+        logError(`Error getting latestBlockhash from node ${url}`, e);
+      }
+    }
+
+    // -- if the list is empty, then return an error
+    if (respList.length === 0) {
+      throwError({
+        message: 'Failed to get latestBlockhash from any node',
+        errorKind: LIT_ERROR.INIT_ERROR.kind,
+        errorCode: LIT_ERROR.INIT_ERROR.name,
```
Adarsh-Kumar28, line 534: Is the initialization check happening at other places? I assume the logic will remain the same? Let's extract it out in a function in that case?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -495,18 +497,66 @@ export class LitNodeClientNodeJs
...
+      {}
+    );
+
+    // -- get the most frequent blockhash
+    const mostFrequestBlockhash = Object.keys(blockhashCounts).reduce(
+      (a: string, b: string) =>
+        blockhashCounts[a] > blockhashCounts[b] ? a : b
+    );
+
+    return mostFrequestBlockhash;
```
Adarsh-Kumar28, line 559: Let's use Lodash here?

Adarsh-Kumar28 (commented):
We also want to remove the poller? I don't see it being removed?
Also is the purpose of this PR to always pull the latest blockhash from the nodes before making a request?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
glitch003: i think this getLatestBlockhash is that function?  Or are you thinking this has overlap with the connect() function which actually does the same thing (handshake w/all nodes) ?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
glitch003: i think we have this functionality in here: https://github.com/LIT-Protocol/js-sdk/blob/8c0a098c75cf3c6949718e5d0fb2cfaa0970182d/packages/misc/src/lib/misc.ts#L60

should we use this instead?

glitch003:
this makes sense to me and looks good.  i think we should add caching?  if the block hash is less than 30s old, no need to talk to the nodes and get the latest?  probably a good idea to prevent developers that like, accidentally call getLatestBlockhash() on every react re-render from hitting the nodes more than necessary.

glitch003 (dismissed):
approving pending resolution of open questions

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: @Adarsh-Kumar28 I wasn't sure if we should introduce a new library particularly for this function. I would like to have a new PR/task dedicated to replace everything that can be refactored to use Lodash instead. What's your opinion on this @MaximusHaximus

Ansonhkg: > @Adarsh-Kumar28 I wasn't sure if we should introduce a new library particularly for this function. I would like to have a new PR/task dedicated to replace everything that can be refactored to use Lodash instead. What's your opinion on this @MaximusHaximus

@Adarsh-Kumar28  I wasn't sure if we should introduce a new library particularly for this function. I would like to have a new PR/task dedicated to replace everything that can be refactored to use Lodash instead. What's your opinion on this @MaximusHaximus

Ansonhkg: > Let's use Lodash here?

@Adarsh-Kumar28 I wasn't sure if we should introduce a new library particularly for this function. I would like to have a new PR/task dedicated to replace everything that can be refactored to use Lodash instead. What's your opinion on this @MaximusHaximus



> i think we have this functionality in here:
> 
> https://github.com/LIT-Protocol/js-sdk/blob/8c0a098c75cf3c6949718e5d0fb2cfaa0970182d/packages/misc/src/lib/misc.ts#L60
> 
> should we use this instead?

Yea, we can for the `mostFrequestBlockhash` constant, but we still need to use the .reduce of extract all blockhash as array.

MaximusHaximus (changes requested):
This is re-implementing a lot of existing logic from `lit-core`, where we do this exact process already during the `connect()` flow. In fact, functionally it _is_ the same, as it handshakes with all nodes, and even rebuilds the `connectedNodes`, but in its current state it leaves `connectedNodes` in an inconsistent state for the duration of this process -- while we're waiting on handshakes to return, the set will be mutated 'in place' - calling `connect()` no longer does that as of LIT-2560 -- and this could possibly run multiple concurrent handshakes if the method is called multiple times close together, which `connect()` will no longer do.

1. Can we move this code to live in `core`, since it is managing the same state as `core` does anyway?
2. Can we avoid re-implementing 'get most common value' and error handling/logging, by instead adding the method `core` that calls `this.connect()` and then returns `this.latestBlockhash`?
- We're already updating the `latestBlockhash` property during `connect()` execution anyway, using another implementation that already gets the 'most common string' in the returned values from the nodes
- We already have error handling and logging on fetching the blockhash from each node and an error is thrown if we fail to get the blockhash from any node

This method in `lit-node-client` can then be delegated to the `core` implementation.
When we find a better place to source `latestBlockhash` from, we can replace the `connect()` usage in `lit-core` to be lighter-weight.

Ansonhkg:
> This is re-implementing a lot of existing logic from `lit-core`, where we do this exact process already during the `connect()` flow. In fact, functionally it _is_ the same, as it handshakes with all nodes, and even rebuilds the `connectedNodes`, but in its current state it leaves `connectedNodes` in an inconsistent state for the duration of this process -- while we're waiting on handshakes to return, the set will be mutated 'in place' - calling `connect()` no longer does that as of LIT-2560 -- and this could possibly run multiple concurrent handshakes if the method is called multiple times close together, which `connect()` will no longer do.
> 
> 1. Can we move this code to live in `core`, since it is managing the same state as `core` does anyway?
> 2. Can we avoid re-implementing 'get most common value' and error handling/logging, by instead adding the method `core` that calls `this.connect()` and then returns `this.latestBlockhash`?
> 
> * We're already updating the `latestBlockhash` property during `connect()` execution anyway, using another implementation that already gets the 'most common string' in the returned values from the nodes
> * We already have error handling and logging on fetching the blockhash from each node and an error is thrown if we fail to get the blockhash from any node
> 
> This method in `lit-node-client` can then be delegated to the `core` implementation. When we find a better place to source `latestBlockhash` from, we can replace the `connect()` usage in `lit-core` to be lighter-weight.

Great points! Made relevant changes :)

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -28,6 +28,7 @@ import {
 } from '@lit-protocol/constants';
 import { LitCore } from '@lit-protocol/core';
 import {
+  checkSevSnpAttestation,
```
MaximusHaximus, line 31: :hocho: unused import

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -80,6 +81,7 @@ import type {
   GetSigningShareForDecryptionRequest,
   GetWalletSigProps,
   JsonExecutionRequest,
+  JsonHandshakeResponse,
```
MaximusHaximus, line 84: :hocho: unused import

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -253,7 +255,7 @@ export class LitNodeClientNodeJs
       throw new Error('Failed to verify capabilities for resource');
     }
 
-    const nonce = this.getLatestBlockhash();
+    let nonce = await this.getLatestBlockhash();
```
MaximusHaximus, line 258: We're not mutating `nonce`; it should be a `const` (this is also an eslint error for `prefer-const`)

packages/core/src/lib/lit-core.ts
```diff
@@ -469,6 +469,18 @@ export class LitCore {
   };
 
   /**
+   * returns the latest block hash.
+   * will call refresh if the block hash is expired
```
MaximusHaximus, line 473: JSDoc is misleading here -- this will _always_ return the latest blockhash from the nodes

MaximusHaximus (changes requested):
Lookin' good -- thanks for adding a test for it. Just a few things in-line.

MaximusHaximus (changes requested):
We're just about there I think!

- I think there are a couple of instances in the `apps` directory that need still need updating - [here](https://github.com/LIT-Protocol/js-sdk/blob/56ed2a0c41e255bd394a2ecb2a51097e6f3ffdc2/apps/html/manual_tests.html#L47) and [here](https://github.com/LIT-Protocol/js-sdk/blob/f02430dcfa7be5e5ca91e64575a4a2e8a4802592/apps/react/src/app/lit-node-client-page.tsx#L44)
- Also, not a blocker for this PR I assume, but I think we'll also need a PR against [docs](repo:LIT-Protocol/docs getLatestBlockhash) to update any references there

Otherwise this LGTM :+1:!


**NOT blockers for this PR, but the changes to `.vscode/settings.json` made me consider:**
Since we're trying to improve code hygiene and consistency, and eliminate eslint errors and unnecessary reformat commits...
1. It would be helpful if you can test enabling format-on-save using prettier and automatic ESLint fix on save for vscode users and commit those changes either in this PR or in a follow-up since it's not directly related to this PR :).
**Prettier**: Check out # 2 in this post: https://stackoverflow.com/a/70242338
**ESLint**: https://www.masrinastudio.com/post/eslint-formatting-vscode-guide/
2.  It would also probably provide nice auto-complete for folks in vscode using conventional commits to just enter in all of the conventional commit scopes at once rather than only adding a few at a time :)

I'll commit some editor configuration for Webstorm that does the same things, since that's what I'm using :)

Ansonhkg:
> We're just about there I think!
> 
> * I think there are a couple of instances in the `apps` directory that need still need updating - [here](https://github.com/LIT-Protocol/js-sdk/blob/56ed2a0c41e255bd394a2ecb2a51097e6f3ffdc2/apps/html/manual_tests.html#L47) and [here](https://github.com/LIT-Protocol/js-sdk/blob/f02430dcfa7be5e5ca91e64575a4a2e8a4802592/apps/react/src/app/lit-node-client-page.tsx#L44)
> * Also, not a blocker for this PR I assume, but I think we'll also need a PR against [docs](repo:LIT-Protocol/docs getLatestBlockhash) to update any references there
> 
> Otherwise this LGTM 👍!
> 
> **NOT blockers for this PR, but the changes to `.vscode/settings.json` made me consider:** Since we're trying to improve code hygiene and consistency, and eliminate eslint errors and unnecessary reformat commits...
> 
> 1. It would be helpful if you can test enabling format-on-save using prettier and automatic ESLint fix on save for vscode users and commit those changes either in this PR or in a follow-up since it's not directly related to this PR :).
>    **Prettier**: Check out # 2 in this post: https://stackoverflow.com/a/70242338
>    **ESLint**: https://www.masrinastudio.com/post/eslint-formatting-vscode-guide/
> 2. It would also probably provide nice auto-complete for folks in vscode using conventional commits to just enter in all of the conventional commit scopes at once rather than only adding a few at a time :)
> 
> I'll commit some editor configuration for Webstorm that does the same things, since that's what I'm using :)

Thanks for being attentive to all required changes for this! Made relevant changes! I will look into the testing prettier and automatic ESLint fix in a follow-up PR :)

Ansonhkg:
> We're just about there I think!
> 
> * I think there are a couple of instances in the `apps` directory that need still need updating - [here](https://github.com/LIT-Protocol/js-sdk/blob/56ed2a0c41e255bd394a2ecb2a51097e6f3ffdc2/apps/html/manual_tests.html#L47) and [here](https://github.com/LIT-Protocol/js-sdk/blob/f02430dcfa7be5e5ca91e64575a4a2e8a4802592/apps/react/src/app/lit-node-client-page.tsx#L44)
> * Also, not a blocker for this PR I assume, but I think we'll also need a PR against [docs](repo:LIT-Protocol/docs getLatestBlockhash) to update any references there
> 
> Otherwise this LGTM 👍!
> 
> **NOT blockers for this PR, but the changes to `.vscode/settings.json` made me consider:** Since we're trying to improve code hygiene and consistency, and eliminate eslint errors and unnecessary reformat commits...
> 
> 1. It would be helpful if you can test enabling format-on-save using prettier and automatic ESLint fix on save for vscode users and commit those changes either in this PR or in a follow-up since it's not directly related to this PR :).
>    **Prettier**: Check out # 2 in this post: https://stackoverflow.com/a/70242338
>    **ESLint**: https://www.masrinastudio.com/post/eslint-formatting-vscode-guide/
> 2. It would also probably provide nice auto-complete for folks in vscode using conventional commits to just enter in all of the conventional commit scopes at once rather than only adding a few at a time :)
> 
> I'll commit some editor configuration for Webstorm that does the same things, since that's what I'm using :)

Thanks for being attentive to all required changes for this! Made relevant changes! I will look into the testing prettier and automatic ESLint fix in a follow-up PR :)

MaximusHaximus (approved):
👍  Looks good! 💖

Ansonhkg:
Merge conflicts occurred after another PR was rebased and merged to master

MaximusHaximus (dismissed):
:+1: :shipit:

MaximusHaximus:
@Ansonhkg I assume we want this to land in this week's release w/ other breaking changes -- LMK as soon as the conflicts are resolved and I'll re-add approved review so you can get it merged :). It sure will be nice when we get to the point that we're not modifying package.json in our feature PRs anymore :)

joshLong145 (approved):

joshLong145:
@MaximusHaximus conflicts are resolved can you re review?

packages/core/src/lib/lit-core.ts
```diff
@@ -484,6 +484,17 @@ export class LitCore {
   };
 
   /**
+   * Return the latest blockhash from the nodes
+   * @returns { Promise<string> } latest blockhash
+   */
+  getLatestBlockhash = async (): Promise<string> => {
+    await this.connect();
+
+    return this.latestBlockhash as string;
```
jtary, line 493: If we are going to assert this is a string, we should probably actually check and make sure and throw if for some reason latestBlockhash wasn't set on connect.

jtary (approved):

jtary (approved):

packages/core/src/lib/lit-core.ts
joshLong145: good catch!

joshLong145 (approved):

PR #407, opened 2024-03-21
fix(types): LIT-2706 - Modify LitNodeClientConfig properties incorrectly marked as required to be optional

MaximusHaximus:
# Description
After we removed the any union type on the node client constructor in v4.1, users began reporting that they were being forced to add properties to the configuration that they did not expect to be required.

Upon inspection, there is a complete set of default values for the node client provided in lit-core , some properties are set in the code itself based on the value of litNetwork, and others are loaded from the on-chain staking contract.

Thus, every property in the config is theoretically optional except for litNetwork — but this was being hidden by the any union type that was allowing consumers to pass objects with any shape into the constructor.

This PR makes those properties optional.

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -76,13 +76,28 @@ interface EpochCache {
   lastUpdateTime: null | number;
 }
 
+export type LitNodeClientConfigWithDefaults = Required<
+  Pick<
+    LitNodeClientConfig,
+    | 'bootstrapUrls'
+    | 'alertWhenUnauthorized'
+    | 'debug'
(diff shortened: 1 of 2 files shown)


glitch003 (approved):
looks good, thanks

Ansonhkg:
lgtm! 

notes:
gonna "re-visit the structure of this config object; since litNetwork is the only required property to be provided, it may make more sense to have the constructor signature here actually be: `{ litNetwork: LIT_NETWORKS_KEYS, config?: {…LitNodeClientProps omitting litNetwork}`"

PR #414, opened 2024-04-03
(WIP) Auth Unification

Ansonhkg:
# Description

This PR is primarily for testing the authentication unification functionalities. We will gradually select and move the necessary changes into separate PRs for detailed review. 

# _**❗️❗️THIS PR WILL NOT BE MERGED❗️❗️**_


Ansonhkg:
Suggested edit:
```diff
diff --git a/packages/auth-helpers/src/index.ts b/packages/auth-helpers/src/index.ts
index e9cb2baf..f30413d6 100644
--- a/packages/auth-helpers/src/index.ts
+++ b/packages/auth-helpers/src/index.ts
@@ -9,5 +9,4 @@ export * from './lib/siwe';
 export * from './lib/recap/recap-session-capability-object';
 export * from './lib/auth-needed-callback/craft-auth-sig';
 export * from './lib/auth-needed-callback/create-siwe-message/create-siwe-message';
-export * from './lib/auth-needed-callback/get-auth-needed-callback';
 export * from './lib/auth-needed-callback/create-recap-object/create-capacity-delegation-recap';

```

e2e-nodejs/bun-tests/test.ts
```diff
@@ -0,0 +1,475 @@
...
+
+    console.log('res:', res);
+  },
+
+  /**
+   * Test Commands:
+   * ✅ yarn test:integrate --filter=testUseSessionSigsToPkpSign --network=habanero --version=v0
+   * ❌ yarn test:integrate --filter=testUseSessionSigsToPkpSign --network=localchain --version=v1
+   */
+  testUseSessionSigsToPkpSign: async () => {
```
Ansonhkg, line 207: It doesn't work on local nodes, but works on Habanero.

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -369,30 +363,56 @@ export class LitNodeClientNodeJs
...
-        `Storage key "${storageKey}" is missing. Not a problem. Contiune...`
+      log(
+        `[getSessionKey] Storage key "${storageKey}" is missing. Not a problem. Contiune...`
       );
 
+      // Check if a valid session key exists in cache
+      if (sessionKeyCache && Date.now() - sessionKeyCache.timestamp < 30000) {
+        log(`[getSessionKey] Returning session key from cache.`);
+        return sessionKeyCache.value;
+      }
```
Ansonhkg, line 389: @joshLong145 wdyt about this solution to fix the session key mismatch issue due to a new session key is being generated because the storage provider is not found after the key has been verified

local-tests/test.ts
```diff
@@ -0,0 +1,449 @@
...
+        });
+
+        const authSig = await craftAuthSig({
+          signer: hotWallet,
+          toSign,
+        });
+
+        return authSig;
+      },
+      // NOT SURE ABOUT THIS
```
DashKash54, line 128: This will be required for running against Habanero but not local?

local-tests/test.ts
```diff
@@ -0,0 +1,449 @@
...
+      console.log('Error:', e);
+    }
+  },
+
+  /**
+   * Test Commands:
+   * ✅ yarn test:local --filter=testGetSessionSigs --network=habanero --version=v0
+   * ✅ yarn test:local --filter=testGetSessionSigs --network=localchain --version=v1
+   */
+  testGetSessionSigs: async () => {
```
DashKash54, line 86: What is this testing? It's creating a sessionSig with an EOA Wallet so there should be no difference in the test commands as all of this happens locally without interacting with the nodes?

DashKash54, line 86: Maybe rename it `getEoaSessionSigs`?

local-tests/test.ts
```diff
@@ -0,0 +1,449 @@
...
+
+    return sessionSigs;
+  },
+
+  /**
+   * Test Commands:
+   * ✅ yarn test:local --filter=testUseSessionSigsToExecuteJsConsoleLog --network=habanero --version=v0
+   * ✅ yarn test:local --filter=testUseSessionSigsToExecuteJsConsoleLog --network=localchain --version=v1
+   */
+  testUseSessionSigsToExecuteJsConsoleLog: async () => {
```
DashKash54, line 142:
```suggestion
  testUseEoaSessionSigsToExecuteJsConsoleLog: async () => {
```

local-tests/test.ts
DashKash54, line 1: 1. There's test missing for pkpSessionSigs for executeJs
2. Also can you run a Lit Action for pkpSessionSigs and then use that sessionSig for pkpSign & executeJs

These will ensure proper test coverage

DashKash54 (commented):
Left a few comments for the `local-tests/test.ts` file

local-tests/test.ts
Ansonhkg: It was just testing to get the sessionSigs cus in the beginning I wasn't able to.

Ansonhkg: If we are not testing the functionality of getting sessionSigs alone, I think we can just remove the test

local-tests/test.ts
Ansonhkg: Makes sense

local-tests/test.ts
DashKash54: This is completely SDK side right? So no need to check this?

DashKash54: Cool let's remove this if not required?

PR #418, opened 2024-04-11
feat(package.json): add local dev build script

Ansonhkg:
# Description

Average build time now running yarn build on a M1 Max pro machine takes around 17s (which is not that bad), but we can optimise it further by stripping away operations that don't matter in local development.

Introducing `yarn build:dev` command

## Type of change

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "reset:dev": "yarn tools --remove-local-dev",
     "reset": "rm -rf ./dist/packages && yarn reset:dev",
     "build": "yarn build:packages",
+    "build:dev": "yarn tools --remove-local-dev && rm -rf ./dist && yarn tools check --no-empty-directories=true && yarn tools fixTsConfig && yarn tools --build --packages && yarn tools --setup-local-dev && yarn build:verify",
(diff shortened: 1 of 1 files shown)


FedericoAmura (dismissed):

joshLong145:
@Ansonhkg does this mean when we want to `publish` we should first run `yarn build` instead of `yarn build:dev` this will only be relevant until we have a ci publish pipeline.

Ansonhkg:
> @Ansonhkg Does this mean when we want to `publish` we should first run `yarn build` instead of `yarn build:dev` this will only be relevant until we have a ci publish pipeline.

Yup - you should never need to use `yarn build` unless you want to test or publish it, `yarn build:dev` all the way for local development

joshLong145:
> > @Ansonhkg Does this mean when we want to `publish` we should first run `yarn build` instead of `yarn build:dev` this will only be relevant until we have a ci publish pipeline.
> 
> Yup - you should never need to use `yarn build` unless you want to test or publish it, `yarn build:dev` all the way for local development

Sounds good. maybe we add this to the README?

joshLong145 (dismissed):

PR #425, opened 2024-04-11
feat: decouple eth wallet from provider

Ansonhkg:
# Description

At the moment, if you want to create an eth wallet auth method from Lit Auth Client, you will need to provide an API key. This shouldn't be the case because it's all client side signing the siwe message, and all we do is wrap around their authSig as an access token as an auth method. 

# Proposed solution

Create a static `authenticate` method, and then the original method can consume the static method, so that it can still use the base methods that relies on the relayer 

```ts
  const authMethod = await EthWalletProvider.authenticate({
    signer: wallet,
    litNodeClient: client,
  });
```

## Type of change

# How Has This Been Tested?

## Test command
```
DEBUG=true NETWORK=cayenne yarn test:e2e:nodejs --filter=test-pkp-eth-wallet-provider.mjs
```

## Test result

```
✓    authMethod created: {"authMethodType":1,"accessToken":"{\"sig\":\"0x433c7c22d488dc604b610312c2ad074d1c3e339d42fb14a80430fee69637f6011cae28ed0fe7110fc6f9ffce276475daf243bdb6bab4d5253aa7da0440e45fc81b\",\"derivedVia\":\"web3.eth.personal.sign\",\"signedMessage\":\"localhost wants you to sign in with your Ethereum account:\\n0xeF71c2604f17Ec6Fc13409DF24EfdC440D240d30\\n\\n\\nURI: http://localhost\\nVersion: 1\\nChain ID: 1\\nNonce: 0x81347476f3c006acb4a7863d31de774d6b90c1600bcfb6af110b3ff7dbe28cd0\\nIssued At: 2024-04-11T22:32:46.376Z\\nExpiration Time: 2024-04-12T22:32:46.375Z\",\"address\":\"0xeF71c2604f17Ec6Fc13409DF24EfdC440D240d30\"}"} (12ms) 
```

# Checklist:

### DIFF
--- a/packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
+++ b/packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
@@ -9,7 +9,10 @@ import { LIT_CHAINS, AuthMethodType, LIT_ERROR } from '@lit-protocol/constants';
 import { SiweMessage } from 'siwe';
 import { ethers } from 'ethers';
 import { BaseProvider } from './BaseProvider';
-import { checkAndSignAuthMessage } from '@lit-protocol/lit-node-client';
+import {
+  LitNodeClient,
+  checkAndSignAuthMessage,
+} from '@lit-protocol/lit-node-client';
 import { log, throwError } from '@lit-protocol/misc';
 
 export default class EthWalletProvider extends BaseProvider {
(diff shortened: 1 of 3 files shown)


packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
```diff
@@ -50,47 +54,92 @@ export default class EthWalletProvider extends BaseProvider {
...
+   * @memberof EthWalletProvider
+   * 
+   * @example
+   * ```typescript
+   *   const authMethod = await EthWalletProvider.authenticate({
+   *      signer: wallet,
+   *      litNodeClient: client,
+   *   });
+   * ```
+   */
```
joshLong145, line 87: nice use of jsdoc annotations. we should start using the `@example` in more

packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
```diff
@@ -50,47 +54,92 @@ export default class EthWalletProvider extends BaseProvider {
...
+  public static async authenticate({
+    signer,
+    address,
+    chain,
+    litNodeClient,
+    expiration,
+    domain,
+    origin,
+  }: {
+    signer: ethers.Signer | ethers.Wallet | any;
```
joshLong145, line 97: why do we union with `any` since by doing so we will essentially omit the typing of `ethers.Signer | ethers.Wallet`

packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
```diff
@@ -50,47 +54,92 @@ export default class EthWalletProvider extends BaseProvider {
...
       // Use provided function to sign message
-      const signature = await signMessage(toSign);
+      const signature = await signer.signMessage(toSign);
 
       authSig = {
         sig: signature,
         derivedVia: 'web3.eth.personal.sign',
         signedMessage: toSign,
-        address: address,
+        address: signer.address,
```
joshLong145, line 139: same as above

packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
```diff
@@ -50,47 +54,92 @@ export default class EthWalletProvider extends BaseProvider {
...
       };
     } else {
-      if (!this.litNodeClient.latestBlockhash) {
+      if (!litNodeClient.latestBlockhash) {
```
joshLong145, line 142: If we are going to throw an exception if the block hash isn't available we will want to check right when we enter this function as to not sign if we cannot get a `nonce` value.

packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
Ansonhkg: You are right - I've changed it to `EthWalletAuthenticateOptions` basically a eth signer implementation

joshLong145 (approved):
🔥

MaximusHaximus (commented):
I'm a bit concerned about the behaviour differences for the static `authenticate()` method having different behaviours when `domain` and `uri` are not provided vs. constructing an instance without providing those arguments, then calling the instance method.  Added a couple of other things in-line

PR #426, opened 2024-04-12
Fix/lit network must be provided of lit auth client

Ansonhkg:
# Description

We should ~~force~~ **warn** users to provide `lit network` when initialising lit auth client, and we shouldn't default to `cayenne` 

## Type of change

### DIFF
--- a/packages/lit-auth-client/src/lib/lit-auth-client.ts
+++ b/packages/lit-auth-client/src/lib/lit-auth-client.ts
@@ -87,8 +87,7 @@ export class LitAuthClient {
     if (options?.litNodeClient) {
       this.litNodeClient = options?.litNodeClient;
     } else {
-
-      if(!options.litNetwork){
+      if (!options.litNetwork) {
         throw new Error(`litNetwork is required to create a LitNodeClient`);
       }
 
--- a/packages/lit-auth-client/src/lib/lit-auth-client.ts
+++ b/packages/lit-auth-client/src/lib/lit-auth-client.ts
@@ -87,8 +87,13 @@ export class LitAuthClient {
     if (options?.litNodeClient) {
       this.litNodeClient = options?.litNodeClient;
     } else {
+
+      if(!options.litNetwork){
+        throw new Error(`litNetwork is required to create a LitNodeClient`);
+      }
+
       this.litNodeClient = new LitNodeClient({
-        litNetwork: 'cayenne',
+        litNetwork: options.litNetwork,
         debug: options.debug ?? false,
       });
     }
--- a/packages/lit-auth-client/src/lib/lit-auth-client.ts
+++ b/packages/lit-auth-client/src/lib/lit-auth-client.ts
@@ -109,7 +109,8 @@ export class LitAuthClient {
 
       if (!supportedNetworks.includes(this.litNodeClient.config.litNetwork)) {
         throw new Error(
-          `Unsupported litNetwork: ${this.litNodeClient.config.litNetwork
+          `Unsupported litNetwork: ${
+            this.litNodeClient.config.litNetwork
           }. Supported networks are: ${supportedNetworks.join(', ')}`
         );
       }
--- a/packages/lit-auth-client/src/lib/lit-auth-client.ts
+++ b/packages/lit-auth-client/src/lib/lit-auth-client.ts
@@ -88,9 +88,8 @@ export class LitAuthClient {
       this.litNodeClient = options?.litNodeClient;
     } else {
       if (!options.litNetwork) {
-        throw new Error(`litNetwork is required to create a LitNodeClient`);
+        console.warn(`litNetwork is required to create a LitNodeClient`);
       }
-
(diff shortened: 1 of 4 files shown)


packages/types/src/lib/interfaces.ts
```diff
@@ -1178,6 +1178,8 @@ export interface LitAuthClientOptions {
   debug?: boolean;
 
   litOtpConfig?: OtpProviderOptions;
+
+  litNetwork?: LIT_NETWORKS_KEYS;
```
jtary, line 1182: Would it make sense to make this a required field?

Anytime you don't provide a network, the consuming function would have to default it (which is probably not the behaviour we want).

Ansonhkg: I think it should be optional, because if they pass in `litNodeClient` they don't need to provide `litNetwork` because it's already set in the `litNodeClient`

packages/lit-auth-client/src/lib/lit-auth-client.ts
```diff
@@ -87,8 +87,11 @@ export class LitAuthClient {
     if (options?.litNodeClient) {
       this.litNodeClient = options?.litNodeClient;
     } else {
+      if (!options.litNetwork) {
+        console.warn(`litNetwork is required to create a LitNodeClient`);
+      }
       this.litNodeClient = new LitNodeClient({
-        litNetwork: 'cayenne',
+        litNetwork: options.litNetwork,
```
joshLong145, line 94: I think we should still default it in case the user doesnt supply a network option. But maybe we want this to be a breaking change?

Ansonhkg: Yea, this will still default to `cayenne` as this is set by default in `litNodeClient`. I agreed with Daryl that we should pass in by instance, and at the same time we need to prioritize addressing and patching any issues as soon as they arise. This PR can serve as a useful reference for future refactoring efforts..

MaximusHaximus (changes requested):
We're about to ship a major release, why don't we just go ahead and remove the fuzzy logic that constructs an instance _sometimes_ in this constructor, and make the constructor always require the user to provide an instance of the `LitNodeClient`?

Otherwise this won't just stop at adding litNetwork -- people will eventually want the entire litNodeClient config to be possible to be passed in here slowly over time as they run into different things they need to configure about the client; best if they just construct it and pass it in. The sooner we make the change the fewer consumers will be impacted and the less work it will take in the end -- we won't need to make API changes (and update documentation and examples, etc.) that are then removed entirely when we go ahead and make the API require an instance to be passed in to the constructor.

Also, the current API is really quite awkward right now, because if the consumer needs to actually _use_ the node client that we constructed for them automatically inside the auth client, then they need to call `connect()` on it by referencing it as a property on their auth instance; it makes it difficult to reason about how the dependencies actually work here.  My guess is that some logic inside of our auth providers is going to either call `connect()` or throw an error when it tries to actually _use_ the node client that we auto-created here on their behalf, which means those methods can take a very long time to run for the first execution and can throw any error that `connect()` can throw in _addition_ to any errors specific to their actual purposes.

This pattern also makes it much more likely that they end up with another instance of the client (complete with event listeners/polling) somewhere else in their app.

Also, removing the code that constructs this and defining that the consumer _always_ provides an instance would be less effort and easier to review and reason about the behaviour of - note the discussions in this PR about whether the property should be optional or required, and whether we need to continue to provide a default value to the instance.  Imagine we change the lit node client API in the future to no longer default to `cayenne` if you don't provide a network -- now that change will be a _breaking change for this constructor too_.  Why don't we just go ahead and move the construction out of the class and ship it with the current set of other breaking changes we're about to release? :)

packages/lit-auth-client/src/lib/lit-auth-client.ts
MaximusHaximus: I agree, we should prioritize addressing/patching the issue... I just don't see why we can't do that by removing the offending implicit behaviour entirely.

I'm also confused about the content of this PR -- the current code doesn't even fix the problem; a console.warn() doesn't actually make the new argument required like the PR title implies this code does.

The PR description says:
`We should force to provide lit network when initialising lit auth client, and we shouldn't default to cayenne`
However, the current PR doesn't actually _force_ consumers to provide the lit network (it's optional), _and_ it still implictly defaults to `cayenne` in cases where they don't provide a litNetwork property.

The fact that we're even talking about this change this much IMO is a clear sign that we should just _let the consumer create the client instance and pass it in here_.  The current API is just extra mental overhead for people to grok, and adding a `litNetwork` property here is even _worse_.  What happens if I pass a `custom` lit node client configured for `cayenne`, but then I provide the _optional_ argument with `litNetwork` set to `manzano`?  As it turns out, I think that the new `litNetwork` argument will be _ignored_ if I provide a custom client, but how am I supposed to know that? As a consumer of this API I have no idea what will happen in that case without reading the source code.

Ansonhkg:
better solution incoming

PR #427, opened 2024-04-15
LIT-2841 - Remove `ipfs-http-client` and create encryption APIs for composition with any storage layer

MaximusHaximus:
# Description

### Overview
This PR removes a problematic dependency that also happened to be very large code and dependency-wise - `ipfs-http-client`.  We were using this in a single place in our SDK as part of a 'nice-to-have' function.  Unfortunately, the package has issues with bundling in multiple environments and brings a very large # of dependencies into our bundle that a lot of our users never actually needed.

The API for the method that used this dependency was also problematic; it was hard-coded to use Infura credentials and endpoints, and Infura has disabled the ability to create IPFS-enabled API keys since November of 2023.  Rather than re-work the API to allow arbitrary providers, authentication, etc, we decided that it would be more flexible for consumers, and better for the size of our bundles and compatibility in multiple environments, to remove the dependency and instead provide some functions that are un-opinionated about where the encrypted data is going or coming from. 

Anyone who would like to encrypt to/from IPFS can simply call these new methods and then use the client _of their choice_ to put the data in IPFS.

Even better, anyone who wants to encrypt/decrypt somewhere _other_ than IPFS, but without compressing the data using our existing APIs that encrypt into ZIP files with metadata stored inside the ZIP archive can use these same methods to work with any other arbitrary data store they want to use.

Before this change, the only code that would wrap up the encrypted data with the metadata required to easily decrypt it later was the 'encryptToIpfs()' code which stored the encrypted data and metadata in the same JSON blob, always on IPFS, or `encryptFileAndZipWithMetadata()` code which stored the metadata as a separate file from the encrypted data in the ZIP archive and only worked with files (not bare strings).

### Changes:
- Removed `encryptToIpfs()` and `decryptFromIpfs()` methods in `encryption`, which were also exported from `lit-node-client`
- Added `encryptToJson()` and `decryptFromJson()` in `encryption`, and export those methods from `lit-node-client`
    - These methods produce and consume JSON blobs in the same format as the previous `encryptToIpfs()` and `decryptFromIpfs()` methods produced and consumed.

### Fixes:
- Resolved an incorrectly applied polyfill for `fetch`, where we were mutating global fetch _even if it was previously defined_. 
    - We now use `cross-fetch` to handle this polyfill, which is friendly to environments where it is already defined -- and was already part of our dependency tree due to other dependencies we already rely on.
    - Removed `node-fetch` direct dependency now that `cross-fetch` is being used to polyfill.
- Removed `ipfs-htttp-client` dependency, which was causing bundling errors in mixed environments (NextJS)

### Internal changes:
- Removed build tooling hooks for post-build mutations that were we were not running and no longer used
- Cleaned up some duplication in types touching access control conditions
- Fixed a bunch of eslint/tslint errors in the files where changes were made

## Type of change

# How Has This Been Tested?

- [X] Added tests for encrypting/decrypting from JSON strings using both file input and string input
- [X] Ran entire E2E suite

# Checklist:


package.json
```diff
@@ -199,9 +199,7 @@
     "ts-jest": "29.1.2",
     "ts-node": "10.9.1",
     "typedoc": "^0.23.10",
-    "typescript": "~4.7.2",
-    "@nx/next": "17.3.0",
-    "@nx/node": "17.3.0"
```
joshLong145, line 204: Why did we remove these packages?

packages/encryption/src/lib/encryption.ts
```diff
@@ -182,49 +135,39 @@ export const decryptFromIpfs = async (
...
+        ciphertext: parsedJsonData.ciphertext,
+        dataToEncryptHash: parsedJsonData.dataToEncryptHash,
+        chain: parsedJsonData.chain,
+        authSig,
+        sessionSigs,
+      },
+      litNodeClient
+    );
+  } else {
+    return decryptToFile(
```
joshLong145, line 156: Do we always want to default to the file?

package.json
MaximusHaximus: Actually, they didn't get removed, but Yarn automatically alphabetizes the entries in `package.json` (it reduces merge conflicts), and I think someone previously had manually added the entries to the bottom of the file instead of using the yarn add command to add them, so when I added/removed _other_ packages, it re-sorted them to the correct place in the file, making it appear they were removed by me.

packages/encryption/src/lib/encryption.ts
MaximusHaximus: The `dataType` currently can only be `string` or `file` (it's a string union type), so currently the only possible option if `dataType` is not `string` is that it must be `file` :)

joshLong145: yes, it might be worth adding an explicit check on the `dataType` so if in the future we add more to the type union it will not default if someone doesn't update this conditional check.

MaximusHaximus: Roger that :+1: :) -- I updated this to make the check explicit and to explicitly throw an error if the value isn't `string` or `file` -- note that we're also validating it in the `decryptFromJson` params validator, so someone will also need to modify that validator to support the new string before this `else{}` logic will throw the new exception.

joshLong145 (dismissed):

MaximusHaximus:
@joshLong145 I went ahead and made the `connect()` logging in the contracts SDK less spammy like we talked about -- the contents of `abi` were just arrays of numbers, so I create a map of contract name->address that looks like this:
```
resolved contract addresses for:  cayenne {
  StakingBalances: '0x095251de2aD2A78aDe96F2a11F7feAA7CF93e6B5',
  Staking: '0x5bFa704aF947b3b0f966e4248DED7bfa6edeF952',
  Multisender: '0xD4e3D27d21D6D6d596b6524610C486F8A9c70958',
  LITToken: '0x53695556f8a1a064EdFf91767f15652BbfaFaD04',
  PubkeyRouter: '0x4B5E97F2D811520e031A8F924e698B329ad83E29',
  PKPNFT: '0x58582b93d978F30b4c4E812A16a7b31C035A69f7',
  RateLimitNFT: '0x19593CbBC56Ddd339Fde26278A544a25166C2388',
  PKPHelper: '0xF02b6D6b0970DB3810963300a6Ad38D8429c4cdb',
  PKPPermissions: '0xD01c9C30f8F6fa443721629775e1CC7DD9c9e209',
  PKPNFTMetadata: '0xeD46dDcbFF662ad89b0987E0DFE2949901498Da6',
  Allowlist: '0xfc7Bebd150b36921549595A776D7723fBC4Bb2D9'
}
```

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -392,13 +390,26 @@ export class LitContracts {
...
+    const logAddresses = Object.entries(addresses).reduce(
+      (output, [key, val]) => {
+        // @ts-expect-error since the object hash returned by `getContractAddresses` is `any`, we have no types here
+        output[key] = val.address;
+        return output;
+      },
+      {}
+    );
+
+    console.log(
```
joshLong145, line 408: can we use `this.log` over `console.log`

joshLong145 (approved):

packages/contracts-sdk/src/lib/contracts-sdk.ts
MaximusHaximus: All set! Rebased to eliminate conflicts also.

joshLong145 (dismissed):

packages/encryption/src/lib/encryption.ts
```diff
@@ -182,49 +135,41 @@ export const decryptFromIpfs = async (
...
-        litNodeClient
-      );
-    }
-  } catch (e) {
-    return throwError({
-      message: 'Unable to fetch or decrypt from IPFS',
-      errorKind: LIT_ERROR.INVALID_ARGUMENT_EXCEPTION.kind,
-      errorCode: LIT_ERROR.INVALID_ARGUMENT_EXCEPTION.name,
-    });
+  // FIXME: The return type of this function is inferrable based on the value of `params.dataType`
```
jtary, line 138: Here the type sig if you want fix this:
```
export function decryptFromJson<T extends DecryptFromJsonProps>(
  params: T
): T extends { parsedJsonData: { dataType: 'file' } }
  ? ReturnType<typeof decryptToFile>
  : T extends { parsedJsonData: { dataType: 'string' } }
  ? ReturnType<typeof decryptToString>
  : never;
export function decryptFromJson<T extends DecryptFromJsonProps>(
  params: T
) {
```

jtary: You need to overload the function to get the inferred return to work with a discriminated union, so you can't use an anonymous function as far as I know 🤷

MaximusHaximus: 💖 That's awesome. I'll add that now.
Named functions are better for stack traces anyway :+1: I copypasta'd some existing code without thinking... have been trying to migrate to using named functions as I go along, thanks for the reminder

packages/encryption/src/lib/encryption.ts
```diff
@@ -182,49 +135,41 @@ export const decryptFromIpfs = async (
...
+        ciphertext: parsedJsonData.ciphertext,
+        dataToEncryptHash: parsedJsonData.dataToEncryptHash,
+        chain: parsedJsonData.chain,
+        authSig,
+        sessionSigs,
+      },
+      litNodeClient
+    );
+  } else {
+    throw new Error(`Datatype "${parsedJsonData.dataType}" is not supported.`);
```
Ansonhkg, line 172: Can we also log the supported types out?

packages/sev-snp-utils-sdk/src/lib/sev-snp-utils-sdk.spec.ts
```diff
@@ -1,13 +1,6 @@
-// @ts-nocheck
-import { TextEncoder, TextDecoder } from 'util';
-import * as fetch from 'node-fetch';
-
-global.TextEncoder = TextEncoder;
-// @ts-ignore
-global.TextDecoder = TextDecoder;
+import fetch from 'cross-fetch';
```
Ansonhkg, line 1: Have we tried if this would work happily with js frameworks? eg. nextjs and vite

Ansonhkg (changes requested):

packages/lit-node-client-nodejs/src/index.ts
```diff
@@ -1,14 +1,11 @@
-import { isNode, log } from '@lit-protocol/misc';
+import 'cross-fetch/polyfill';
```
joshLong145, line 1: do we think its worth it to check after the polyfill loading for `fetch.polyfill` to be defined? 
from looking at their implementations they do set this `boolean` to true if its found a polyfill is needed.
https://github.com/lquixada/cross-fetch/blob/v4.x/src/node-polyfill.js

MaximusHaximus: What should we do with that information if we check for it?  I'm not sure it's relevant for us, though I guess a consumer might want to check it for some reason?

packages/encryption/src/lib/encryption.ts
MaximusHaximus: Got it to work without the overload and with ReturnType<> :tada:.  Just had to use Promise<> to wrap the entire thing - guessing the weirdness was because the other methods were marked `async` and the error message we saw was misleading :+1:

packages/sev-snp-utils-sdk/src/lib/sev-snp-utils-sdk.spec.ts
MaximusHaximus: :+1: Updated this to return the same string as our `paramsValidator` does, which mentions the possible strings

packages/encryption/src/lib/encryption.ts
MaximusHaximus: :+1: Updated this to return the same string as our `paramsValidator` does, which mentions the possible strings

packages/sev-snp-utils-sdk/src/lib/sev-snp-utils-sdk.spec.ts
MaximusHaximus: Yes, this is actually why I replaced our manual polyfill, which was not checking to see if it was already defined and was overwriting the existing version when it shouldn't be -- it was _always_ overwriting the existing `fetch` implementation, in both nextjs and nodejs where fetch is natively supported, even when it was already defined.  `cross-fetch` will only inject `node-fetch` into the environment if `global.fetch` is not already defined.  This version of `cross-fetch` depends on the same version of `node-fetch` that we were manually requiring and always injecting on to `global`, so in cases where `cross-fetch` _does_ inject, it will be the same implementation we previously relied on.

packages/lit-node-client-nodejs/src/index.ts
joshLong145: if its false for some reason then i think we might want to warn that the polyfill might have failed. so we could do our own check on the global for fetch and conditionally import, and then check if `polyfill` is set to true. The purpose is to just warn end consumers that fetch might have failed polyfill.

Ansonhkg (dismissed):
lgtm!

joshLong145 (dismissed):
Approving as the polyfill has been tested.

joshLong145 (approved):

Ansonhkg (approved):
lgtm

PR #428, opened 2024-04-15
feat(AMD Cert): Always fetch from CORS

Ansonhkg:
# Description

AMD certificate endpoints are rate limited, so we are going to always use our Lit CORS server caches certificates locally to avoid hitting the rate limit.

## Type of change

# How Has This Been Tested?

## Test Command

```
DEBUG=true yarn test:e2e:node --filter=multi-network          
```
## Result
image

# Checklist:

### DIFF
--- a/packages/crypto/src/lib/crypto.ts
+++ b/packages/crypto/src/lib/crypto.ts
@@ -377,28 +377,30 @@ async function getAmdCert(url: string): Promise<Uint8Array> {
     `[getAmdCert] Fetching AMD cert using proxy URL ${proxyUrl} due to CORS restrictions.`
   );
 
-  let response;
-
-  try {
-    response = await fetch(proxyUrl);
-    if (!response.ok) {
+  async function fetchAsUint8Array(targetUrl) {
+    const res = await fetch(targetUrl);
+    if (!res.ok) {
       throw new Error(`[getAmdCert] HTTP error! status: ${response.status}`);
     }
-  } catch (e) {
-    log(`[getAmdCert] Failed to fetch AMD cert from proxy:`, e);
-
-    // Try direct fetch only if proxy fails
-    log('[getAmdCert] Attempting to fetch directly without proxy.');
-    try {
-      response = await fetch(url);
-      if (!response.ok) {
-        throw new Error(`HTTP error! status: ${response.status}`);
-      }
-    } catch (e) {
-      log('[getAmdCert] Direct fetch also failed:', e);
-      throw e; // Re-throw to signal that both methods failed
-    }
+    const arrayBuffer = await res.arrayBuffer();
+    return new Uint8Array(arrayBuffer);
   }
+
+   try {
+    return await fetchAsUint8Array(proxyUrl);
+  } catch (e) {
+      log(`[getAmdCert] Failed to fetch AMD cert from proxy:`, e);
+  }
+  
+  // Try direct fetch only if proxy fails
+  log('[getAmdCert] Attempting to fetch directly without proxy.');
+  try {
+    return await fetchAsUint8Array(url);
+  } catch (e) {
+    log('[getAmdCert] Direct fetch also failed:', e);
+    throw e; // Re-throw to signal that both methods failed
+  }
+
 
   const arrayBuffer = await response.arrayBuffer();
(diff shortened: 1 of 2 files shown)


packages/crypto/src/lib/crypto.ts
```diff
@@ -359,14 +358,44 @@ function base64ToBufferAsync(base64) {
...
+ * 
+ * @param url The URL from which to fetch the AMD cert.
+ * @returns A Promise that resolves to a Uint8Array containing the AMD certification data.
+ * @throws An error detailing HTTP or network issues encountered during the fetch process.
+ */
+async function getAmdCert(url: string): Promise<Uint8Array> {
+
+  const proxyUrl = `${LIT_CORS_PROXY}/${url}`;
+
+  console.log(`Fetching AMD cert using proxy URL due to CORS restrictions.`);
```
joshLong145, line 374: should we use `log` instead of `console`
also i think we should log out the `url` from the attestation report so users may audit the cert which was fetched through the proxy

joshLong145 (approved):

packages/crypto/src/lib/crypto.ts
```diff
@@ -359,14 +361,45 @@ function base64ToBufferAsync(base64) {
...
+ *
+ * @param url The URL from which to fetch the AMD cert.
+ * @returns A Promise that resolves to a Uint8Array containing the AMD certification data.
+ * @throws An error detailing HTTP or network issues encountered during the fetch process.
+ */
+async function getAmdCert(url: string): Promise<Uint8Array> {
+  const proxyUrl = `${LIT_CORS_PROXY}/${url}`;
+
+  log(
+    `[getAmdCert] Fetching AMD cert using proxy URL ${proxyUrl} due to CORS restrictions.`
```
MaximusHaximus, line 377: This is no longer only used to bypass CORS, so if we're gonna log and JSDoc this, we should mention that we always call it to avoid being rate limited by AMD in these logs and update the JSDoc content to describe that we now _always_ use this proxy to bypass rate limiting

packages/crypto/src/lib/crypto.ts
```diff
@@ -359,14 +361,45 @@ function base64ToBufferAsync(base64) {
...
+    try {
+      response = await fetch(url);
+      if (!response.ok) {
+        throw new Error(`HTTP error! status: ${response.status}`);
+      }
+    } catch (e) {
+      log('[getAmdCert] Direct fetch also failed:', e);
+      throw e; // Re-throw to signal that both methods failed
+    }
   }
```
MaximusHaximus, line 401: Would be nice to unravel the nested try/catch and local variable mutation here so that the code flow is linear and we don't duplicate error handling. Maybe something like... (caution: I did not run this code, `fetchAsUint8Array()` should be in module scope, and I couldn't include deleted lines in the suggestion so the existing arrayBuffer handling would need to be manually deleted to apply this code change)
```suggestion
  async function fetchAsUint8Array(targetUrl) {
    const res = await fetch(targetUrl);
    if (!res.ok) {
      throw new Error(`[getAmdCert] HTTP error! status: ${response.status}`);
    }
    const arrayBuffer = await res.arrayBuffer();
    return new Uint8Array(arrayBuffer);
  }

   try {
    return await fetchAsUint8Array(proxyUrl);
  } catch (e) {
      log(`[getAmdCert] Failed to fetch AMD cert from proxy:`, e);
  }
  
  // Try direct fetch only if proxy fails
  log('[getAmdCert] Attempting to fetch directly without proxy.');
  try {
    return await fetchAsUint8Array(url);
  } catch (e) {
    log('[getAmdCert] Direct fetch also failed:', e);
    throw e; // Re-throw to signal that both methods failed
  }

```

This structure also has the added bonus that if building the array buffer or converting it to a Uint8Array throws an error during the first attempt against the proxy, we will still try to fetch directly from the AMD service instead of throwing the error to the caller.

MaximusHaximus (changes requested):
Added a few things in-line 💖 

Just out of curiosity, do we know what the rate limits are or how they are enforced (e.g. by IP address?).

joshLong145 (approved):

MaximusHaximus (approved):
:+1: LGTM! 🚢 :shipit:

PR #429, opened 2024-04-16
(experiment) feat(test): setup jest testing environment

Ansonhkg:
# Description

This PR serves as an experiment for testing e2e tests with `jest` and `bun`. In this experiment, we are testing the connection to multiple networks.

## Jest test

- No hot reload - It requires rebuilding each time if you want to see the changes reflected in the tests. This issue may be due to misconfiguration or may be specific to a different workspace.

### Test Command

```
npx jest './e2e-tests/jest-e2e.test.ts' -c './e2e-tests/jest.config.ts' -t 'connections'
```

### Result
image

### Test Command

```
DEBUG=true bun test bun-e2e.test.ts --timeout 20000
```

- ot reload out of the box - No rebuild required. Changes are immediately reflected in the tests.

### Result
image

## Type of change


MaximusHaximus:
Nice work, thanks for digging into this stuff.

Thinking about the current pattern, I think that if you're running into issues that are specific to these tests existing in the root of the repo instead of as a package, it would be fine to just define these e2e tests _as another package in the same workspace_ -- there's really no requirement that e2e tests live in the root of the repository, and we don't have to publish it anywhere.

By making the e2e tests a package, we can also let NX handle re-building cross-package dependencies for us with caching as things change in the dependencies that our e2e tests rely on. Also, because `nx` provides a 'native' jest runner that integrates tightly with nx, this might resolve some of the other issues you're experiencing, like no hot reload or whatever configuration issue was causing you to need to use `bun`.  I know that Jest is correctly handling on-the-fly transpilation for other packages in the repo, as they already use `ts-jest`... since ts-jest provides on-the-fly transpilation in its base feature set, we shouldn't need `bun`, and more importantly, because `bun` doesn't have feature parity with nodejs, it would be best to avoid using it as `bun` might break at any time in the future if we need to start relying on a nodejs function that it doesn't support.

PR #434, opened 2024-04-22
chore: update README with new NodeJS version

joshLong145:
changes `node` version requirements to `19` to support `crypto` being an included module by default in `global`

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ If you're a tech-savvy user and wish to utilize only specific submodules that ou
 
 ## Prerequisite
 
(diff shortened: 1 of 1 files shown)


joshLong145:
@Ansonhkg @DashKash54
Curious how you think we should handle this. Thinking we could also just require `node v20` as it is a named LTS release. Testing has found it to support global crypto out of the box.

joshLong145:
Closing as for implementations directly in our SDK packages we do not require a version bump to the runtime. Will be handled with documentation updates.

PR #435, opened 2024-04-23
Auth Unification, Custom Auths, better tests 

Ansonhkg:
# Description

This PR introduces Tinny, an improved mini e2e testing framework that allows us to write tests in TypeScript. It leverages `esbuild` for rapid test compilation and execution. Tinny includes typed test helper functions to generate new identities within tests, and provides APIs to modify the development environment.

See docs:
https://github.com/LIT-Protocol/js-sdk/tree/feature/lit-2958-auth-unification-ts-tests/local-tests

# Note

We are not merging this PR to master until all related PRs are merged to this branch. 
- https://github.com/LIT-Protocol/js-sdk/pull/438
- https://github.com/LIT-Protocol/js-sdk/pull/440
- https://github.com/LIT-Protocol/js-sdk/pull/439
- https://github.com/LIT-Protocol/js-sdk/pull/436

Cherry Picked:
- https://github.com/LIT-Protocol/js-sdk/pull/446

We should also raise a PR to remove all existing V0 features when we are ready to deploy. They are now being tracked [here](https://linear.app/litprotocol/issue/LIT-3032/[auth-unification]-remove-all-v0-stuff-after-we-deployed-v1)

## Changes
- [x] **Local Network Context**: Fixed issues to ensure local network settings are accurately recognized during test runs.
- [x]  **Helper Functions**: Introduced new helper functions for `siweMessage` creation and `authSig` crafting, addressing circular dependencies in the `auth-helpers` package.
- [x] **Fetch Polyfill**: Implemented a temporary fix to ensure the fetch API is polyfilled correctly when bundling the tests.
- [x] **Interface Updates**: Restored the litNodeClient to the ILitNodeClient interface to ensure type consistency.
- [x] **Auth unification params**: Expanded `authNeededCallbacks` to support additional parameters and introduced a - higher-level function, `getPkpSessionSigs`, to streamline PKP sessionSigs creation process.
- [x] **Type Migration**: Migrated type definitions to the types package to enhance modularity and maintenance.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

https://github.com/LIT-Protocol/js-sdk/assets/4049673/9216c2ce-9774-45b4-92ed-3f666c00470f

# Checklist:

- [x] My code follows the style guidelines of this project
[...]

PR #436, opened 2024-04-23
Feature/lit 2962 auth unification better siwe authsig creation

Ansonhkg:
# Description

Previously, we had many conditional/optional parameters when creating a SIWE message, which led to confusion about its inputs and outputs. This PR introduces helper functions tailored to users' intentions.

1. `createSiweMessage` - A general function that allows users to create any type of SIWE message, ideal for advanced users. 
2. `createSiweMessageWithRecaps` - Using `createSiweMessage` under the hood, but adding recap resources to the SIWE message.
3. `createSiweMessageWithCapacityDelegation` - Using `createSiweMessage` under the hood, but setting the uri as `lit:capability:delegation` and different optional parameters

These helper functions can be found [here](https://github.com/LIT-Protocol/js-sdk/tree/63a3ffc1dc63f0e15fbe72bb09cf111e1b0bcd74/packages/auth-helpers/src/lib/siwe) and [here](https://github.com/LIT-Protocol/js-sdk/blob/63a3ffc1dc63f0e15fbe72bb09cf111e1b0bcd74/packages/auth-helpers/src/lib/craft-auth-sig.ts)

Note that the actual changes are in [feature/lit-2958-auth-unification-ts-tests](https://github.com/LIT-Protocol/js-sdk/pull/435) cus they are needed for the tests to work. However, in this PR, we are actually using them in the core component.

- [x] (Breaking) Renaming `SIG_TYPE` to `LIT_CURVE`
- [x] Added types to facilitate new shape of JSON requests to node
- [x] Replace operations to use the following helpers:

```ts
import {
  craftAuthSig,
  createSiweMessageWithCapacityDelegation,
  createSiweMessageWithRecaps,
  createSiweMessage
} from '@lit-protocol/auth-helpers';
```

## Type of change

# How Has This Been Tested?

All tests will be provided in a separated PR

# Checklist:


packages/types/src/lib/interfaces.ts
```diff
@@ -300,6 +309,31 @@ export interface JsonSignChainDataRequest {
...
+  result: boolean;
+  signatureShare: {
+    ProofOfPossession: string;
+  };
+  shareIndex: number;
+  curveType: string;
+  siweMessage: string;
+  dataSigned: string;
+  blsRootPubkey: string;
+}
```
joshLong145, line 335: nice, we have a new model for this in the wasm combination branch but for now this is better than what we have. making a note to unify with the wasm combination work.

joshLong145 (approved):

glitch003 (approved):
looks great, thanks.  this gets us BLS session sigs too, right?

DashKash54:
> createSiweMessageWithRecaps - Using createSiweMessage under the hood, but adding recap resources to the SIWE message.

Doesn't all the Siwe messages require resource to be allowed from the user? @Ansonhkg

packages/misc/src/lib/misc.ts
```diff
@@ -859,3 +859,41 @@ export async function executeWithRetry<T>(
     requestId,
   };
 }
+
+/**
+ * Attempts to normalize a string by unescaping it until it can be parsed as a JSON object,
+ * then stringifies it exactly once. If the input is a regular string that does not represent
+ * a JSON object or array, the function will return it as is without modification.
+ * This function is designed to handle cases where strings might be excessively escaped due
```
DashKash54, line 867: When do we encounter this situation?

packages/misc/src/lib/misc.ts
```diff
@@ -859,3 +859,41 @@ export async function executeWithRetry<T>(
...
+ * to multiple layers of encoding, ensuring that JSON data is stored in a consistent and
+ * predictable format, and regular strings are left unchanged.
+ *
+ * @param input The potentially excessively escaped string.
+ * @return A string that is either the JSON.stringify version of the original JSON object
+ *         or the original string if it does not represent a JSON object or array.
+ */
+export function normalizeAndStringify(input: string): string {
```
DashKash54, line 875: Can we have a unit test for this?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2487,75 +2442,50 @@ export class LitNodeClientNodeJs
...
-          params?.domain || globalThis.location?.host || 'litprotocol.com',
-        address: pkpEthAddress,
-        statement: siwe_statement,
-        uri: sessionKeyUri,
-        version: '1',
-        chainId: params.chainId ?? 1,
-        expirationTime: _expiration,
-        resources: params.resources,
-        nonce: this.latestBlockhash!,
+    if (params.resourceAbilityRequests) {
```
DashKash54, line 2461: SessionSigs should always have resources?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2609,74 +2539,215 @@ export class LitNodeClientNodeJs
...
     // ========== Extract shares from response data ==========
     // -- 1. combine signed data as a list, and get the signatures from it
-    const signedDataList = responseData.map(
-      (r: any) => (r as SignedData).signedData
-    );
+    let curveType = responseData[0]?.curveType;
+
+    if (!curveType) {
+      log(`[signSessionKey] curveType not found. Defaulting to ECDSA.`);
+      curveType = 'ECDSA';
```
DashKash54, line 2552: signSessionKey endpoint can't be ECDSA. Replace all the ECDSA logic from below? 
cc: @glitch003

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2609,74 +2539,215 @@ export class LitNodeClientNodeJs
...
+      const err = `[signSessionKey] signedDataList is empty.`;
+      log(err);
+      throw new Error(err);
+    }
+
+    logWithRequestId(
+      requestId,
+      '[signSessionKey] signedDataList',
+      signedDataList
+    );
```
DashKash54, line 2589: You're combing the BLS shares into the final signature? Extract this out in a function and add unit tests for it? How can we be sure whether the BLS share combination is correct?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
DashKash54, line 1: We no longer allow ECDSA signatures in PKP sessionSigs: https://github.com/LIT-Protocol/lit-assets/blob/8d26f49faf8d1648fb50deb9ff8d45a68ed49f80/rust/lit-node/src/endpoints/versions/v1.rs#L47
So we should remove all the ECDSA logic for PKP SessionSig

DashKash54 (changes requested):
Left a major comment

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: This is needed when you want to test all existing V0 networks and V1. A separated PR will be raised to remove all of these once we are ready to merge this in to master

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: This is needed when you want to test all existing V0 networks and V1. A separated PR will be raised to remove all of these once we are ready to merge this in to master

packages/misc/src/lib/misc.ts
Ansonhkg: in the signed siwe message

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: don't think you need that on cayenne?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: https://linear.app/litprotocol/issue/LIT-3032/[auth-unification]-remove-all-v0-stuff-after-we-deployed-v1

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: That makes sense. Will do that

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -180,115 +174,57 @@ export class LitNodeClientNodeJs
...
-      dAppOwnerWallet,
-      capacityTokenId,
-      delegateeAddresses,
-      uses,
-      domain,
-      expiration,
-      statement,
-    } = params;
+    // Useful log for debugging
+    if (!params.delegateeAddresses || params.delegateeAddresses.length === 0) {
```
MaximusHaximus, line 178: Not sure this log is accurate -- I think that `delegateeAddresses` being provided as an _empty array_ actually means that **_nobody** can use it_, but _no delegateeAddresses at all_ means that _anyone can use it_, right?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -180,115 +174,57 @@ export class LitNodeClientNodeJs
...
-      expirationTime: _expiration,
+    const siweMessage = await createSiweMessageWithCapacityDelegation({
+      uri: 'lit:capability:delegation',
+      litNodeClient: this,
+      walletAddress: dAppOwnerWalletAddress,
+      nonce: nonce,
+
+      // -- default configuration for recap object capability
+      expiration:
+        params.expiration ?? new Date(Date.now() + 1000 * 60 * 7).toISOString(),
```
MaximusHaximus, line 210: Can we remove the defaults here and let the defaults defined inside of `createSiweMessage()` do the defaulting?   

It's great that we have a central place defined for creating messages; can we let all defaulting be delegated to that method?  

For example, If we need to default `domain` to `example.com` here instead of letting he `createSiweMessage()` function provide the default, can we add a comment here as to _why_ we want `example.com` to be the default in this codepath, but `localhost` to be the default inside the `createSiweMessage()` method?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -741,12 +677,12 @@ export class LitNodeClientNodeJs
...
-      ipfsId,
-      jsParams,
-      authMethods,
+    const data: JsExecutionRequestBody = {
+      ...(authSig ? { authSig } : {}),
+      ...(code ? { code } : {}),
+      ...(ipfsId ? { ipfsId } : {}),
+      ...(authMethods ? { authMethods } : {}),
+      ...(jsParams ? { jsParams } : {}),
     };
```
MaximusHaximus, line 686: Just curious why we needed to change this to use this pattern

MaximusHaximus (changes requested):
Comments, observations and questions added in-line!

My main feedback is that I think that we would benefit greatly from just defining a new `signSessionKey()` method specific to BLS; there are so many conditional branches in that method that it is making things quite complex to reason about (and provide strong types for).

The idea would be to rename existing `signSessionKey()` to `signSessionKeyECDSA()`, make new method with _just the BLS logic_ called `signSessionKeyBLS()`, implement new method called `signSessionKey()` that calls the appropriate version of the method based on the `curveType`.

packages/lit-node-client-nodejs/src/lib/helpers/handle-bls-response.ts
```diff
@@ -0,0 +1,26 @@
+import { log } from '@lit-protocol/misc';
+import { BlsResponseData } from '@lit-protocol/types';
+
+/**
+ * Handles the response data from BLS  signature scheme.
+ * @param responseData - The response data from BLS signature scheme.
+ * @returns An array of signed data.
+ * @throws Error if no data is provided.
+ */
+export function handleBlsResponseData(
```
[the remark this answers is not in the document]
Ansonhkg: The interface `BlsResponseData` is manually defined instead of automatically generated, in this case the unit test would at least have the original structure of the data when `handleBlsResponseData` throws an error

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: Yea, I think you are technically correct if you pass an empty array to the rust function directly, I need to double check how it actually looks like in this siwe message after using the recap lib to add this auth sig in.

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: I was checking if the node would accept the request if a key is undefined or when it's omitted in the new endpoint. but i have created a separate interface for the request body in another pr

packages/misc/src/lib/misc.ts
DashKash54: So the node returns the message containing nested "\"? How do we handle it currently? Shouldn't deserializing into Json take care of it automatically?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
DashKash54: Any updates @Ansonhkg ?

packages/misc/src/lib/misc.ts
DashKash54: @MaximusHaximus what do you think? ^

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
DashKash54: Do we have unit tests for it now?

packages/constants/src/lib/enums.ts
```diff
@@ -5,7 +5,7 @@ export enum VMTYPE {
   CVM = 'CVM',
 }
 
-export enum SIGTYPE {
+export enum LIT_CURVE {
```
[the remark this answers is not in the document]
DashKash54: This isn't use facing?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
DashKash54: > no delegateeAddresses at all means that anyone can use it, right

Yes that's the idea. Might be worth to have a test for it as many users wanted this feature

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
DashKash54: Can we do a constructor type pattern or something similar which set the default values if none is provided?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -174,10 +174,15 @@ export class LitNodeClientNodeJs
...
+    // -- validate
+    if (!params.dAppOwnerWallet) {
+      throw new Error('dAppOwnerWallet must exist');
+    }
+
     // Useful log for debugging
     if (!params.delegateeAddresses || params.delegateeAddresses.length === 0) {
       log(
-        `[createCapacityDelegationAuthSig] No delegatee addresses provided. It means that the capability will not restrict access based on delegatee list, but it may still enforce other restrictions such as usage limits (uses) and specific NFT IDs (nft_id).`
+        `[createCapacityDelegationAuthSig] 'delegateeAddresses' is an empty array. It means that no body can use it. However, if the 'delegateeAddresses' field is omitted, It means that the capability will not restrict access based on delegatee list, but it may still enforce other restrictions such as usage limits (uses) and specific NFT IDs (nft_id).`
```
DashKash54, line 185: Are you sure about this? I think an empty array would mean anyone can use it?

DashKash54 (approved):
Looks good!

packages/types/src/lib/interfaces.ts
```diff
@@ -322,8 +322,43 @@ export interface JsonSignSessionKeyRequestV1 {
...
+//     "dataSigned": "b2efe867176b9212fd6acd39a33004a17e03d5a931250c700e31af95e2e7e4d5",
+//     "blsRootPubkey": "a6f7c284ac766db1b43f8c65d8ff15c7271a05b0863b5205d96459fd32aa353e9390ce0626560fb76720c1a5c8ca6902"
+//   }
+// ]
 export interface BlsResponseData {
-  result: boolean;
+  result: boolean | 'success';
```
DashKash54, line 361: This is weird, should be only either a bool or string like "success"/"failure"

DashKash54:
I'm approving this PR, although there are still comments unresolved. Might be worth responding & resolving it. For certain comments we may also defer addressing them later but we should track them in Linear @Ansonhkg

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: though i'm not sure if there's a use case where you would generate a delegation auth sig where nobody uses it lol

so @DashKash54 

```ts
const params = {
  delegateeAddresses: [] // <-- nobody can use it
};
```

```ts
const params = {
  // delegateeAddresses: [] // <-- commented out, key is omitted, meaning anyone can use it
}
```

packages/misc/src/lib/misc.ts
Ansonhkg: Yea, the problem is we do not know how many times the json has been serialised, there were cases/bugs before where we would receive a json that has been serialised twice, and we spent ages to figure out why wasn't it working because the value looked correct, ended up we had to deserialise it twice.

MaximusHaximus: It feels very odd for us to need to be handling specific characters with bespoke string replacement calls in data that should be JSON (or any other standard encoding) encoded instead of just _using the decoder_ to decode it recursively.

If the problem is double-json-serialized data, could we instead check to see if the result of a `json.parse()` call returns a `string` where it should be an `object` (or `array`, or whatever we expected), then just run `parse()` on the string again to get what we need?

Ansonhkg: Actually, I'm gonna check if this is still the case. Ideally, we should always receive a JSON string that has been serialised only once, and we shouldn't do any magic here.

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
DashKash54: This is fixed

PR #437, opened 2024-04-24
feat: allow cert caching if local storage is defined

joshLong145:
# Description

Changes a conditional check for storing AMD SEV SNP certificates in `localstorage` to now check if the api is available rather than doing a check for a browser environment. this will allow for caching in local storage in non browser environments if the api is available. 

## Type of change

# How Has This Been Tested?

A new `connection` e2e test has been added which defines a `localstorage` provider and checks that certs have been cached after `connect` finishes.

# Checklist:

### DIFF
--- a/e2e-nodejs/group-connection/test-connection-cert-caching.mjs
+++ b/e2e-nodejs/group-connection/test-connection-cert-caching.mjs
@@ -1,0 +1,53 @@
+import path from 'path';
+import { success, fail, testThis, log } from '../../tools/scripts/utils.mjs';
+import { LitNodeClient } from '@lit-protocol/lit-node-client';
+import { LocalStorage } from 'node-localstorage';
+export async function main() {
+  const networks = ['habanero', 'manzano'];
+  const storageProvider = new LocalStorage('./storage.test.db');
+  for (const network of networks) {
+    // ==================== Test Logic ====================
+    const client = new LitNodeClient({
+      litNetwork: network,
+      debug: globalThis.LitCI.debug,
+      storageProvider: {
+        provider: storageProvider,
+      },
+    });
+    log(`connecting to ${network.toUpperCase()}...`);
+    await client.connect();
+
+    // ==================== Post-Validation ====================
+    if (!client.ready) {
+      return fail('client not ready');
+    }
+    if (client.config.litNetwork !== network) {
+      return fail(`client not connected to ${network}`);
+    }
+
+    if (storageProvider.length < 1) {
+      fail(`cache is not hydrated with certificates`);
+    }
+
+    if (storageProvider.length < client.config.bootstrapUrls) {
+      fail(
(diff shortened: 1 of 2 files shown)


joshLong145:
> Nice! e2e seems to be failing though with your cert test
> 
> ```
> ❌ 1 test(s) failed
> -------------------
> - [1] Error happened in /home/runner/work/js-sdk/js-sdk/e2e-nodejs/group-connection/test-connection-cert-caching.mjs, see below for details 👇 
> Error: missing revert data in call exception; Transaction reverted without a reason string [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (data="0x", transaction={"to":"0xBC7F8d7864002b6629Ab49781D5199C8dD1DDcE1","data":"0x43cb0a0e","accessList":null}, error={"reason":"missing response","code":"SERVER_ERROR","requestBody":"{\"method\":\"eth_call\",\"params\":[{\"to\":\"0xbc7f8d7864002b6629ab49781d5199c8dd1ddce1\",\"data\":\"0x43cb0a0e\"},\"latest\"],\"id\":45,\"jsonrpc\":\"2.0\"}","requestMethod":"POST","serverError":{"errno":-104,"code":"ECONNRESET","syscall":"read"},"url":"https://chain-rpc.litprotocol.com/http"}, code=CALL_EXCEPTION, version=providers/5.7.2)
> ```

Did a re run of that job since the error looked to be an rpc timeout issue. The `e2e` tests now pass.

Ansonhkg (approved):
lgtm

e2e-nodejs/group-connection/test-connection-cert-caching.mjs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    if (storageProvider.length < client.config.bootstrapUrls) {
+      fail(
+        `cache is not hydrated with enough certificates found: ${storageProvider.length} need ${client.config.bootstrapUrls.length}`
+      );
+    }
+    for (let i = 0; i < storageProvider.length; i++) {
+      const key = storageProvider.key(i);
+      if (!key.includes('https://kdsintf.amd.com/vcek/v1/Milan/')) {
```
glitch003, line 39: let's just make this check on the top level domain `https://kdsintf.amd.com/` because the URL path could change someday.  like for example Milan is hardcoded in there but we need to support Genoa, which would have a different URL path.

glitch003 (changes requested):
looks good - just the one change in the test so that when we start supporting Genoa, or if the AMD URL path changes, the test doesn't break.

e2e-nodejs/group-connection/test-connection-cert-caching.mjs
joshLong145: good point, has been updated.

Ansonhkg (approved):

joshLong145:
@glitch003 
Can you re review?

PR #441, opened 2024-04-24
Fix/siwe

joshLong145:
- Locks our `siwe-parser` version to `2.0.0` to assure api compatibility with `siwe`
- Changes our `*` import of `siwe` to a explicit import of `SiweMessage` as we no longer need to import the module as a named import.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "@playwright/test": "^1.25.2",
     "@simplewebauthn/browser": "^7.2.0",
     "@simplewebauthn/typescript-types": "^7.0.0",
+    "@spruceid/siwe-parser": "2.0.0",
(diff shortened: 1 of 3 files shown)


Ansonhkg (approved):
Just noting that we might need to lock several other packages so we won't get auto updates and break our sdk

joshLong145:
> Just noting that we might need to lock several other packages so we won't get auto updates and break our sdk

What other packages would we need to lock? I think this is the only one relating to `siwe` compatibility since I saw minor versions under `2.x.x` being used which broke compatibility with `siwe` so now that we are locking `siwe-parser` to `2.0.0` we shouldnt have api compatibility issues.

PR #442, opened 2024-04-25
auth method id for eth wallet is just the wallet address

glitch003:
The encoding for the EthWallet auth method type was wrong.  It should just be the raw address.  You can see this in the PKPPermissions facet here: https://github.com/LIT-Protocol/lit-assets/blob/8d26f49faf8d1648fb50deb9ff8d45a68ed49f80/blockchain/contracts/contracts/lit-node/PKPPermissions/PKPPermissionsFacet.sol#L506

Note that there is not ":lit" appended, no keccak256, etc.  it's just the address bytes.

For context, this is the function where the changed line is: 

```
public static async authMethodId(authMethod: AuthMethod): Promise<string> {
    let address: string;

    try {
      address = JSON.parse(authMethod.accessToken).address;
    } catch (err) {
      throw new Error(
        `Error when parsing auth method to generate auth method ID for Eth wallet: ${err}`
      );
    }

    return address;
  }
```

### DIFF
--- a/packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
+++ b/packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts
@@ -204,6 +204,6 @@ export default class EthWalletProvider extends BaseProvider {
       );
     }
 
-    return ethers.utils.keccak256(ethers.utils.toUtf8Bytes(`${address}:lit`));
+    return address;
   }
 }


joshLong145:
@glitch003 
We need to hold off merging this until the below ticket is ready on the nodes:
https://linear.app/litprotocol/issue/LIT-3035/fix-check-for-raw-address-when-verifying-the-eth-authmethod

Ansonhkg:
Is this still valid @glitch003 @joshLong145 @DashKash54 ?

Ansonhkg:
Closing for now.

glitch003:
i looked into this - it's done already on the node side https://github.com/LIT-Protocol/lit-assets/pull/1077

the node accepts either format - this is the function that checks if you're curious - https://github.com/LIT-Protocol/lit-assets/blob/74fcbfa0aa4425a94e264168a2290c8fe8826267/rust/lit-node/src/pkp/auth/mod.rs#L548

PR #443, opened 2024-04-25
fix(sessionKey): signed session key mismatch with the one being signe…

Ansonhkg:
# Description

Solution cherry-picked from https://github.com/LIT-Protocol/js-sdk/pull/439 due to having different base branch, and we want to merge this to the master asap. as a patch.

## Update 1

- Moved `getSessionKey` out as a helper function instead of a class so we can do better unit testing instead of instantiating the whole class. We still have `getSessionKey` in the `LitNodeClient` to comply the `LitClientSessionManager` interface

## Type of change

# How Has This Been Tested?

```
node 'node_modules/.bin/jest' './packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.spec.ts' -c './Projects/issue-443/packages/lit-node-client-nodejs/jest.config.ts' -t 'getOrCreateSessionKey'
```

```
getOrCreateSessionKey
    ✓ should not be fool (7 ms)
    ✓ should not be negative number (1 ms)
    ✓ should always use the session key that was initially set if it is not expired. (40 ms)
    ✓ should generate a new session key if the cached session key has expired. (46 ms)
```

# Checklist:

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.ts
+++ b/packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.ts
@@ -11,17 +11,22 @@ let sessionKeyCache: SessionKeyCache | undefined = undefined;
 
 /**
  * Retrieves the session key from local storage or cache, or generates a new session key if none exists.
- * @param expiration - The expiration date of the session key.
+ * @param expiration - The ISO string of expiration date of the session key.
  * @returns The session key pair.
  */
 export const getOrCreateSessionKey = (expiration: string): SessionKeyPair => {
   if (!expiration) throw new Error('Expiration date is required');
 
+  // Check if the input expiration is a negative number
+  if (parseInt(expiration) < 0) {
+    throw new Error('Expiration date cannot be negative.');
+  }
+
   const siweExpiration = new Date(expiration).getTime();
 
-  // if it's not an instance of Date
-  if (!((siweExpiration as any) instanceof Date)) {
-    throw new Error('Invalid expiration date');
+  // Check if siweExpiration is not a number, meaning the date conversion failed
+  if (isNaN(siweExpiration)) {
+    throw new Error(`Invalid expiration date.`);
   }
 
   const storageKey = LOCAL_STORAGE_KEYS.SESSION_KEY;
--- a/packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.ts
+++ b/packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.ts
@@ -15,7 +15,14 @@ let sessionKeyCache: SessionKeyCache | undefined = undefined;
  * @returns The session key pair.
  */
 export const getOrCreateSessionKey = (expiration: string): SessionKeyPair => {
+  if (!expiration) throw new Error('Expiration date is required');
+
   const siweExpiration = new Date(expiration).getTime();
+
+  // if it's not an instance of Date
+  if (!((siweExpiration as any) instanceof Date)) {
+    throw new Error('Invalid expiration date');
+  }
 
   const storageKey = LOCAL_STORAGE_KEYS.SESSION_KEY;
   const storedSessionKeyOrError = getStorageItem(storageKey);
--- a/packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.ts
+++ b/packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.ts
@@ -14,7 +14,7 @@ let sessionKeyCache: SessionKeyCache | undefined = undefined;
  * @param expiration - The expiration date of the session key.
  * @returns The session key pair.
  */
-export const getSessionKey = (expiration: string): SessionKeyPair => {
+export const getOrCreateSessionKey = (expiration: string): SessionKeyPair => {
   const siweExpiration = new Date(expiration).getTime();
 
   const storageKey = LOCAL_STORAGE_KEYS.SESSION_KEY;
--- a/packages/lit-node-client-nodejs/src/lib/helpers/get-session-key.ts
+++ b/packages/lit-node-client-nodejs/src/lib/helpers/get-session-key.ts
@@ -1,0 +1,64 @@
+import { EITHER_TYPE, LOCAL_STORAGE_KEYS } from '@lit-protocol/constants';
+import { generateSessionKeyPair } from '@lit-protocol/crypto';
+import { log } from '@lit-protocol/misc';
+import { getStorageItem } from '@lit-protocol/misc-browser';
+import { SessionKeyCache, SessionKeyPair } from '@lit-protocol/types';
+
+// Global cache variable
+let sessionKeyCache: SessionKeyCache | null = null;
+
+// const _expiration = new Date(Date.now() + 24 * 60 * 60 * 1000).toISOString();
+
+/**
+ * Retrieves the session key from local storage or cache, or generates a new session key if none exists.
+ * @param expiration - The expiration date of the session key.
+ * @returns The session key pair.
+ */
+export const getSessionKey = (expiration: string): SessionKeyPair => {
+  const siweExpiration = new Date(expiration).getTime();
+
+  const storageKey = LOCAL_STORAGE_KEYS.SESSION_KEY;
+  const storedSessionKeyOrError = getStorageItem(storageKey);
+
+  if (
+    storedSessionKeyOrError.type === EITHER_TYPE.ERROR ||
+    !storedSessionKeyOrError.result ||
+    storedSessionKeyOrError.result === ''
+  ) {
+    console.warn(
+      `Storage key "${storageKey}" is missing. Not a problem. Contiune...`
+    );
+
+    // Check if a valid session key exists in cache
+    if (sessionKeyCache && Date.now() < sessionKeyCache.timestamp) {
+      log(`[getSessionKey] Returning session key from cache.`);
+      return sessionKeyCache.value;
+    }
+    // Cache is missing or expired, generate a new session key
+    const newSessionKey = generateSessionKeyPair();
+
+    // Try to set to local storage
+    try {
+      localStorage.setItem(storageKey, JSON.stringify(newSessionKey));
+    } catch (e) {
+      console.warn(`Localstorage not available. Not a problem. Contiune...`);
+    }
+    // Store in cache
+    sessionKeyCache = {
+      value: newSessionKey,
+      timestamp: siweExpiration,
+    };
+
+    return newSessionKey;
+  } else {
+    console.log('G');
+    return JSON.parse(storedSessionKeyOrError.result as string);
+  }
+};
+
+/**
+ * Clears the session key cache.
+ */
+export const clearSessionKeyCache = (): void => {
+  sessionKeyCache = null;
+};
--- a/packages/lit-node-client-nodejs/src/lib/helpers/get-session-key.ts
+++ b/packages/lit-node-client-nodejs/src/lib/helpers/get-session-key.ts
@@ -5,7 +5,7 @@ import { getStorageItem } from '@lit-protocol/misc-browser';
 import { SessionKeyCache, SessionKeyPair } from '@lit-protocol/types';
 
 // Global cache variable
-let sessionKeyCache: SessionKeyCache | null = null;
+let sessionKeyCache: SessionKeyCache | undefined = undefined;
 
 // const _expiration = new Date(Date.now() + 24 * 60 * 60 * 1000).toISOString();
 
@@ -60,5 +60,5 @@ export const getSessionKey = (expiration: string): SessionKeyPair => {
  * Clears the session key cache.
  */
 export const clearSessionKeyCache = (): void => {
-  sessionKeyCache = null;
+  sessionKeyCache = undefined;
 };
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -373,7 +373,11 @@ export class LitNodeClientNodeJs
    * if not, generates one.
    * @return { SessionKeyPair } session key pair
    */
-  getSessionKey = (expiration: string): SessionKeyPair => {
+  getSessionKey = (expiration?: string): SessionKeyPair => {
+    if (!expiration) {
+      expiration = LitNodeClientNodeJs.getExpiration();
+    }
+
     return getOrCreateSessionKey(expiration);
   };
 
@@ -2699,11 +2703,10 @@ export class LitNodeClientNodeJs
   getSessionSigs = async (
     params: GetSessionSigsProps
   ): Promise<SessionSigsMap> => {
-    const _expiration = params.expiration || this.getExpiration(); // 24 hours from now
-
     // -- prepare
     // Try to get it from local storage, if not generates one~
-    const sessionKey = params.sessionKey ?? this.getSessionKey(_expiration);
+    const sessionKey =
+      params.sessionKey ?? this.getSessionKey(params.expiration);
 
     const sessionKeyUri = this.getSessionKeyUri(sessionKey.publicKey);
 
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -106,6 +106,7 @@ import type {
   ValidateAndSignECDSA,
   WebAuthnAuthenticationVerificationParams,
   ILitNodeClient,
+  SessionKeyCache,
 } from '@lit-protocol/types';
 
 // TODO: move this to auth-helper for next patch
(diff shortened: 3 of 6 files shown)


packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -125,6 +126,9 @@ interface CapacityCreditsRes {
   capacityDelegationAuthSig: AuthSig;
 }
 
+// Global cache variable
+let sessionKeyCache: SessionKeyCache | null = null;
```
joshLong145, line 130: can we make this `undefined` instead of null?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -384,6 +390,15 @@ export class LitNodeClientNodeJs
         `Storage key "${storageKey}" is missing. Not a problem. Contiune...`
       );
 
+      // Check if a valid session key exists in cache
+      if (
+        sessionKeyCache &&
+        Date.now() - sessionKeyCache.timestamp < expirationInMs
```
joshLong145, line 396: can we use the timestamp from the `siwe` here instead of creating a new timestamp for the expiration?

packages/lit-node-client-nodejs/src/lib/helpers/get-session-key.ts
```diff
@@ -0,0 +1,64 @@
...
+let sessionKeyCache: SessionKeyCache | undefined = undefined;
+
+// const _expiration = new Date(Date.now() + 24 * 60 * 60 * 1000).toISOString();
+
+/**
+ * Retrieves the session key from local storage or cache, or generates a new session key if none exists.
+ * @param expiration - The expiration date of the session key.
+ * @returns The session key pair.
+ */
+export const getSessionKey = (expiration: string): SessionKeyPair => {
```
joshLong145, line 17: we should rename this as to not get confused with `LitNodeClient.getSessionSigs` can we maybe add something that outlines its retrieving from cache or creating? Maybe `getOrCreateSessionKey`

joshLong145 (commented):
Looks good happy we now are explicitly typing `expiration`

joshLong145 (approved):

joshLong145:
@Ansonhkg can you see why `e2e` tests are failing?

MaximusHaximus (changes requested):
I think that having a very low-level assumption that when we call getSessionKey() multiple times, in different functions, that it will always return the same thing is a bad smell; why should we expect it to always be the same when we call it multiple times in different functions?  If we require that function B is called with the same session key that function A fetched, can't we just pass function B the correct session key rather than having it call getSessionKey() also, which is not guaranteed to return the same thing even in the browser?
Imagine cases where devs are using the SDK to manage multiple session keys, and/or running concurrent code that happens to (currently) persist a new sessionKey while they are using an existing one.  I think this case is just highlighting the brittle nature of our assumptions around this code.
Instead of trying to finagle the environment our code is running in to make multiple getSessionKey() calls return the same value (which they may or may not in an environment where multiple async operations are being done concurrently), could we pass the session key we first fetched to the code that needs it to be the same?  That way we are actually, truly 100% guaranteed that it is the same, and we haven't coupled our nodejs functionality to a single session key that must be persisted to a polyfilled storage medium.

joshLong145:
> I think that having a very low-level assumption that when we call getSessionKey() multiple times, in different functions, that it will always return the same thing is a bad smell; why should we expect it to always be the same when we call it multiple times in different functions? If we require that function B is called with the same session key that function A fetched, can't we just pass function B the correct session key rather than having it call getSessionKey() also, which is not guaranteed to return the same thing even in the browser? Imagine cases where devs are using the SDK to manage multiple session keys, and/or running concurrent code that happens to (currently) persist a new sessionKey while they are using an existing one. I think this case is just highlighting the brittle nature of our assumptions around this code. Instead of trying to finagle the environment our code is running in to make multiple getSessionKey() calls return the same value (which they may or may not in an environment where multiple async operations are being done concurrently), could we pass the session key we first fetched to the code that needs it to be the same? That way we are actually, truly 100% guaranteed that it is the same, and we haven't coupled our nodejs functionality to a single session key that must be persisted to a polyfilled storage medium.

i see what you mean 

> I think that having a very low-level assumption that when we call getSessionKey() multiple times, in different functions, that it will always return the same thing is a bad smell; why should we expect it to always be the same when we call it multiple times in different functions? If we require that function B is called with the same session key that function A fetched, can't we just pass function B the correct session key rather than having it call getSessionKey() also, which is not guaranteed to return the same thing even in the browser? Imagine cases where devs are using the SDK to manage multiple session keys, and/or running concurrent code that happens to (currently) persist a new sessionKey while they are using an existing one. I think this case is just highlighting the brittle nature of our assumptions around this code. Instead of trying to finagle the environment our code is running in to make multiple getSessionKey() calls return the same value (which they may or may not in an environment where multiple async operations are being done concurrently), could we pass the session key we first fetched to the code that needs it to be the same? That way we are actually, truly 100% guaranteed that it is the same, and we haven't coupled our nodejs functionality to a single session key that must be persisted to a polyfilled storage medium.

This makes sense but it will be difficult to implement what you are suggesting in a patch release. The implementation in this PR is currently compatible with the `5.0.0` API. I think your point has many merits but for the sake of fixing this in a minor version bump I think we should try this pr and make a task to allow for passing a `session key` to session generation.

packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    // Store in cache
+    sessionKeyCache = {
+      value: newSessionKey,
+      timestamp: siweExpiration,
+    };
+
+    return newSessionKey;
+  } else {
+    console.log('G');
```
joshLong145, line 61: Oh we should remove this log?

joshLong145 (changes requested):

Ansonhkg:
> @Ansonhkg can you see why `e2e` tests are failing? Looks like the expiration change might have caused some regression in our intigration tests pasting the error below:
> 
> ```js
> Error: Expiration date is required
>     at getOrCreateSessionKey (/home/runner/work/js-sdk/js-sdk/dist/packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.js:18:15)
>     at LitNodeClientNodeJs.getSessionKey (/home/runner/work/js-sdk/js-sdk/dist/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.js:166:74)
>     at Object.sessionSigsAndAuthContext [as fn] (file:///home/runner/work/js-sdk/js-sdk/e2e-nodejs/group-pkp-session-sigs/test-pkp-multiple-auth-contexts.mjs:74:33)
>     at testThis (file:///home/runner/work/js-sdk/js-sdk/tools/scripts/utils.mjs:638:42)
>     at testThese (file:///home/runner/work/js-sdk/js-sdk/tools/scripts/utils.mjs:679:11)
>     at async file:///home/runner/work/js-sdk/js-sdk/e2e-nodejs/group-pkp-session-sigs/test-pkp-multiple-auth-contexts.mjs:139:1
> Error: Invalid expiration date
>     at getOrCreateSessionKey (/home/runner/work/js-sdk/js-sdk/dist/packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.js:22:15)
>     at LitNodeClientNodeJs.getSessionKey (/home/runner/work/js-sdk/js-sdk/dist/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.js:166:74)
>     at LitNodeClientNodeJs.getSessionSigs (/home/runner/work/js-sdk/js-sdk/dist/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.js:1687:58)
>     at PKPEthersWallet.runSign (/home/runner/work/js-sdk/js-sdk/dist/packages/pkp-base/src/lib/pkp-base.js:280:72)
>     at PKPEthersWallet.signMessage (/home/runner/work/js-sdk/js-sdk/dist/packages/pkp-ethers/src/lib/pkp-ethers.js:153:36)
>     at Object.main [as fn] (file:///home/runner/work/js-sdk/js-sdk/e2e-nodejs/group-pkp-session-sigs/test-pkp-sign-with-session-sigs-eth-wallet-auth-method.mjs:60:37)
>     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
>     at async testThis (file:///home/runner/work/js-sdk/js-sdk/tools/scripts/utils.mjs:638:31)
>     at async file:///home/runner/work/js-sdk/js-sdk/e2e-nodejs/group-pkp-session-sigs/test-pkp-sign-with-session-sigs-eth-wallet-auth-method.mjs:81:1
> Error: Invalid expiration date
>     at getOrCreateSessionKey (/home/runner/work/js-sdk/js-sdk/dist/packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.js:22:15)
>     at LitNodeClientNodeJs.getSessionKey (/home/runner/work/js-sdk/js-sdk/dist/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.js:166:74)
>     at LitNodeClientNodeJs.getSessionSigs (/home/runner/work/js-sdk/js-sdk/dist/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.js:1687:58)
>     at PKPEthersWallet.runSign (/home/runner/work/js-sdk/js-sdk/dist/packages/pkp-base/src/lib/pkp-base.js:280:72)
>     at PKPEthersWallet.signMessage (/home/runner/work/js-sdk/js-sdk/dist/packages/pkp-ethers/src/lib/pkp-ethers.js:153:36)
>     at Object.main [as fn] (file:///home/runner/work/js-sdk/js-sdk/e2e-nodejs/group-pkp-session-sigs/test-pkp-sign-with-session-sigs-session-cache.mjs:79:37)
>     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
>     at async testThis (file:///home/runner/work/js-sdk/js-sdk/tools/scripts/utils.mjs:638:31)
>     at async file:///home/runner/work/js-sdk/js-sdk/e2e-nodejs/group-pkp-session-sigs/test-pkp-sign-with-session-sigs-session-cache.mjs:92:1
> ```

That's because we've made the `expiration` param to be required for the `getSessionKey` function but we are not passing in any in the tests. I've changed `expiration` to optional, and create an expiration if it's not provided. Also changed that we shouldn't have checked the `Date` type cus we are expecting a number from the `getTime()` function. 

E2E and Unit tests are working as expected now.

```

 ✓    It should validate that only one of sessionSigs and authContext is defined.


 PASS   lit-node-client-nodejs  packages/lit-node-client-nodejs/src/lib/helpers/get-or-create-session-key.spec.ts
  getOrCreateSessionKey
    ✓ should not be fool (5 ms)
    ✓ should not be negative number (1 ms)
    ✓ should always use the session key that was initially set if it is not expired. (43 ms)
    ✓ should generate a new session key if the cached session key has expired. (37 ms)
```

MaximusHaximus:
> This makes sense but it will be difficult to implement what you are suggesting in a patch release. The implementation in this PR is currently compatible with the `5.0.0` API. I think your point has many merits but for the sake of fixing this in a minor version bump I think we should try this pr and make a task to allow for passing a `session key` to session generation.

I'm not sure I understand why we couldn't fix this in a patch; I think it's actually a bug fix that requires no new behaviours and simply fixes our usage of our own APIs in our existing code.

Let's review the original issue as described by Chris:
`if you're using authProvider.getSessionSigs in nodejs it doesn't work.  because when it first tries to sign the session key, in the authNeededCallback. it generates a new one.  and then when it actually tries to sign it, it tries to pull from localstorage, which fails, and then we generate a new session key.  and then the nodes complain about session key mismatch...`

So the existing code that is causing a problem is in `BaseProvider.ts` [(link here)](https://github.com/LIT-Protocol/js-sdk/blob/dccdaf8dfe7bc04dace25e30eb100b2cc9b650ca/packages/lit-auth-client/src/lib/providers/BaseProvider.ts#L131)

Note that both `litNodeClient.signSessionKey()` [(link)](https://github.com/LIT-Protocol/js-sdk/blob/8efcb0f7bd6b7206d2daeae0fff4e0c657fd276f/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts#L2456) and `litNodeClient.getSessionSigs()` [(link)](https://github.com/LIT-Protocol/js-sdk/blob/8efcb0f7bd6b7206d2daeae0fff4e0c657fd276f/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts#L2729) take an _optional argument_ of `sessionKey`. If that argument is provided, no new key will be generated and whether or not localstorage is defined is irrelevant because the passed argument is used as the `sessionKey` instead of internal auto-magic.

Note that the `BaseProvider.getSessionSigs()` method _also_  already takes an optional argument of `params.sessionSigsParams.sessionKey`

If someone calls `BaseProvider.getSessionSigs()` and they provide `sessionSigsParams.sessionKey`, neither `litNodeClient.signSessionKey()` nor `litNodeClient.getSessionSigs()` will make any internal caching decisions -- per the above, both will _always use that value instead of trying to create one or load one from localstorage_. The same value will be passed into `signSessionKey()` in our `authNeededCallback` definition _and_ to `litNodeClient.getSessionSigs()` on L203, and the code will work as expected.

So the problem case is when the caller of `BaseProvidfer.getSessionSigs()` does not provide the optional `params.sessionSigsParams.sessionKey` argument to `BaseProvider.getSessionSigs()` -- then the assumption is that internal code will _either_ generate a new one, or load it from localstorage _if it exists_.

If localstorage exists, whether it is stored there already or a new one is generated doesn't matter - the code always works.
However, if localstorage does not exist, it should generate a new one, and the code should still work.  The bug is that this case doesn't currently work, because the code in our pre-defined `authNeededCallback` always passes the (not defined in this case) `params.sessionSigsParams.sessionKey` value to `litNodeClient.signSessionKey()` -- and since it is not defined, `litNodeClient.signSessionKey()` will incorrectly generate a new one that is different than the one that was just generated inside of `litNodeClient.getSessionSigs()`.

But remember -- _both methods will use whatever `sessionKey` that they are passed instead of running internal logic to generate or load it from localstorage when it is provided_.

So imagine that we update the code in BaseProvider.ts so that on L143 we check to see if `sessionSigsParams.sessionKey` is undefined, and _if it is_, we call `litNodeClient.getSessionKey()` right then, inside of `BaseProvider.getSessionSigs()`.

That code will generate a new session key or load it from localstorage, just like it does at the very top of `litNodeClient.getSessionSigs()`, but because we haven't delegated it being called to `litNodeClient.getSessionSigs()`, we now have the (one and only one, correct) sessionKey in scope for both our call to `litNodeClient.getSessionKeys()` _and_ to `litNodeClient.signSessionKey()` inside of our auth callback. We can just _pass the sessionKey that was returned by `litNodeClient.getSessionKey()` into both the `getSessionSigs()` and `signSessionKey()` method calls_.  Now it works in nodeJs without localstorage, we haven't changed _any API at_, and we haven't added caching that makes this logic even harder to reason about where the source of truth is.  IMO, our existing API already provides for consistent and deterministic usage, it's just that we're not _using it correctly_.

joshLong145:
@Ansonhkg @MaximusHaximus 
So after conversations around the current api I think we reached agreement on adding passthrough to the `signSessionKey` call here: https://github.com/LIT-Protocol/js-sdk/blob/068a0f4dd9a6705105c1bab81c20931cfd597ed1/packages/lit-auth-client/src/lib/providers/BaseProvider.ts#L166

So Should we change this pr to simply passing in the provided `sessionKey` or just generating in scope of `BaseProvider.getSessionSigs` if the user does not provide one?

MaximusHaximus:
> @Ansonhkg @MaximusHaximus So after conversations around the current api I think we reached agreement on adding passthrough to the `signSessionKey` call here:
> 
> https://github.com/LIT-Protocol/js-sdk/blob/068a0f4dd9a6705105c1bab81c20931cfd597ed1/packages/lit-auth-client/src/lib/providers/BaseProvider.ts#L166
> 
> So Should we change this pr to simply passing in the provided `sessionKey` or just generating in scope of `BaseProvider.getSessionSigs` if the user does not provide one?

I think we were going to update any example apps or tests to generate it _outside_ of the `BaseProvider` method and pass it as an argument, but also add logic _inside_ the `BaseProvider` method that will generate it in that scope if it is not provided, and pass it into both of the `LitNodeClient` calls that are called inside of the `BaseProvider` method :+1:

joshLong145:
> > @Ansonhkg @MaximusHaximus So after conversations around the current api I think we reached agreement on adding passthrough to the `signSessionKey` call here:
> > https://github.com/LIT-Protocol/js-sdk/blob/068a0f4dd9a6705105c1bab81c20931cfd597ed1/packages/lit-auth-client/src/lib/providers/BaseProvider.ts#L166
> > 
> > So Should we change this pr to simply passing in the provided `sessionKey` or just generating in scope of `BaseProvider.getSessionSigs` if the user does not provide one?
> 
> I think we were going to update any example apps or tests to generate it _outside_ of the `BaseProvider` method and pass it as an argument, but also add logic _inside_ the `BaseProvider` method that will generate it in that scope if it is not provided, and pass it into both of the `LitNodeClient` calls that are called inside of the `BaseProvider` method 👍

Yes, thinking we can probably just close this pr and #443 and go with the above mentioned implementation.

PR #444, opened 2024-04-25
Feature/lit 2961 auth unification breaking remove existing authsigs

Ansonhkg:
# Description

- Removed `authSig` as an option to pass in to endpoint functions eg. `pkpSign` `executeJs` and `signSessionKey`
- Extracted out all parser, helper, transcompiler functions in the `pkpSign` and `executeJs` to a dedicated [helpers](https://github.com/LIT-Protocol/js-sdk/tree/feature/lit-2961-auth-unification-breaking-remove-existing-authsigs/packages/lit-node-client-nodejs/src/lib/helpers) folder with isolated unit tests
- Removing `getXXPromiseShares` methods in the `LitNodeClient` & its interface, and replacing it with a generic `generatePromise` instead. See lit-node-client-nodejs.ts
- Better types for IOs

# Note

The commit history is a bit overlapping with other PRs because I need the tests and some other features, please review from the commit below onward (or I just open a new PR and cherry pick the commits if that's better, lmk.)
- [35ede36](https://github.com/LIT-Protocol/js-sdk/pull/444/commits/35ede368ee826b7323dcdb1ff87f332393cdf935) 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

- [x] Unit test (Jest)
- [x] E2E test (Tinny)

# Checklist:

- [x] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #446, opened 2024-04-29
Feature/lit 3041 js sdk cherry pick fixes for session key mismatch 443#issuecomment 2082614246

Ansonhkg:
# Description

New Implementation suggested here:
https://github.com/LIT-Protocol/js-sdk/pull/443#issuecomment-2079735472

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/lit-auth-client/src/lib/providers/BaseProvider.ts
+++ b/packages/lit-auth-client/src/lib/providers/BaseProvider.ts
@@ -178,29 +178,20 @@ export abstract class BaseProvider {
           }),
         };
 
-        // prepare auth-specific data based on the authentication method
-        let authSpecificData = {};
-
         if (params.authMethod.authMethodType === AuthMethodType.EthWallet) {
-          authSpecificData = {
-            authSig: JSON.parse(params.authMethod.accessToken),
+          const authSig = JSON.parse(params.authMethod.accessToken);
+
+          response = await nodeClient.signSessionKey({
+            ...commonData,
+            authSig: authSig,
             authMethods: [],
-          };
+          });
         } else {
-          authSpecificData = {
+          response = await nodeClient.signSessionKey({
+            ...commonData,
             authMethods: [params.authMethod],
-          };
+          });
         }
-
-        // Merge the common and auth-specific data
-        response = await nodeClient.signSessionKey({
-          // default
-          authMethods: [],
-
-          // override
-          ...commonData,
-          ...authSpecificData,
-        });
 
         return response.authSig;
       };
--- a/packages/lit-auth-client/src/lib/providers/BaseProvider.ts
+++ b/packages/lit-auth-client/src/lib/providers/BaseProvider.ts
@@ -200,6 +200,7 @@ export abstract class BaseProvider {
     // Generate session sigs with the given session params
     const sessionSigs = await this.litNodeClient.getSessionSigs({
       ...params.sessionSigsParams,
+      sessionKey,
       authNeededCallback,
     });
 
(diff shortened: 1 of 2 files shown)


packages/types/src/lib/interfaces.ts
```diff
@@ -1424,7 +1424,7 @@ export interface BaseProviderSessionSigsParams {
   /**
    * Lit Node Client to use. If not provided, will use an existing Lit Node Client or create a new one
    */
-  litNodeClient?: any;
+  litNodeClient?: ILitNodeClient;
```
MaximusHaximus, line 1427: 💖 slowly but surely :D

packages/lit-auth-client/src/lib/providers/BaseProvider.ts
```diff
@@ -161,40 +165,43 @@ export abstract class BaseProvider {
 
         let response: SignSessionKeyResponse;
 
+        // common data for the signSessionKey function call
+        const commonData = {
```
MaximusHaximus, line 169: 💖 Nice, is a lot easier to follow the differences between EthWallet and other branches now

packages/lit-auth-client/src/lib/providers/BaseProvider.ts
```diff
@@ -161,40 +165,43 @@ export abstract class BaseProvider {
...
         }
 
+        // Merge the common and auth-specific data
+        response = await nodeClient.signSessionKey({
+          // default
+          authMethods: [],
+
+          // override
+          ...commonData,
+          ...authSpecificData,
```
MaximusHaximus, line 202: While combining the common properties into a single `commonData` is awesome and reduces cognitive load here, I think `authSpecificData` as a dedicated variable does the opposite because it is entirely different depending on the code that has run before this line.  

It would be easier to follow if we kept the existing pattern of composing the arguments to send to `signSessionKey()` as separate blocks and didn't define an intermediary `authSpecificData` that is a different shape of object depending on the code above it... this way there is an extra layer of indirection that we need to follow to figure out what `authSpecificData` will be and what we're passing into this method.

After in-lining this, we can also get rid of the `// default` entry with an empty `authMethods` array; it's actually duplicative given that we define it as an empty array on L187 anyway :)

MaximusHaximus (changes requested):
Added a couple of things in-line :)

Also, IIRC, in order to fix the original issue we need to pass `sessionKey` into the params passed into `getSessionSigs()` on L210 also; the bug was that both `getSessionSigs()` and `signSessionKey()` would generate new keys when localstorage didn't exist and they would be different.

MaximusHaximus (approved):
:+1: I think this is good to go 💖

packages/lit-auth-client/src/lib/providers/BaseProvider.ts
DashKash54: Seconded

packages/lit-auth-client/src/lib/providers/BaseProvider.ts
```diff
@@ -161,37 +165,31 @@ export abstract class BaseProvider {
...
+
         if (params.authMethod.authMethodType === AuthMethodType.EthWallet) {
           const authSig = JSON.parse(params.authMethod.accessToken);
+
           response = await nodeClient.signSessionKey({
-            statement: authCallbackParams.statement,
-            sessionKey: params.sessionSigsParams.sessionKey,
-            authMethods: [],
+            ...commonData,
             authSig: authSig,
```
DashKash54, line 186: We no longer pass AuthSig. I guess it's removed in another PR?

DashKash54 (approved):
LGTM!

packages/lit-auth-client/src/lib/providers/BaseProvider.ts
joshLong145: this will be removed in #444

PR #447, opened 2024-04-29
Feat/add authenticate method pkp base

joshLong145:
# Description
Adds an `authenticate` method which will use the `authContext` if provided to reassign the `controllerSessionSigs` property on `PKPBase` which allows for `authentication` operations to be explicitly invoked. This is helpful when we require to first check the authentication status before allowing other operations to occur. Without this addition it would not be possible to get the `controllerSessionSigs` before calling a higher level function on `PKPEthers` for example.

## Type of change

# How Has This Been Tested?

An Additional check has been added to [test-pkp-sign-with-session-sigs-session-cache.mjs](https://github.com/LIT-Protocol/js-sdk/compare/feat/add-authenticate-method-pkp-base?expand=1#diff-bce9a9e7d7f9d561dc46d4831e321663be35d1c1ea15be2249d7544a731c9bde)
to assert that `session` has been defined before calling `signMessage`

# Checklist:

### DIFF
--- a/packages/pkp-base/src/lib/pkp-base.ts
+++ b/packages/pkp-base/src/lib/pkp-base.ts
@@ -255,6 +255,18 @@ export class PKPBase<T = PKPBaseDefaultParams> {
     }
   }
 
+  public async authenticate(): Promise<void> {
+    this.validateAuthContext();
+
+    const controllerSessionSigs =
+      (await this.authContext?.client?.getSessionSigs(
+        this.authContext.getSessionSigsProps
+      )) || this.controllerSessionSigs;
+
+    this.log('got session: ', controllerSessionSigs);
+    this.controllerSessionSigs = controllerSessionSigs;
+  }
+
   /**
    * Runs the specified Lit action with the given parameters.
    *
@@ -275,17 +287,12 @@ export class PKPBase<T = PKPBaseDefaultParams> {
       this.throwError('pkpPubKey (aka. uncompressPubKey) is required');
     }
 
-    this.validateAuthContext();
-
-    const controllerSessionSigs =
-      (await this.authContext?.client?.getSessionSigs(
-        this.authContext.getSessionSigsProps
-      )) || this.controllerSessionSigs;
+    await this.authenticate();
 
     const executeJsArgs: ExecuteJsProps = {
       ...(this.litActionCode && { code: this.litActionCode }),
       ...(this.litActionIPFS && { ipfsId: this.litActionIPFS }),
-      sessionSigs: controllerSessionSigs,
+      sessionSigs: this.controllerSessionSigs,
       authSig: this.controllerAuthSig,
       authMethods: this.authContext?.authMethods,
       jsParams: {
@@ -358,12 +365,7 @@ export class PKPBase<T = PKPBaseDefaultParams> {
       this.throwError('pkpPubKey (aka. uncompressPubKey) is required');
     }
 
-    this.validateAuthContext();
-
-    const controllerSessionSigs =
-      (await this.authContext?.client?.getSessionSigs(
-        this.authContext.getSessionSigsProps
-      )) || this.controllerSessionSigs;
+    await this.authenticate();
 
     try {
       let sig;
@@ -374,12 +376,12 @@ export class PKPBase<T = PKPBaseDefaultParams> {
           authSig: this.controllerAuthSig as AuthSig,
           authMethods: [],
         });
-      } else if (controllerSessionSigs) {
+      } else if (this.controllerSessionSigs) {
         sig = await this.litNodeClient.pkpSign({
           toSign,
           pubKey: this.uncompressedPubKey,
           authMethods: this.authContext?.authMethods ?? [],
-          sessionSigs: controllerSessionSigs,
+          sessionSigs: this.controllerSessionSigs,
         });
       } else if (this.authContext?.authMethods) {
         sig = await this.litNodeClient.pkpSign({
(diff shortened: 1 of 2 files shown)


packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -255,6 +254,18 @@ export class PKPBase<T = PKPBaseDefaultParams> {
...
   }
 
+  public async authenticate(): Promise<void> {
+    this.validateAuthContext();
+
+    const controllerSessionSigs =
+      (await this.authContext?.client?.getSessionSigs(
+        this.authContext.getSessionSigsProps
+      )) || this.controllerSessionSigs;
+
```
Ansonhkg, line 264: can we have an ultimate check to see if `controllerSessionSigs` exists?

```ts
if(!controllerSessionSigs){
  throw new Error('bruh.. no session sigs');
}
```

packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -255,6 +254,18 @@ export class PKPBase<T = PKPBaseDefaultParams> {
     }
   }
 
+  public async authenticate(): Promise<void> {
+    this.validateAuthContext();
+
+    const controllerSessionSigs =
+      (await this.authContext?.client?.getSessionSigs(
+        this.authContext.getSessionSigsProps
+      )) || this.controllerSessionSigs;
```
MaximusHaximus, line 263: This strikes me as a bit odd -- why would we keep the existing value for `this.controllerSessionSigs` here rather than throwing an error if our `authenticate` logic returns a falsy value, or set `this.controllerSessionSigs` to the falsy value that was returned?

packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -255,6 +254,18 @@ export class PKPBase<T = PKPBaseDefaultParams> {
     }
   }
 
+  public async authenticate(): Promise<void> {
```
MaximusHaximus, line 257: Would be good to have JSDoc on this if it's part of our public API, especially if we're going to leave the existing `controllerSessionSigs` intact if our call to `getSessionSigs()` returns a falsy value, explaining why that is

packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -358,12 +364,7 @@ export class PKPBase<T = PKPBaseDefaultParams> {
       this.throwError('pkpPubKey (aka. uncompressPubKey) is required');
     }
 
-    this.validateAuthContext();
```
MaximusHaximus, line 361: Can we really get rid of this call and assume everything is correct? It seems like we would still want to call it, _after_ calling `authenticate`?  And if we're going to do that, wouldn't we want to leave the logic _inside_ `validateAuthContext()` intact rather than removing the check for `this.controllerSessionSigs`?

MaximusHaximus (changes requested):
Questions in-line

packages/pkp-base/src/lib/pkp-base.ts
Ansonhkg, line 257: copy/paste this if you need:
```
  /**
   * Authenticates the user - generate sessionSigs from litNodeClient(authContext.client) is provided. Otherwise, use
   * controllerSessionSigs.
   *
   * @returns A Promise that resolves to void.
   * @throws An error if `controllerSessionSigs` is not found.
   */
```

packages/pkp-base/src/lib/pkp-base.ts
joshLong145: I was also on the fence about this but wanted a fallback if a previous authenticate call was successful. But probably should just let it throw if not defined after `getSessionSigs` is called.

packages/pkp-base/src/lib/pkp-base.ts
joshLong145: ah yes this makes much more sense. Ill update to check after the variable assignment. good catch!

packages/pkp-base/src/lib/pkp-base.ts
joshLong145: Updated the pr with a new validation method `validateAuthenticate() ` which will throw if `controllerSessionSigs` is not defined after `authenticate` is called.

packages/pkp-base/src/lib/pkp-base.ts
joshLong145: Update: I have added a new `validateAuthenticate()` method to check on the internal properties of the object after `getSessionSigs` is called on the `client` which will perform the needed checks similar to `validateAuthContext` but is specific to the checks needed after we generate a new session instance.

packages/pkp-base/src/lib/pkp-base.ts
joshLong145: added doc header

packages/pkp-base/src/lib/pkp-base.ts
```diff
@@ -255,6 +255,75 @@ export class PKPBase<T = PKPBaseDefaultParams> {
...
+  private validateAuthenticate() {
+    const providedAuthentications = [
+      this.controllerAuthSig,
+      this.authContext,
+      this.controllerSessionSigs,
+    ].filter(Boolean).length;
+
+    if (providedAuthentications !== 2) {
+      // log which authentications has the user provided
+      if (this.controllerAuthSig) {
```
Ansonhkg, line 267: Note for myself (@Ansonhkg): remove this for 0.1 release https://linear.app/litprotocol/issue/LIT-3032/[auth-unification]-remove-all-v0-stuff-before-v1-release

Ansonhkg (approved):

MaximusHaximus (changes requested):
Added some questions in-line -- I'm still not sure about what the golden path is for using this API and I think there are some issues w/ error handling.  Happy to huddle if you'd like to work it out sync :)

joshLong145:
Closing as after some investigation with @MaximusHaximus It will be lowering the performance of the overall `PKP-*` implementations. Will move forward with the Alchemy upgrade with the current implementations using `controllerSessionSigs`

PR #448, opened 2024-04-29
fix: Remove invalid test case and adds an exception if no `arguments` are supplied to `LitNodeClient`

joshLong145:
Adds an undefined check for using the `in` syntax for property check against `args` instance.

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -134,9 +134,9 @@ export class LitNodeClientNodeJs
 
   // ========== Constructor ==========
   constructor(args: LitNodeClientConfig | CustomNetwork) {
-    if(!args) {
+    if (!args) {
       throwError({
-        message: "must probvide LitNodeClient parameters",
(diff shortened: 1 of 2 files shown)


Ansonhkg (dismissed):

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -133,6 +134,14 @@ export class LitNodeClientNodeJs
 
   // ========== Constructor ==========
   constructor(args: LitNodeClientConfig | CustomNetwork) {
+    if (!args) {
+      throwError({
+        message: 'must probvide LitNodeClient parameters',
```
MaximusHaximus, line 139:
```suggestion
        message: 'must provide LitNodeClient parameters',
```

MaximusHaximus (approved):
LGTM but small typo :)

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
joshLong145: DOH

joshLong145:
Merging as was previously approved.

Issue #416, opened 2024-04-10
[Team Name] <title-for-the-issue>


gilbert:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SDK version

3.0.32, but issue exists on master

### Lit Network

N/A, trying to bundle package for browser

### Description of the bug/issue

```
Can't find variable: global
../../node_modules/@lit-protocol/auth-browser/src/lib/chains/eth.js
```

which points to [this line](https://github.com/LIT-Protocol/js-sdk/blob/7ec1caa8f2a69a851d3346463d72f1702549d0e6/packages/auth-browser/src/lib/chains/eth.ts#L45-L46)

### Severity of the bug

Can't bundle for browser in a create-react-app

### Steps To Reproduce

Using `globalThis` instead of `global` fixes the problem. The rest of the file also uses globalThis, so this would maintain consistency.

joshLong145:
thanks for calling this out! Nice find.

gilbert:
FYI the issue isn't fixed yet. The error was that `global` is not defined. The change in the PR...

```diff
- if (typeof global.Buffer === 'undefined') {
+ if (global && typeof global.Buffer === 'undefined') {
```

...still attempts to use the `global` variable, which will still produce the error.

The solution is to replace all usage of `global` with `globalThis`. The [MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#description) describes why you only need `globalThis`:

> Unlike similar properties such as `window` and `self` [and `global`], `globalThis` is guaranteed to work in window and non-window contexts. In this way, you can access the global object in a consistent manner without having to know which environment the code is being run in.

PR #452, opened 2024-05-03
Finish fix for #416

gilbert:
See here for details: https://github.com/LIT-Protocol/js-sdk/issues/416#issuecomment-2050838008. In a nutshell, you only need `globalThis`.

### DIFF
--- a/packages/auth-browser/src/lib/chains/eth.ts
+++ b/packages/auth-browser/src/lib/chains/eth.ts
@@ -42,10 +42,6 @@ import {
 } from '@lit-protocol/misc';
 import { getStorageItem } from '@lit-protocol/misc-browser';
 
-if (global && typeof global.Buffer === 'undefined') {
-  global.Buffer = BufferPolyfill;
-}
-
 if (globalThis && typeof globalThis.Buffer === 'undefined') {
   globalThis.Buffer = BufferPolyfill;
 }


Ansonhkg (commented):
Hey, thanks for the PR. I'm not sure if this will completely fix the issue, because we are now using `cross-fetch` [PR](https://github.com/LIT-Protocol/js-sdk/pull/427), which uses global under the hood. Therefore, the same problem might occur. 

image

May I suggest to configure your build tool to shim the global object while we investigate this issue? 

eg. in vite

```
import { defineConfig } from "vite";
import react from "@vitejs/plugin-react-swc";


export default defineConfig({
  plugins: [react()],
  define: {
    global: "window",
  },
});
```

cc @MaximusHaximus @joshLong145  can you take a look?

gilbert:
Indeed, I'm upgrading to lit v6 and am now running into the cross-fetch issue, but that seems to a problem with their library and not lit's.

Either way, there's no reason to use `global` over `globalThis`, and I see a4dd28ffc4386d64e2fbc6de19ab08e0ca7bc82d implemented this fix. Thanks!

Ansonhkg:
> cross-fetch

Thanks, we are considering ripping out `cross-fetch` completely, as it's too hard to satisfy all environments, we will just let users to polyfill themselves if required.

PR #454, opened 2024-05-07
Feature/lit 3146 discord cannot read properties of undefined reading topics

joshLong145:
# Description
Connects https://github.com/LIT-Protocol/Issues-and-Reports/issues/32
Where in the event of no topic data in our transaction receipt our `mintWithAuth` implementation would throw the exception
`Cannot read properties of undefined (reading topics).`
This pr adds checks to the event topic we require for the `tokenId` and will throw an exception with the `transactionHash` included to let users better understand why an exception is occurring. If `topics` are not present in the event data we cannot return the `pkpInfo` as we do not have a `tokenId` to perform look ups with.


Ansonhkg (approved):

joshLong145:
closing as was merged into #455

joshLong145:
Closing as was merged into #455

PR #456, opened 2024-05-08
Feature/lit 3173 cayenne network bootstrap

joshLong145:
Refactors the special bootstrapping logic added for `cayenne` which will now use the same code paths as `Habanero` and `Manzano` for its url bootstrapping logic. A new check has been on the port range to see if the node's port is between
`8470 - 8479` as these have been determined to be `https` ports.


packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -915,7 +915,11 @@ export class LitContracts {
 
     const networks = activeValidatorStructs.map((item: any) => {
       let proto = 'https://';
-      if (item.port !== 443) {
+      /**
+       * ports in range of 8470 - 8479 are configured for https on cayenne network
+       * we shouold resepct https on these ports as they are using OpenSSL
+       */
+      if (item.port !== 443 || (item.port < 8480 && item.port > 8469)) {
```
hwrdtm, line 922: Think the ge and le operators are the other way around, and need to use || instead. Or you can just ! The whole expression

joshLong145: refactored the above to
```js
if (item.port !== 443 && item.port > 8480 && item.port < 8469) {
```

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -915,7 +915,11 @@ export class LitContracts {
 
     const networks = activeValidatorStructs.map((item: any) => {
       let proto = 'https://';
-      if (item.port !== 443) {
+      /**
+       * ports in range of 8470 - 8479 are configured for https on cayenne network
+       * we shouold resepct https on these ports as they are using OpenSSL
```
hwrdtm, line 920:
```suggestion
       * we shouold resepct https on these ports as they are using trusted ZeroSSL certs
```

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -915,7 +915,11 @@ export class LitContracts {
 
     const networks = activeValidatorStructs.map((item: any) => {
       let proto = 'https://';
-      if (item.port !== 443) {
+      /**
+       * ports in range of 8470 - 8479 are configured for https on cayenne network
```
hwrdtm, line 919:
```suggestion
       * ports in range of 8470 - 8479 are configured for https on custom networks (eg. cayenne)
```

hwrdtm (approved):

Ansonhkg:
Resolved comment & all e2e tests against cayenne passed. lgtm!

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -915,7 +915,11 @@ export class LitContracts {
...
+       * we shouold resepct https on these ports as they are using OpenSSL
+       */
+      if (item.port !== 443 && item.port > 8480 && item.port < 8469) {
```
hwrdtm, line 922: technically should be `> 8479` and `< 8470` but no big deal

PR #460, opened 2024-05-09
Feature/lit 3166 add response aggregation options for executejs

joshLong145:
# Description

Adds in a new concept of `LitAction` strategy which allows for one of the three options
- most common lit action response
- least common lit action response
- custom lit action response filtering

## Type of change

# How Has This Been Tested?

Additions of new unit tests for testing the new utility.

# Checklist:


packages/lit-node-client-nodejs/src/lib/helpers/process-lit-action-response-strategy.ts
```diff
@@ -0,0 +1,70 @@
...
+  NodeShare,
+} from '@lit-protocol/types';
+import { log, logError } from '@lit-protocol/misc';
+
+/**
+ * Finds the most and least common object within an of objects array
+ * @param responses any[]
+ * @returns an object which contains both the least and most occuring item in the array
+ */
+const findFrequency = (responses: string[]): { min: any; max: any } => {
```
Ansonhkg, line 13: can we export this and test this specifically? 

```ts
import { findFrequency } from './process-lit-action-response-strategy';

describe('findFrequency', () => {
  it('should return the least and most common items in the array', () => {
    const responses: string[] = ['a', 'b', 'a', 'c', 'b', 'a'];
    const result = findFrequency(responses);
    expect(result.min).toBe('c');
    expect(result.max).toBe('a');
  });
});
```

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -1201,6 +1201,12 @@ export class LitNodeClientNodeJs
       responseData
     ) as NodeShare;
 
+    const responseFromStrategy: any = processLitActionResponseStrategy(
+      responseData,
+      params.responseStrategy ?? { strategy: 'leastCommon' }
```
MaximusHaximus, line 1206: I don't see `responseStrategy` as a valid option for `JsonExecutionSdkParams`, yet we check for it in the new code.  If the idea is that `executeJs()` needs to support `responseStrategy`, shouldn't the `JsonExecutionSdkParams` be updated also?

MaximusHaximus (commented):
Seems like this changed our default behaviour in `executeJs` to be to find the least common result instead of the most common, which it did previously.

Added some comments and questions in-line! :)

I am curious about `leastCommon` -- under what circumstances do we expect to need to know the least common response?  If this exists to facilitate running on a single node, can we infer whether we should use `leastCommon` or `mostCommon` based on `params.targetNodeRange` instead of adding a new supported public argument to the `executeJs` method? Right now the new code appears to _always_ be using `leastCommon`.

Likewise, I am curious about `custom` -- do we actually have a purpose for that currently? If we don't have a specific need for it and it's only run by our tests, I would tend to lean towards us not implementing it until it becomes necessary.

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
MaximusHaximus: Missed it in the new def :+1:

PR #469, opened 2024-05-15
feat: remove e2e-nodejs and replace it with Tinny, and use Tinny for CI

Ansonhkg:
# Description

Remove old `e2e-nodejs` e2e tests and replacing it with Tinny, including CI.


joshLong145 (approved):
yay! thanks this is nice to see.

PR #472, opened 2024-05-20
Feature/lit 3269 js sdknode fix custom auth issue

Ansonhkg:
# Description

This PR addresses a few issues:
1. Unable to mint pkp with custom auth using `mintWithAuth` because the function is throwing an error when getting `authId` from Lit supported auth methods.
2. Not passing the whole set of params back to authNeededCallback when resign is required
3. `jsParams` type not accepting any other keys

## New feature:
- Add `getLitActionSessionSigs()` with explicit interface properties (either `litActionCode` or `litActionIpfsId`, and required `jsParams`).

## Related PRs:
- https://github.com/LIT-Protocol/lit-assets/pull/1157
- https://github.com/LIT-Protocol/lit-assets/pull/1158
- developer guide: https://github.com/LIT-Protocol/developer-guides-code/pull/13

## Type of change

# How Has This Been Tested?

## Test Command:

```
DEBUG=true NETWORK=localchain yarn test:local --filter=testUseCustomAuthSessionSigsToPkpSign
```

```
✅ res: {
  r: '4605d1450c67301575e0306b89b1ce6a68548b720e785820cc7ca6f835c858ab',
  s: '2e8558d89f69a2b37296a510721698c8989a0b5c1df58eb74a0b960f9683860e',
  recid: 0,
  signature: '0x4605d1450c67301575e0306b89b1ce6a68548b720e785820cc7ca6f835c858ab2e8558d89f69a2b37296a510721698c8989a0b5c1df58eb74a0b960f9683860e1b',
  publicKey: '04F7CD6AA4977DB70C87D615F9E9F6660A30FB979AD7E4858AED7D9EBECF8A39D66C23916286990A121A406FF534DD1A5087B3AFD6E8BD35A31EB858387C0C2389',
  dataSigned: '7D87C5EA75F7378BB701E404C50639161AF3EFF66293E9F375B5F17EB50476F4'
}
```
# Checklist:

### DIFF
--- a/local-tests/tests/testUseCustomAuthSessionSigsToPkpSign.ts
+++ b/local-tests/tests/testUseCustomAuthSessionSigsToPkpSign.ts
@@ -8,9 +8,9 @@ import { TinnyEnvironment } from 'local-tests/setup/tinny-environment';
 
 /**
  * Test Commands:
- * ❌ NETWORK=cayenne yarn test:local --filter=testUseCustomAuthSessionSigsToPkpSign
- * ❌ NOT AVAILABLE IN HABANERO
- * ❌ NETWORK=localchain yarn test:local --filter=testUseCustomAuthSessionSigsToPkpSign
+ * NETWORK=cayenne yarn test:local --filter=testUseCustomAuthSessionSigsToPkpSign
+ * NOT AVAILABLE IN HABANERO
+ * NETWORK=localchain yarn test:local --filter=testUseCustomAuthSessionSigsToPkpSign
  */
 export const testUseCustomAuthSessionSigsToPkpSign = async (
   devEnv: TinnyEnvironment
--- a/local-tests/tests/testUseCustomAuthSessionSigsToPkpSign.ts
+++ b/local-tests/tests/testUseCustomAuthSessionSigsToPkpSign.ts
@@ -1,0 +1,121 @@
+import {
+  LitAbility,
+  LitActionResource,
+  LitPKPResource,
+} from '@lit-protocol/auth-helpers';
+import { AuthMethodScope } from '@lit-protocol/constants';
+import { TinnyEnvironment } from 'local-tests/setup/tinny-environment';
+
+/**
+ * Test Commands:
+ * ❌ NETWORK=cayenne yarn test:local --filter=testUseCustomAuthSessionSigsToPkpSign
+ * ❌ NOT AVAILABLE IN HABANERO
+ * ❌ NETWORK=localchain yarn test:local --filter=testUseCustomAuthSessionSigsToPkpSign
+ */
+export const testUseCustomAuthSessionSigsToPkpSign = async (
+  devEnv: TinnyEnvironment
+) => {
+  const alice = await devEnv.createRandomPerson();
+
+  /**
+   * This is a custom auth method. It can be anything you want.
+   */
+  const customAuthMethod = {
+    authMethodType: 89989,
+    accessToken: 'xxx',
+  };
+
+  console.log('✅ customAuthMethod:', customAuthMethod);
+
+  const customAuthMethodOwnedPkp =
+    await alice.contractsClient.mintWithCustomAuth({
+      customAuthId: 'custom-app-user-id', // ipfs hash for the auth method
+      authMethod: customAuthMethod,
+      scopes: [AuthMethodScope.SignAnything],
+    });
+  console.log('✅ customAuthMethodOwnedPkp:', customAuthMethodOwnedPkp);
+
+  const addPermittedAuthMethodReceipt =
+    await alice.contractsClient.addPermittedAuthMethod({
+      pkpTokenId: alice.pkp.tokenId,
+      authId: 'app-id-xxx:user-id-yyy',
+      authMethodType: customAuthMethod.authMethodType,
+      authMethodScopes: [AuthMethodScope.SignAnything],
+    });
+
+  console.log(
+    '✅ addPermittedAuthMethodReceipt:',
+    addPermittedAuthMethodReceipt
+  );
+
+  /**
+   * Please note that the code below is first converted to a CID and stored in the smart contract.
+   * Therefore, the Lit action code executed in the `getPkpSessionSigs` function must match the CID stored in the smart contract.
+   *
+   * You can use https://explorer.litprotocol.com/create-action to create a Lit action and get the CID.
+   */
+  const litActionCodeString = `(async () => {
+    const a = 1;
+    const b = 2;
+
+    if (a + b === 3 && customAuthMethod.authMethodType === 89989) {
+      LitActions.setResponse({response:"true"});
+    } else {
+      LitActions.setResponse({response:"false"});
+    }
+
+    console.log("16 Lit.Auth:", Lit.Auth);
+  })()`;
+
+  const IPFSID = 'QmeG3spjJmqzFo4dpWte5Prv1xdUbPT3sYqo3n3ABQXgMF';
+
+  console.log('✅ IPFSID:', IPFSID.toString());
+
+  // Grant an action permission to use a PKP
+  const addPermittedActionReceipt =
+    await alice.contractsClient.addPermittedAction({
+      ipfsId: IPFSID,
+      pkpTokenId: alice.pkp.tokenId,
+      authMethodScopes: [AuthMethodScope.SignAnything],
+    });
+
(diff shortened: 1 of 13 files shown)


packages/contracts-sdk/src/lib/auth-utils.ts
```diff
@@ -338,3 +338,16 @@ function _resolveAuthFactor(factor: any): {
...
 }
+
+/**
+ * Converts a string into a byte array (arrayified value).
+ * @param str - The input string to be converted.
+ * @returns A Uint8Array representing the arrayified value of the string.
+ */
+export const stringToArrayify = (str: string) => {
+  try {
+    return ethers.utils.toUtf8Bytes(str);
```
joshLong145, line 349: could we use something other than ethers? or maybe add a TODO to replace when we start trying to decouple ethers from the system?

packages/types/src/lib/interfaces.ts
```diff
@@ -1891,3 +1901,34 @@ export interface SignatureData {
...
+  /**
+   * only applies to webauthn auth method
+   */
+  pubkey?: string;
+
+  /**
+   * The Auth ID of the given auth method. If it's custom auth, then it could be
+   * anything.
+   */
+  authId?: Uint8Array;
```
joshLong145, line 1926: why do we need this we have an `authMethod` already which should contain the id?

Ansonhkg: Not with custom auth though?

joshLong145: well they could define it however they want for example
```js
{
'autMethodType`: 81234,
'authMerthodId': 'abc123'
}
```
then we convert the `authMethodId` to an `uint8array`

Ansonhkg: hmm the default authMethod interface only has `authMethodType` and `accessToken` no? then the `authId` should be something like `app-id-xxx:user-id-yyy` for the custom auth.

DashKash54: Custom Auth Access token can be anything and it's not required that they store it on our chain

local-tests/tests/testUseCustomAuthSessionSigsToPkpSign.ts
```diff
@@ -0,0 +1,121 @@
...
+      customAuthId: 'custom-app-user-id', // ipfs hash for the auth method
+      authMethod: customAuthMethod,
+      scopes: [AuthMethodScope.SignAnything],
+    });
+  console.log('✅ customAuthMethodOwnedPkp:', customAuthMethodOwnedPkp);
+
+  const addPermittedAuthMethodReceipt =
+    await alice.contractsClient.addPermittedAuthMethod({
+      pkpTokenId: alice.pkp.tokenId,
+      authId: 'app-id-xxx:user-id-yyy',
```
DashKash54, line 41: So you're overriding the above custom auth permission? Note the Custom Auth method is already permitted since you're using `mintWithCustomAuth`

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -980,15 +981,22 @@ export class LitContracts {
     return data;
   }
 
+  /**
+   * Mints a new token with authentication.
+   *
+   * @param authMethod - The authentication method.
+   * @param scopes - The permission scopes.
+   * @param pubkey - The public key.
+   * @param authId - (optional) The authentication ID.
```
DashKash54, line 990: Why is authId separate from authMethod? Isn't it included in the authMethod object?

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1032,7 +1040,7 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
       return scope;
     });
 
-    const authId = await getAuthIdByAuthMethod(authMethod);
+    const _authId = authId ?? (await getAuthIdByAuthMethod(authMethod));
```
DashKash54, line 1043: This will fail authMethod is custom?

DashKash54, line 1043: What happens with an empty authMethod/authId?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2406,6 +2406,13 @@ const resourceAbilityRequests = [
           uri: sessionKeyUri,
           nonce,
           resourceAbilityRequests: params.resourceAbilityRequests,
+
+          // -- optional fields
+          ...(params.litActionCode && { litActionCode: params.litActionCode }),
+          ...(params.litActionIpfsId && {
+            litActionIpfsId: params.litActionIpfsId,
+          }),
+          ...(params.jsParams && { jsParams: params.jsParams }),
```
DashKash54, line 2415: This was the bug right?

packages/contracts-sdk/src/lib/contracts-sdk.ts
DashKash54, line 1: Nice to create abstraction functions for adding permitted actions and authMethods. I was gonna ping us to do that

DashKash54 (commented):
Looks good. Left a couple comments

packages/types/src/lib/interfaces.ts
joshLong145: I think the problem is the only place we have `authMethodId`  as a property on an `authMethod` is in our smart contract implementations for registering it to a pkp. Within the sdk we derive the id we put on chain from the `authMethodId` as an example see the [discord](https://github.com/LIT-Protocol/js-sdk/blob/6af6a7c1847f2881764f99ed07e6777e5bf30372/packages/lit-auth-client/src/lib/providers/DiscordProvider.ts#L98) auth method. 
So we drive the `authMethodId` from the access token if it's an auth method we support. so I don't know if we need this outside of the new `mintWithCustomAuthParams`

packages/types/src/lib/interfaces.ts
```diff
@@ -1891,3 +1901,34 @@ export interface SignatureData {
...
+}
+
+export interface mintWithCustomAuthParams extends MintWithAuthParams {
+  /**
+   * For a custom authentication method, the custom auth ID should uniquely identify the user for that project. For example, for Google, we use appId:userId, so you should follow a similar format for Telegram, Twitter, or any other custom auth method.
+   */
+  customAuthId: string | Uint8Array;
```
joshLong145, line 1933: Should we also support other authMethods when using custom auth? I don't think we can right now if we only support a single `id`

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1041,7 +1049,7 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
     const tx = await this.pkpHelperContract.write.mintNextAndAddAuthMethods(
       2, // key type
       [authMethod.authMethodType],
-      [authId],
+      [_authId],
```
joshLong145, line 1052: so do we not use the `accessToken` from the `authMethod`? looks like we just take the `authMethodType`

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: yup

packages/contracts-sdk/src/lib/contracts-sdk.ts
Ansonhkg: yea, that's why `mintWithCustomAuth` function is created

Ansonhkg: You can't have empty `authMethod` for this function, only `authId` is optional

packages/contracts-sdk/src/lib/contracts-sdk.ts
Ansonhkg: We might have different terms being used in the SDK and the node?
image

joshLong145 (commented):
Approved, merging into staging as we have tested with [6.0.0-test3](https://www.npmjs.com/package/@lit-protocol/lit-node-client-nodejs/v/6.0.0-beta.testing3)

PR #479, opened 2024-05-24
Fix/execute with retry wrapper exec

joshLong145:
# Description

Removes `executeWithRetry` as we no longer need to retry operations due to network stability fixes. We now will only retry operations which are critical to the standup of the sdk. The included node endpoints which are eligible for retry at this time are:
- `/handshake`

Status codes we will retry on:
- 408 
- 502 
- 503
- 504

Introduces a new package [fetch-retry](https://github.com/jonbern/fetch-retry) which now handles retry logic for single requests. As we no longer need to group requests and retry as single network operation.

## Type of change

# How Has This Been Tested?

Tested with local testing to force endpoint failure to observe retry.

# Checklist:


local-tests/setup/tinny-environment.ts
```diff
@@ -381,7 +381,7 @@ export class TinnyEnvironment {
         dAppOwnerWallet: wallet,
         capacityTokenId: capacityTokenId,
         // Sets a maximum limit of 200 times that the delegation can be used and prevents usage beyond it
```
Ansonhkg, line 383:
```suggestion
        // Sets a maximum limit of 2000 times that the delegation can be used and prevents usage beyond it
```

packages/misc/src/lib/misc.ts
```diff
@@ -13,29 +14,24 @@ import {
...
-  RejectedNodePromises,
-  RetryTolerance,
 } from '@lit-protocol/types';
 import { JsonRpcProvider } from '@ethersproject/providers';
 import { Contract } from '@ethersproject/contracts';
 import { LogLevel, LogManager } from '@lit-protocol/logger';
-import { version } from '@lit-protocol/constants';
 import Ajv, { JSONSchemaType } from 'ajv';
+//@ts-ignore no tyoe df
+import fetchRetry from 'fetch-retry';
```
Ansonhkg, line 29: Hmm I'm worried how browser (react, vite, parcel, etc.) would react to this dependency, I'll give it a spin next week

joshLong145: its bundle is 'umd' format. I think it's okay but I'll publish a testing build with it for us to test with more before committing to it.

local-tests/setup/tinny-environment.ts
joshLong145: Ah I have some left over changes I should remove these.

packages/misc/src/lib/fetchWithRetry.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export const getRetryDelayHandlerForInterval: RetryDelayHandlerFactory<
+  typeof globalThis.fetch
+> = (interval: number) => {
+  return function getRetryDelay(
+    attempt: number
+    // error: Error | null,
+    // response: Response | null
+  ) {
+    return Math.random() * (Math.pow(2, attempt) * interval);
```
joshLong145, line 24: Is this truly exponential? since we are multiplying by a random value couldn't we end up with a lower interval?

joshLong145 (commented):
Looks good. Left one comment. Have we done a browser test? once we do then I think we are good to merge it.

joshLong145:
Screenshot 2024-06-03 at 10 21 10 AM
After doing some testing in browser environments it looks like the `umd` bundle target is non compatible with our build pipeline. Going to make a task in the current sprint to upgrade `fetch-retry` to a modern cross target bundling implementation. Once complete this pr can be merged.

joshLong145:
Closing for #497

PR #481, opened 2024-05-28
Ref/replace sequencer with replica url

joshLong145:
# Urgent change for Chronicle Timeout issues

Urgent change to replace our `Chronicle` RPC URL with a `replica endpoint` rather than using the `sequencer url` for all operations. We will now only use the `sequencer` url for transactions while using the `replica` in all other cases.


Ansonhkg (approved):
Should we also add that to the constants package,  `LIT_CHAINS` > `chronicleTestnet` > `rpcUrls` array?

glitch003:
> Should we also add that to the constants package, `LIT_CHAINS` > `chronicleTestnet` > `rpcUrls` array?

yes please

PR #487, opened 2024-05-31
Feature/lit 3125 sdk for wrapping up walletx 2

Ansonhkg:
# What

We add a new SDK package: `wrapped-keys`. It's been published with the `beta` tag: https://www.npmjs.com/package/@lit-protocol/wrapped-keys/v/6.0.3-wrapped-keys.beta.2

## AWS

1. The keys are encrypted and stored in a DynamoDB instance managed by us.
2. Every wrapped key is associated with a PKP which is used only for authentication and not for signing since the wrapped keys are used for signing. 
3. The access to the database is guarded by Lambda which allows only the PKP sessionSig to query the wrapped key item associated with it.
4. We only provide a single sessionSig to the Lambda since we only need to compare the delegation/inner AuthSig for establishing the wrapped key ownership.

# SDK package

The idea is to provide very easy to use functions for:
1. `importPrivateKey`- Imports a private key as a string and stores in our DynamoDB
6. `generatePrivateKey`- For new users generate a random private key inside the Lit Action and return it to the SDK that internally stores it in DynamoDB
7. `exportPrivateKey`- Fetches the DynamoDB item and decrypts using `decryptString` and provides the decrypted private key to the user after stripping the pre-pended "lit_"
8. `signTransactionWithEncryptedKey`- Fetches the item from DynamoDB and calls the Lit Action to sign the user provided transaction object. Optionally, broadcasts the signed transaction and returns the tx hash or just returns the signed tx that the user can broadcast themselves
9. `signMessageWithEncryptedKey`- Fetches the item from DynamoDB and calls the Lit Action to sign the user provided message object

## Access Control

The private data is encrypted with the condition to allow only the specific PKP wallet address to decrypt it. We prepend "lit_" to the private key hex string for security reasons.

## Supported Chains

We currently support all the EVM chains supported by the Lit nodes in the `rpc.config.yaml` file and all 3 Solana chains.

## Lit Actions

The idea is that the only place where the bare (un-encrypted) private key exist is the Lit Action (that too on a **single node**). Hence the SDK only fetches the encrypted private key from DynamoDB and provide it to the Lit Action where it is internally decrypted and used to sign and finally return the signed tx/message.

Since the idea is to provide a default function that doesn't require the user to provide a Lit Action we have published all the Lit Actions and use their `ipfsCid` instead of passing the code as a string.

There are separate Lit Actions for Solana & EVM wrapped keys because they have different dependencies like Solana Lit Actions need the `solana/web3` to be bundled (`wrapped-keys/esbuild.config.js`).

Since Solana tx have several params we accept an unsigned serialized tx in the Lit Action param whereas for EVM we accept tx params and craft the transaction object inside the Lit Action.

**Note:** We don't wait for the transaction to be confirmed after broadcasting from inside the Lit Action since it could take very long and the Lit Action can timeout.

## Tests

There are unit test for utility functions and Tinny test to test several valid and invalid scenarios for all the 5 functionalities listed above.

# Special Attention

While reviewing please pay special attention to the assertion in the tests which should be strict and the error handling i.e. we're throwing error when either the AWS DB throws an error or when the SDK throws an error eg: `executeJs`. Finally we're doing a bunch of if conditional checks after receiving the response from the `executeJs`


packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,216 @@
...
+    ciphertext,
+    dataToEncryptHash,
+  };
+
+  try {
+    const response = await fetch(ENCRYPTED_PRIVATE_KEY_ENDPOINT, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        pkpsessionsig: JSON.stringify(firstSessionSig),
```
joshLong145, line 46: why are we sending this as a header?

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,216 @@
...
+      },
+      body: JSON.stringify(data),
+    });
+
+    if (!response.ok) {
```
joshLong145, line 51: we should check the `Content-Type` and before we assume `text` as the serializer. This might not actually matter with this particular api

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,216 @@
...
+        `Could not import the encrypted key due to the error: ${errorBody}`
+      );
+
+      throw new Error(errorBody);
+    }
+
+    const importedPrivateKey: ImportPrivateKeyResponse = await response.json();
+    return importedPrivateKey.pkpAddress;
+  } catch (error) {
+    const errorMessage = `There was a problem fetching from the database: ${error}`;
```
joshLong145, line 63: We might not want to expose the actual database error to the caller since they do not own the resource.

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,216 @@
...
+    const response = await fetch(ENCRYPTED_PRIVATE_KEY_ENDPOINT, {
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+        pkpsessionsig: JSON.stringify(firstSessionSig),
+      },
+    });
+
+    if (!response.ok) {
+      const errorBody = await response.text();
```
joshLong145, line 88: Same as above

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,216 @@
...
+      throw new Error(errorBody);
+    }
+
+    const exportedPrivateKeyData: ExportPrivateKeyResponse =
+      await response.json();
+
+    const decryptedPrivateKey = await decryptToString(
+      {
+        accessControlConditions: allowPkpAddressToDecrypt,
+        chain: CHAIN_ETHEREUM,
```
joshLong145, line 102: Should we expose this param to the end user?

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,216 @@
...
+
+      throw new Error(errorBody);
+    }
+
+    const exportedPrivateKeyData: ExportPrivateKeyResponse =
+      await response.json();
+
+    ({ pkpAddress, ciphertext, dataToEncryptHash } = exportedPrivateKeyData);
+  } catch (error) {
+    const errorMessage = `There was a problem fetching from the database: ${error}`;
```
joshLong145, line 153: Same as above

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,216 @@
...
+        accessControlConditions: getPkpAccessControlCondition(pkpAddress),
+      },
+    });
+  } catch (err) {
+    throw new Error(
+      `Lit Action threw an unexpected error: ${JSON.stringify(err)}`
+    );
+  }
+
+  console.log(`Lit Action result: ${JSON.stringify(result)}`);
```
joshLong145, line 179: Are we guaranteed this is always json?

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,216 @@
...
+      `Expected "response" in Lit Action result: ${JSON.stringify(result)}`
+    );
+  }
+
+  if (typeof response !== 'string') {
+    // As the return value is a hex string
+    throw new Error(
+      `Lit Action should return a string response: ${JSON.stringify(result)}`
+    );
+  }
```
joshLong145, line 200: call to stringily will fail if `response` is not type of string.

packages/wrapped-keys/src/lib/wrapped-keys.spec.ts
```diff
@@ -0,0 +1,110 @@
...
...: (1) 'Threshold': 'Execution' for 'lit-litaction://*'. (2) 'Threshold': 'Signing' for 'lit-pkp://*'. (3) 'Auth': 'Auth' for 'lit-resolvedauthcontext://*'.\\n\\nURI: lit:session:4fd3d6ae41190cdd33a07bc5feb4a51b0c882474e6b51eb37cf799d6668eb44b\\nVersion: 1\\nChain ID: 1\\nNonce: 0xa8b687976835989b8ac57e8e6cb17fa316cc9ef74ea6174a588f08b11571829c\\nIssued At: 2024-06-02T19:46:47Z\\nExpiration Time: 2024-06-03T19:47:14.907Z\\nResources:\\n- urn:recap:eyJhdHQiOnsibGl0LWxpdGFjdGlvbjovLyoiOnsiVGhyZXNob2xkL0V4ZWN1dGlvbiI6W3t9XX0sImxpdC1wa3A6Ly8qIjp7IlRocmVzaG9sZC9TaWduaW5nIjpbe31dfSwibGl0LXJlc29sdmVkYXV0aGNvbnRleHQ6Ly8qIjp7IkF1dGgvQXV0aCI6W3siYXV0aF9jb250ZXh0Ijp7ImFjdGlvbklwZnNJZHMiOltdLCJhdXRoTWV0aG9kQ29udGV4dHMiOlt7ImFwcElkIjoibGl0IiwiYXV0aE1ldGhvZFR5cGUiOjEsImV4cGlyYXRpb24iOjE3MTc0NDQwMjAsInVzZWRGb3JTaWduU2Vzc2lvbktleVJlcXVlc3QiOnRydWUsInVzZXJJZCI6IjB4MkY2ZjU4NzRhNGQyNTFlMzVDZDc4YjM1NzZDQTkwYkQyZjA1RmUwQiJ9XSwiYXV0aFNpZ0FkZHJlc3MiOm51bGwsImN1c3RvbUF1dGhSZXNvdXJjZSI6IiIsInJlc291cmNlcyI6W119fV19fSwicHJmIjpbXX0","address":"0xd1Af1AAC50aC837C873200D17b78664aFCde597C"}],"issuedAt":"2024-06-02T19:47:16.707Z","expiration":"2024-06-03T19:47:14.907Z","nodeAddress":"https://207.244.70.36:8474"}`,
```
joshLong145, line 60: Nice use of mocking for api compatibility Is it safe for us to commit thiese signature to source control?

packages/wrapped-keys/src/lib/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+  return firstSessionSig;
+}
+
+export function getPkpAddressFromSessionSig(pkpSessionSig: AuthSig): string {
+  const sessionSignedMessage: SessionKeySignedMessage = JSON.parse(
+    pkpSessionSig.signedMessage
+  );
+
+  const capabilities = sessionSignedMessage.capabilities;
+  if (capabilities.length !== 1) {
```
joshLong145, line 32: Why do we wish to restrict the capabilities to a single item?

packages/wrapped-keys/src/lib/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export function getFirstSessionSig(pkpSessionSigs: SessionSigsMap): AuthSig {
+  const keys = Object.keys(pkpSessionSigs);
+  if (keys.length === 0) {
+    throw new Error(
+      `Invalid pkpSessionSigs, length zero: ${JSON.stringify(pkpSessionSigs)}`
+    );
+  }
+
+  const firstSessionSig: AuthSig = pkpSessionSigs[keys[0]];
```
joshLong145, line 20: Do we check in the lambdas that the node which signed the first signature is still in the active signing set? I feel there might be issues with old sessions which are signed from nodes which have been kicked will get validated in the Lamda context.

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: Because it's for validation

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: It's not the DB error but Lambda error. @MaximusHaximus wdyt

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: Only for exporting the key not for signing

packages/wrapped-keys/src/lib/wrapped-keys.spec.ts
DashKash54: The sessionSigs can be used only once by the nodes

packages/wrapped-keys/src/lib/utils.ts
DashKash54: As long as your sessionSig is valid what's the issue?

packages/wrapped-keys/src/lib/wrapped-keys.ts
joshLong145: Can this be any key type? if so wouldn't we want to specify `Solana Chains` for example.

packages/wrapped-keys/src/lib/wrapped-keys.spec.ts
joshLong145: Yes but the `authSig` can be used for certain network operations are we sure there is no risk to making these signatures public? Just trying to make sure we are not leaking. Also will these compatibility tests fail if the session expires?

packages/wrapped-keys/src/lib/utils.ts
joshLong145: If the node is not in the active set. In theory we should not trust the signature from that node. But since the sdk combines T number of shares where T is the network threshold for each `auth sig` composing the `session signature` map it only matters if we want to make sure an active network participant's mapped signatrue is being sent to the lamda. If we want to enforce this we can check the current `connectedUrls` in the sdk and see if the url from the authSig chosen from the map is indeed active.

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: The chain info is used here just to check the PKP address which is an Eth address

packages/wrapped-keys/src/lib/utils.ts
DashKash54: The inner authSig is a collection from all the nodes so even if a node is kicked out the innerAuthSig is valid and we validate that in the lambda. We check that the inner AuthSig is signed by the BLS root key?

packages/wrapped-keys/src/lib/utils.ts
DashKash54: Since this is on Cayenne there has to be only 1 BLS innerAuthSig. But for Habanero & Manzano we need another one for payment delegation so the Lambda need to be updated for it

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: Someone can dump the DB, make sure on Lambda that not happen

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: This is the Lit Action result which is always a JSON

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    // TODO!: Update to use decryptToSingleNode()
+    const decryptedPrivateKey = await Lit.Actions.decryptAndCombine({
+        accessControlConditions,
+        ciphertext,
+        dataToEncryptHash,
+        chain: 'ethereum',
+        authSig: null,
+    });
```
DashKash54, line 37: Add try/catch block here so making the error readable

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: @MaximusHaximus I think our Lambda will always return a test response?

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: https://github.com/LIT-Protocol/js-sdk/pull/487#discussion_r1630408698

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: https://github.com/LIT-Protocol/js-sdk/pull/487#discussion_r1630215051

packages/wrapped-keys/src/lib/wrapped-keys.ts
DashKash54: Our response is always a string. This can be verified in the `litAction.ts` file

packages/wrapped-keys/src/lib/wrapped-keys.spec.ts
DashKash54: It's a unit test so it doesn't hit the nodes hence even if it is expired it passes the test

DashKash54: The inner authSig is already expired so it can't be used to spoof

packages/wrapped-keys/src/lib/constants.ts
```diff
@@ -0,0 +1,13 @@
+const CHAIN_ETHEREUM = 'ethereum';
+const LIT_PREFIX = 'lit_';
+// Update the endpoint to Wrapped Key project endpoint
+const ENCRYPTED_PRIVATE_KEY_ENDPOINT =
+  'https://8wugrwstu1.execute-api.us-east-2.amazonaws.com/encrypted';
```
joshLong145, line 5: Should we make this configurable? We should also add some docs on what the different aws envs are.

DashKash54: @MaximusHaximus

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,217 @@
...
+  ExportPrivateKeyParams,
+  ExportPrivateKeyResponse,
+  ImportPrivateKeyParams,
+  SignTransactionWithEncryptedKeyParams,
+  SignMessageWithEncryptedKeyParams,
+  GeneratePrivateKeyParams,
+  GeneratePrivateKeyResponse,
+} from './interfaces';
+
+export async function generatePrivateKey({
```
Ansonhkg, line 27: JsDoc is needed

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,217 @@
...
+    data,
+    firstSessionSig
+  );
+  return {
+    pkpAddress: importedPrivateKey.pkpAddress,
+    generatedPublicKey: publicKey,
+  };
+}
+
+export async function importPrivateKey({
```
Ansonhkg, line 71: JsDoc is needed

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,217 @@
...
+  };
+
+  const importedPrivateKey = await storePrivateKeyMetadataToDatabase(
+    data,
+    firstSessionSig
+  );
+  return importedPrivateKey.pkpAddress;
+}
+
+export async function exportPrivateKey({
```
Ansonhkg, line 102: JsDoc is needed

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,217 @@
...
+    litNodeClient
+  );
+
+  // It will be of the form lit_<privateKey>
+  return decryptedPrivateKey.startsWith(LIT_PREFIX)
+    ? decryptedPrivateKey.slice(LIT_PREFIX.length)
+    : decryptedPrivateKey;
+}
+
+export async function signTransactionWithEncryptedKey<T = LitTransaction>({
```
Ansonhkg, line 155: JsDoc is needed

packages/wrapped-keys/src/lib/wrapped-keys.ts
```diff
@@ -0,0 +1,217 @@
...
+  } catch (err: any) {
+    throw new Error(
+      `Lit Action threw an unexpected error: ${JSON.stringify(err)}`
+    );
+  }
+
+  return postLitActionValidation(result);
+}
+
+export async function signMessageWithEncryptedKey({
```
Ansonhkg, line 188: JsDoc is needed

packages/wrapped-keys/src/lib/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+import { CHAIN_ETHEREUM, ENCRYPTED_PRIVATE_KEY_ENDPOINT } from './constants';
+import { ethers } from 'ethers';
+import {
+  ExportPrivateKeyResponse,
+  ImportPrivateKeyResponse,
+  StoreToDatabaseParams,
+} from './interfaces';
+// import { log } from 'console';
+
+export function getFirstSessionSig(pkpSessionSigs: SessionSigsMap): AuthSig {
```
Ansonhkg, line 18: JsDoc is needed

packages/wrapped-keys/src/lib/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  const pkpAddress = delegationAuthSig.address;
+
+  log(`pkpAddress to permit decryption: ${pkpAddress}`);
+
+  return pkpAddress;
+}
+
+export function getPkpAccessControlCondition(
```
Ansonhkg, line 59: JsDoc is needed

packages/wrapped-keys/src/lib/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+      parameters: [':userAddress'],
+      returnValueTest: {
+        comparator: '=',
+        value: pkpAddress,
+      },
+    },
+  ];
+}
+
+export async function fetchPrivateKeyMedataFromDatabase(
```
Ansonhkg, line 83: JsDoc is needed

packages/wrapped-keys/src/lib/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+    return await response.json();
+  } catch (error) {
+    const errorMessage = `There was a problem fetching from the database: ${error}`;
+    console.error(errorMessage);
+
+    throw new Error(errorMessage);
+  }
+}
+
+export async function storePrivateKeyMetadataToDatabase(
```
Ansonhkg, line 115: JsDoc is needed

packages/wrapped-keys/src/lib/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+    return await response.json();
+  } catch (error) {
+    const errorMessage = `There was a problem fetching from the database: ${error}`;
+    console.error(errorMessage);
+
+    throw new Error(errorMessage);
+  }
+}
+
+export function postLitActionValidation(
```
Ansonhkg, line 147: JsDoc is needed

local-tests/setup/session-sigs/get-pkp-session-sigs.ts
```diff
@@ -8,7 +8,8 @@ import { TinnyPerson } from '../tinny-person';
 export const getPkpSessionSigs = async (
   devEnv: TinnyEnvironment,
   alice: TinnyPerson,
-  resourceAbilityRequests?: LitResourceAbilityRequest[]
+  resourceAbilityRequests?: LitResourceAbilityRequest[],
+  expiration?: string
```
DashKash54, line 12: Adding this since we need PKP sessionSig expiry for Lambdas to be less than 15 mins

packages/wrapped-keys/esbuild.config.js
DashKash54, line 1: TODO: Bundle Eth Lit Actions and publish to Ipfs

packages/wrapped-keys/src/index.ts
```diff
@@ -0,0 +1,9 @@
+export * from './lib/wrapped-keys';
+export * from './lib/litActions/ethereum';
+export * from './lib/interfaces';
+
+export const signTransactionWithSolanaEncryptedKeyLitActionIpfsCid =
+  'QmSi9GL2weCFEP1SMAUw5PDpZRr436Zt3tLUNrSECPA5dT';
+
+export const signMessageWithSolanaEncryptedKeyLitActionIpfsCid =
+  'QmYQC6cd4EMvyB4XPkfEEAwNXJupRZWU5JsTCUrjey4ovp';
```
DashKash54, line 9: TODO: Move to `constants.ts` in `LIT_ACTION_CID_REPOSITORY`

packages/wrapped-keys/src/lib/interfaces.ts
```diff
@@ -0,0 +1,74 @@
+import { ILitNodeClient, SessionSigsMap } from '@lit-protocol/types';
+
+export interface StoreToDatabaseParams {
+  ciphertext: string;
+  dataToEncryptHash: string;
+}
+
+export interface GeneratePrivateKeyParams {
+  pkpSessionSigs: SessionSigsMap;
+  litActionCode: string; // TODO!: Update to use ipfsCid only when the Lit Actions are published
```
DashKash54, line 10: TODO: Publish Lit Action to Ipfs and remove `litActionCode `

packages/wrapped-keys/src/lib/interfaces.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export interface ExportPrivateKeyResponse {
+  pkpAddress: string;
+  ciphertext: string;
+  dataToEncryptHash: string;
+}
+
+export interface SignMessageWithEncryptedKeyParams {
+  pkpSessionSigs: SessionSigsMap;
+  litActionCode?: string; // TODO!: Update to use ipfsCid only when the Lit Actions are published
```
DashKash54, line 42: TODO: Remove and keep only Ipfs

packages/wrapped-keys/src/lib/interfaces.ts
```diff
@@ -0,0 +1,74 @@
...
+  ipfsCid?: string;
+  messageToSign: string | Uint8Array;
+  litNodeClient: ILitNodeClient;
+}
+
+export interface SignTransactionWithEncryptedKeyParams<T> {
+  pkpSessionSigs: SessionSigsMap;
+  litActionCode?: string; // TODO!: Update to use ipfsCid only when the Lit Actions are published
```
DashKash54, line 50: TODO: Remove and keep only Ipfs

packages/wrapped-keys/src/lib/interfaces.ts
```diff
@@ -0,0 +1,74 @@
...
+  chain: string;
+}
+
+export interface EthereumLitTransaction extends BaseLitTransaction {
+  toAddress: string;
+  value: string;
+  chainId: number;
+  gasPrice?: string;
+  gasLimit?: number;
+  dataHex?: string;
```
DashKash54, line 67: All the required Eth transaction fields

packages/wrapped-keys/src/lib/interfaces.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export interface SolanaLitTransaction extends BaseLitTransaction {
+  serializedTransaction: string;
```
DashKash54, line 71: We're expecting the user to bundle the Solana tx as unlike Eth tx these have several configurable params

packages/wrapped-keys/src/lib/litActions/ethereum.ts
DashKash54, line 1: TODO: Publish Eth Lit Actions to Ipfs

packages/wrapped-keys/src/lib/litActions/ethereum.ts
```diff
@@ -0,0 +1,250 @@
...
+    response: resp,
+  });
+};
+
+const _signTransactionWithEthereumEncryptedKeyLitAction = async () => {
+  const LIT_PREFIX = 'lit_';
+
+  // TODO!: Remove ALL the console.log statements
+  console.log('unsignedTransaction');
+  console.log(unsignedTransaction);
```
DashKash54, line 36: TODO: Remove logs

packages/wrapped-keys/src/lib/litActions/ethereum.ts
```diff
@@ -0,0 +1,250 @@
...
+    });
+  } catch (err) {
+    const errorMessage =
+      'Error: When decrypting to a single node- ' + err.message;
+    Lit.Actions.setResponse({ response: errorMessage });
+    return;
+  }
+
+  console.log('decryptedPrivateKey');
+  console.log(decryptedPrivateKey);
```
DashKash54, line 83: TODO: Remove logs

packages/wrapped-keys/src/lib/litActions/ethereum.ts
```diff
@@ -0,0 +1,250 @@
...
+    return;
+  }
+
+  const privateKey = decryptedPrivateKey.startsWith(LIT_PREFIX)
+    ? decryptedPrivateKey.slice(LIT_PREFIX.length)
+    : decryptedPrivateKey;
+  const wallet = new ethers.Wallet(privateKey);
+
+  console.log('unsignedTransaction.chain', unsignedTransaction.chain);
+  console.log('pkpAddress', pkpAddress);
```
DashKash54, line 96: TODO: Remove logs

packages/wrapped-keys/src/lib/litActions/ethereum.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  let nonce;
+  try {
+    nonce = await Lit.Actions.getLatestNonce({
+      address: wallet.address,
+      chain: unsignedTransaction.chain,
+    });
+    console.log('nonce');
+    console.log(nonce);
```
DashKash54, line 105: TODO: Remove logs

packages/wrapped-keys/src/lib/litActions/ethereum.ts
```diff
@@ -0,0 +1,250 @@
...
+      tx.gasLimit = await provider.estimateGas(tx);
+    } catch (err) {
+      const errorMessage = 'Error: When estimating gas- ' + err.message;
+      Lit.Actions.setResponse({ response: errorMessage });
+      return;
+    }
+  }
+
+  console.log('tx');
+  console.log(tx);
```
DashKash54, line 162: TODO: Remove logs

packages/wrapped-keys/src/lib/litActions/ethereum.ts
```diff
@@ -0,0 +1,250 @@
...
+    Lit.Actions.setResponse({ response: errorMessage });
+  }
+};
+
+const _signMessageWithEthereumEncryptedKeyLitAction = async () => {
+  const LIT_PREFIX = 'lit_';
+
+  // TODO!: Remove ALL the console.log statements
+  console.log('messageToSign');
+  console.log(messageToSign);
```
DashKash54, line 185: TODO: Remove logs

packages/wrapped-keys/src/lib/litActions/ethereum.ts
```diff
@@ -0,0 +1,250 @@
...
+    });
+  } catch (err) {
+    const errorMessage =
+      'Error: When decrypting to a single node- ' + err.message;
+    Lit.Actions.setResponse({ response: errorMessage });
+    return;
+  }
+
+  console.log('decryptedPrivateKey');
+  console.log(decryptedPrivateKey);
```
DashKash54, line 204: TODO: Remove logs

packages/wrapped-keys/src/lib/litActions/ethereum.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  const privateKey = decryptedPrivateKey.startsWith(LIT_PREFIX)
+    ? decryptedPrivateKey.slice(LIT_PREFIX.length)
+    : decryptedPrivateKey;
+  const wallet = new ethers.Wallet(privateKey);
+
+  try {
+    const signature = await wallet.signMessage(messageToSign);
+    console.log('signature');
+    console.log(signature);
```
DashKash54, line 219: TODO: Remove logs

packages/wrapped-keys/src/lib/litActions/ethereum.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    const recoveredAddress = ethers.utils.verifyMessage(
+      messageToSign,
+      signature
+    );
+    console.log('recoveredAddress');
+    console.log(recoveredAddress);
```
DashKash54, line 226: TODO: Remove logs

DashKash54:
TODO: Add test for generating Solana privateKey Lit Action

DashKash54:
TODO: Use Uint8Array instead of strings wherever possible

packages/wrapped-keys/esbuild.config.js
```diff
@@ -0,0 +1,28 @@
...
+      './src/lib/litActions/ethereum/src/signTransactionWithEthereumEncryptedKey.js',
+      './src/lib/litActions/ethereum/src/signMessageWithEthereumEncryptedKey.js',
+      './src/lib/litActions/ethereum/src/generateEncryptedEthereumPrivateKey.js',
+    ],
+    bundle: true,
+    minify: true,
+    sourcemap: false,
+    outdir: './src/lib/litActions/ethereum/dist',
+    inject: ['./buffer.shim.js'],
+  });
```
spacesailor24, line 27: Not sure when it was added, but I don't think this needs to be here twice

DashKash54: We publishing Ethereum Lit Actions to IPFS just like Solana

packages/wrapped-keys/src/lib/constants.ts
DashKash54: He's working on a PR to do this

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -0,0 +1,186 @@
...
+        }
+    } catch (err) {
+        const errorMessage = 'Error: When signing transaction- ' + err.message;
+        Lit.Actions.setResponse({ response: errorMessage });
+    }
+})();
+`;
+
+export const signMessageWithEthereumEncryptedKeyLitAction = `
+const LIT_PREFIX = 'lit_';
```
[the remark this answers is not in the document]
DashKash54: There's only a single prefix

Ansonhkg:
Added a diagram to the description @DashKash54 @spacesailor24 @joshLong145

packages/wrapped-keys/esbuild.config.js
Ansonhkg: @spacesailor24 If you look at the files in the `entryPoints` they are actually different files. Solana & Ethereum

packages/wrapped-keys/src/lib/constants.ts
```diff
@@ -0,0 +1,31 @@
+const CHAIN_ETHEREUM = 'ethereum';
+const LIT_PREFIX = 'lit_';
+export const NETWORK_EVM = 'evm';
+export const NETWORK_SOLANA = 'solana';
+export type Network = typeof NETWORK_EVM | typeof NETWORK_SOLANA;
+
+// Update the endpoint to Wrapped Key project endpoint
```
Ansonhkg, line 7: can we at least add a link to the PR in the comment for now?

packages/wrapped-keys/src/lib/constants.ts
```diff
@@ -0,0 +1,31 @@
...
+const ENCRYPTED_PRIVATE_KEY_ENDPOINT =
+  'https://8wugrwstu1.execute-api.us-east-2.amazonaws.com/encrypted';
+
+const LIT_ACTION_CID_REPOSITORY = Object.freeze({
```
Ansonhkg, line 11: can we have some comments briefly what they do and what are the expected inputs (think a function params) and output (think return type)?

packages/wrapped-keys/src/lib/constants.ts
DashKash54: There's no PR yet

Ansonhkg: ref: https://github.com/LIT-Protocol/wrapped-keys-lambdas/pull/3

packages/wrapped-keys/src/lib/constants.ts
DashKash54: Added JsDocs to the Lit Action files

packages/wrapped-keys/src/lib/constants.ts
DashKash54: https://github.com/LIT-Protocol/js-sdk/pull/513

packages/wrapped-keys/src/lib/api/export-private-key.ts
```diff
@@ -0,0 +1,55 @@
...
+      dataToEncryptHash,
+      sessionSigs: pkpSessionSigs,
+    },
+    litNodeClient
+  );
+
+  // It will be of the form lit_<privateKey>
+  if (!decryptedPrivateKey.startsWith(LIT_PREFIX)) {
+    throw new Error(
+      `PKey was not encrypted with salt; all wrapped keys must be prefixed with '${LIT_PREFIX}'`
```
joshLong145, line 47: Did we conclude if we want to allow non salted keys to resolve? Since it is technically possible to upload one without the salt can we throw here?

packages/wrapped-keys/src/lib/lit-actions-client/constants.ts
```diff
@@ -0,0 +1,18 @@
...
+  signMessage: Object.freeze({
+    evm: 'QmTMGcyp77NeppGaqF2DmE1F8GXTSxQYzXCrbE7hNudUWx',
+    solana: 'QmUxnWS8VU9QwZRdyfwsEtvhJZcvcdrannjokgZoA1sesy',
+  }),
+  generateEncryptedKey: Object.freeze({
+    evm: 'QmaoPMSqcze3NW3KSA75ecWSkcmWT1J7kVr8LyJPCKRvHd',
+    solana: 'QmdRBXYLYvcNHrChmsZ2jFDY8dA99CcSdqHo3p1ES3UThL',
+  }),
+};
+
```
joshLong145, line 17: if we are freezing the all properties on `LIT_ACTION_CID_REPOSITORY` should we also freeze the top level object such that it cannot be modified at runtime? const will be dropped at runtime and allow for modifications on the root object so we might want to consider a top level freeze as well.

packages/wrapped-keys/src/lib/api/export-private-key.ts
DashKash54: They technically can do that but only in the Custom Wrapped Keys so we can only add it the docs informing them

MaximusHaximus (approved):
:+1: Let's do it!

DashKash54 (approved):
Approving to be merged

DashKash54 (approved):

glitch003 (approved):

spacesailor24 (approved):

PR #491, opened 2024-06-03
Fix/epoch param name

glitch003:
The param to send to the nodes for the epoch is "epoch" not "epochNumber".  i fixed this and confirmed with a manual test that this does indeed fix signing across epochs :-D


joshLong145:
Oh amazing find. will merge this asap. if I test locally can i confirm it working by attempting a signature op when the nodes are in an epoch transition? or is there more to it?

joshLong145 (approved):

PR #501, opened 2024-06-07
Fetch `gasPrice` and estimate gas in Wrapped Keys LA if not provided 

spacesailor24:
Adds to the Wrapped Keys Lit Action:

- Fetch `gasPrice` if not provided by user
- Estimate gas if `gasLimit` not provided by user

Also adds a new test `testEthereumBroadcastWrappedKeyWithDefaultGasParams`


simulified:
There is a problem
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 1000; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

local-tests/tests/testEthereumBroadcastWrappedKeyWithDefaultGasParams.ts
DashKash54, line 1: This test is passing right?

packages/wrapped-keys/src/lib/litActions.ts
DashKash54, line 1: Let's convert the string into a function which will give use syntax highlighting, code completion, etc. and just call the function's `toString()` to return us the code string

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -49,43 +43,95 @@ const DEFAULT_GAS_PRICE = '50'; // in gwei
...
+    const signedTx = await Lit.Actions.runOnce(
+        { waitForResponse: true, name: "signedTx" },
+        async () => {
+            try {
+                return wallet.signTransaction(tx);
+            } catch (error) {
+                const errorMessage = 'Error: When signing transaction- ' + err.message;
+                Lit.Actions.setResponse({ response: errorMessage });
+                return;
+            }
```
DashKash54, line 106: I'm not sure what happens when there's an error in `runOnce` maybe other nodes continue their execution since they aren't running this piece of code? But since it has the `waitForResponse: true` I think they'll time out.

I also think that if the node running this code throws an error then that won't be bubbled up to the user as a majority of the nodes will return "timeout error". We should test this: Make the `runOnce` throw an error and see what happens

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -49,43 +43,95 @@ const DEFAULT_GAS_PRICE = '50'; // in gwei
...
+        )
+    }
+
     console.log('tx');
     console.log(tx);
 
-    try {
-        const signedTx = await wallet.signTransaction(tx);
+    const signedTx = await Lit.Actions.runOnce(
```
DashKash54, line 97: Since only once node is signing the tx why not broadcast in the same `runOnce` block?

DashKash54 (commented):
Largely looks. Left a few comments especially about testing some unexpected behaviour

packages/wrapped-keys/src/lib/litActions.ts
DashKash54, line 1: Why are you `hexlify`ing value, gasPrice & gasLimit?

packages/wrapped-keys/src/lib/litActions.ts
DashKash54: This might be worth consideration...

local-tests/tests/testEthereumBroadcastWrappedKeyWithDefaultGasParams.ts
spacesailor24: It passes using
```
NETWORK=cayenne yarn test:local --filter=testEthereumBroadcastWrappedKeyWithDefaultGasParams
```
but fails with
```
NETWORK=manzano yarn test:local --filter=testEthereumBroadcastWrappedKeyWithDefaultGasParams
```

packages/wrapped-keys/src/lib/litActions.ts
spacesailor24: [Updated](https://github.com/LIT-Protocol/js-sdk/pull/501/commits/cdb404976af05bd588f6de587fafa5f077a335fe)

spacesailor24: Removed

spacesailor24: If I don't hexlify them then I get the error:
```
Lit Action result: {"success":true,"signedData":{},"decryptedData":{},"claimData":{},"response":"Error: When signing transaction- invalid hexlify value (argument=\"value\", value=\"1000000\", code=INVALID_ARGUMENT, version=bytes/5.7.0)","logs":"unsignedTransaction\n{\n  toAddress: \"0x9BA2BeC180A81Aa16F7EF7bC2B9208819Dce7A19\",\n  value: \"0.0001\",\n  chainId: 175177,\n  gasLimit: 30000,\n  dataHex: \"0x54657374207472616e73616374696f6e2066726f6d20416c69636520746f20626f62\",\n  chain: \"chronicleTestnet\"\n}\ndecryptedPrivateKey\n\n"}
response
Error: When signing transaction- invalid hexlify value (argument="value", value="1000000", code=INVALID_ARGUMENT, version=bytes/5.7.0)
✖ 1. testEthereumBroadcastWrappedKeyWithFetchGasParams - Failed after 1 attempts (27428.44 ms)
❌Error: Error: Error executing the Signing Lit Action: Error: When signing transaction- invalid hexlify value (argument="value", value="1000000", code=INVALID_ARGUMENT, version=bytes/5.7.0)
Test Report: Some tests failed.
- testEthereumBroadcastWrappedKeyWithFetchGasParams (Failed in 27428.44 ms) - Error: Error: Error executing the Signing Lit Action: Error: When signing transaction- invalid hexlify value (argument="value", value="1000000", code=INVALID_ARGUMENT, version=bytes/5.7.0)
```
Although I'm not sure why it says `value` is causing the error. This error goes away when hexlify-ing `gasPrice` or `gasLimit`

packages/wrapped-keys/src/lib/litActions.ts
spacesailor24: I've removed this code when merging in your latest changes

packages/wrapped-keys/src/lib/litActions.ts
spacesailor24: I did:
```ts
if (unsignedTransaction.gasLimit) {
    tx.gasLimit = unsignedTransaction.gasLimit;
  } else {
    tx.gasLimit = await Lit.Actions.runOnce(
      { waitForResponse: true, name: 'gasLimit' },
      async () => {
        const rpcUrl = await Lit.Actions.getRpcUrl({
          chain: unsignedTransaction.chain,
        });
        const provider = new ethers.providers.JsonRpcProvider(rpcUrl);

        try {
          const gasLimit = await provider.estimateGas(tx);
          throw new Error('This is a fake error');
        } catch (err) {
          const errorMessage = 'Error: When estimating gas- ' + err.message;
          Lit.Actions.setResponse({ response: errorMessage });
          return;
        }
      }
    );
  }
```

and it seemed to have just skipped past the error and continue execution

local-tests/tests/testEthereumBroadcastWrappedKeyWithDefaultGasParams.ts
spacesailor24: Also worth noting that fetching the gasLimit on Chronicle doesn't work, but I have tested it on Sepolia and it works as expected

DashKash54: Fetching gasLimit doesn't work on Chronicle as you have to provide the from field. What's happening is we're saying to the blockchain nodes "simulate sending this money" but then not telling the node "where" the money comes from, so an undefined behavior. Maybe on other chain they handle this case

DashKash54: Yeah I'll fix the Manzano bug, have to update the Lambda since it could contain a capacityDelegation AuthSig hence 2 inner AuthSigs

local-tests/tests/testEthereumBroadcastWrappedKeyWithFetchGasParams.ts
```diff
@@ -3,19 +3,19 @@ import { ethers } from 'ethers';
...
 } from '@lit-protocol/wrapped-keys';
 import { getPkpSessionSigs } from 'local-tests/setup/session-sigs/get-pkp-session-sigs';
 
 /**
  * Test Commands:
- * ✅ NETWORK=cayenne yarn test:local --filter=testEthereumBroadcastWrappedKeyWithDefaultGasParams
- * ✅ NETWORK=manzano yarn test:local --filter=testEthereumBroadcastWrappedKeyWithDefaultGasParams
- * ✅ NETWORK=localchain yarn test:local --filter=testEthereumBroadcastWrappedKeyWithDefaultGasParams
+ * ✅ NETWORK=cayenne yarn test:local --filter=testEthereumBroadcastWrappedKeyWithFetchGasParams
+ * ❌ NETWORK=manzano yarn test:local --filter=testEthereumBroadcastWrappedKeyWithFetchGasParams Fails with Error: Error: There should be exactly 1 element in the capabilities array but there are: 2
```
DashKash54, line 15: I'll fix this an issue in the AWS Lambda

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,146 +1,193 @@
...
-    if (!unsignedTransaction.chain) {
-        Lit.Actions.setResponse({ response: 'Error: Missing required field: chain' });
-        return;
-    }
-
-    if (!unsignedTransaction.value) {
-        Lit.Actions.setResponse({ response: 'Error: Missing required field: value' });
-        return;
-    }
+export const generatePrivateKeyLitAction = `
```
DashKash54, line 3: Let's convert this to a function from a string too

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,146 +1,193 @@
...
-`;
+  } catch (err) {
+    const errorMessage = 'Error: When signing transaction- ' + err.message;
+    Lit.Actions.setResponse({ response: errorMessage });
+  }
+}).toString();
+
+export const signTransactionWithEthereumEncryptedKeyLitAction = `(${_signTransactionWithEthereumEncryptedKeyLitAction})();`;
 
 export const signMessageWithEthereumEncryptedKeyLitAction = `
```
DashKash54, line 192: Let's convert this to a function from a string too

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,146 +1,193 @@
...
+`;
 
-    console.log('decryptedPrivateKey');
-    console.log(decryptedPrivateKey);
-
-    if (!decryptedPrivateKey) { // Exit the nodes which don't have the decryptedData
-        return;
-    }
+const _signTransactionWithEthereumEncryptedKeyLitAction = (async () => {
+  const LIT_PREFIX = 'lit_';
```
DashKash54, line 46: It's being used at multiple places lets pull this out from this function and reuse in the below functions?

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,146 +1,193 @@
...
+    ),
+    chainId: unsignedTransaction.chainId,
+    data: unsignedTransaction.dataHex,
+    nonce,
+  };
+
+  if (unsignedTransaction.gasPrice) {
+    tx.gasPrice = ethers.utils.parseUnits(unsignedTransaction.gasPrice, 'gwei');
+  } else {
+    tx.gasPrice = await Lit.Actions.runOnce(
```
DashKash54, line 123: This is not need as we already exit if the `decryptedPrivateKey` is undefined since we ran the function `decryptToSingleNode` above. So this point all this is being run only on a single node

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,146 +1,193 @@
...
+      { waitForResponse: true, name: 'gasPrice' },
+      async () => {
+        const rpcUrl = await Lit.Actions.getRpcUrl({
+          chain: unsignedTransaction.chain,
+        });
+        const provider = new ethers.providers.JsonRpcProvider(rpcUrl);
```
DashKash54, line 129: You can just do this once as use the provider for gasLimit and broadcasting

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,146 +1,193 @@
...
+          const gasLimit = await provider.estimateGas(tx);
+          return ethers.utils.hexlify(gasLimit);
+        } catch (err) {
+          const errorMessage = 'Error: When estimating gas- ' + err.message;
+          Lit.Actions.setResponse({ response: errorMessage });
+          return;
+        }
+      }
+    );
+  }
```
DashKash54, line 164: Same as above no need to use `runOnce` as at this point it's only one node executing it

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,146 +1,193 @@
...
-            }
-        );
-
-        const { txHash, rpcUrl } = JSON.parse(resp);
-        console.log(rpcUrl);
-        Lit.Actions.setResponse({ response: txHash });
+      const rpcUrl = await Lit.Actions.getRpcUrl({
+        chain: unsignedTransaction.chain,
+      });
+      const provider = new ethers.providers.JsonRpcProvider(rpcUrl);
```
DashKash54, line 176: Reuse the provider from above

DashKash54 (changes requested):
Left a few comments

packages/wrapped-keys/src/lib/litActions.ts
DashKash54: It may have to do something with stringifying and returning from runOnce as hex isn't required by Ethers. Since you no longer have to use runOnce can you remove hexlify and see if you still get this error

local-tests/tests/testEthereumBroadcastWrappedKeyWithDefaultGasParams.ts
spacesailor24: > Fetching gasLimit doesn't work on Chronicle as you have to provide the from field. 
Yea that was it, fixed in [this commit](https://github.com/LIT-Protocol/js-sdk/pull/501/commits/f3c373f4b27c170dc50f40ecd6444083e3ed9cda)

packages/wrapped-keys/src/lib/litActions.ts
spacesailor24: Yea that was it, removed it

packages/wrapped-keys/src/lib/litActions.ts
spacesailor24: I tried pull it out and got the error
```
✖ 1. testEthereumBroadcastWrappedKeyWithFetchGasParams - Failed after 1 attempts (20303.62 ms)
❌Error: Error: Lit Action threw an unexpected error: {"message":"There was an error getting the signing shares from the nodes","errorCode":"unknown_error"}
Test Report: Some tests failed.
- testEthereumBroadcastWrappedKeyWithFetchGasParams (Failed in 20303.62 ms) - Error: Error: Lit Action threw an unexpected error: {"message":"There was an error getting the signing shares from the nodes","errorCode":"unknown_error"}
```
tried converting the LA functions to regular functions instead of arrow function incase it was a scoping issue, but same error

local-tests/tests/testEthereumBroadcastWrappedKeyWithFetchGasParams.ts
```diff
@@ -3,19 +3,19 @@ import { ethers } from 'ethers';
...
+ * ❌ NETWORK=localchain yarn test:local --filter=testEthereumBroadcastWrappedKeyWithFetchGasParams
```
DashKash54, line 16: Yeah I'll fix it in the SDK and then there's corr fix in the Lambda as well @MaximusHaximus

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,194 +1,232 @@
...
+  // // TODO: Remove the below which is only for demonstrating the error
+  // console.log('accessControlConditions: ', accessControlConditions);
+  // console.log('ciphertext: ', ciphertext);
+  // console.log('dataToEncryptHash: ', dataToEncryptHash);
+  // console.log('decrypted: ', decrypted);
+
+  Lit.Actions.setResponse({
+    response: resp,
+  });
+}).toString();
```
DashKash54, line 45: Let's define it only as a function and then do `toString()` when calling it

packages/wrapped-keys/src/lib/litActions.ts
DashKash54: Let's keep it like this then

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,194 +1,232 @@
...
-            }
-        )
+  if (!unsignedTransaction.chainId) {
+    Lit.Actions.setResponse({
+      response: 'Error: Missing required field: chainId',
+    });
+    return;
+  }
+
+  const decryptedPrivateKey = await Lit.Actions.decryptToSingleNode({
```
DashKash54, line 84: This can be wrapped in a try/catch

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,194 +1,232 @@
...
+    from: wallet.address,
+    value: ethers.utils.hexlify(
+      ethers.utils.parseEther(unsignedTransaction.value)
+    ),
+    chainId: unsignedTransaction.chainId,
+    data: unsignedTransaction.dataHex,
+    nonce,
+  };
+
+  const rpcUrl = await Lit.Actions.getRpcUrl({
```
DashKash54, line 125: This will throw an error when the chain isn't support in the node's rpc.yaml, surround by try/catch

packages/wrapped-keys/src/lib/litActions.ts
```diff
@@ -1,194 +1,232 @@
...
-    console.log('decryptedPrivateKey');
-    console.log(decryptedPrivateKey);
-
-    if (!decryptedPrivateKey) { // Exit the nodes which don't have the decryptedData
-        return;
+  } catch (err) {
+    const errorMessage = 'Error: When signing transaction- ' + err.message;
+    Lit.Actions.setResponse({ response: errorMessage });
+  }
+}).toString();
```
DashKash54, line 171: Same let's not convert to string right away

DashKash54 (changes requested):
Left a few comments but looks good now. It work on Cayenne right?

packages/wrapped-keys/src/lib/litActions.ts
DashKash54: Also add a test for when it fails

spacesailor24:
> Left a few comments but looks good now. It work on Cayenne right?

@DashKash54 yes, the tests are successful against cayenne, and I've also tested sending txs on Sepolia manually using cayenne as well

packages/wrapped-keys/src/lib/litActions.ts
DashKash54: This is no longer done this way right since we exit if the node is not selected for decryption. Resolving

PR #502, opened 2024-06-10
LIT-3394 - Improve connect() performance

MaximusHaximus:
# Description

### Improves connect() performance by parallelizing some operations performed during connection to the LIT network.
- Solution leverages the existing `contractContext`functionality to maintain backwards compatibility with consumers who use that API
- Fixed `config.rpcUrl` being optional string, but also a union with `null` -- this conflicted with the definition of same in the contract SDK and it made no sense to default an optional parameter to `null` anyway 🤦  :)
- Moved verification of the `litNetwork` to the constructor rather than it being nestled inside of `setConfig()` where it would explode the first time someone actually connected with the client after configuring it with an invalid network
- Removed `setNewConfig`, which was an internal method that would leave the core / client in an inconsistent state if ever called outside of the `connect()` flow.
- Standardized initialization so that we no longer have different _looking_ branches that actually _behave identically_ for all networks.  It turns out `setConfig()` was much less complex in _practice_ than it looked at first glance.
- Note that this appears to remove the 'using custom `bootstrapUrls`' logic -- but this functionality was _already not functional_; we always overrode `config.bootstrapUrls` with values returned from the chain or the lit-general-worker during `connect()` processing.

## Type of change

# How Has This Been Tested?

- [x] Ran local tests against Habanero and Cayenn

# Checklist:

### DIFF
--- a/local-tests/setup/tinny-environment.ts
+++ b/local-tests/setup/tinny-environment.ts
@@ -199,7 +199,6 @@ export class TinnyEnvironment {
       const networkContext = this._testnet.ContractContext;
       this.litNodeClient = new LitNodeClient({
         litNetwork: 'custom',
-        bootstrapUrls: this.processEnvs.BOOTSTRAP_URLS,
         rpcUrl: this.processEnvs.LIT_RPC_URL,
         debug: this.processEnvs.DEBUG,
         checkNodeAttestation: false, // disable node attestation check for local testing
--- a/packages/contracts-sdk/src/lib/contracts-sdk.ts
+++ b/packages/contracts-sdk/src/lib/contracts-sdk.ts
@@ -602,7 +602,7 @@ export class LitContracts {
 
       return new ethers.Contract(address, abi, provider);
     } else {
-      // if we have contract context then we determine if there exists a `resolverAddres`
+      // if we have contract context then we determine if there exists a `resolverAddress`
       // if there is a resolver address we assume we are using a contract resolver for bootstrapping of contracts
       if (!context.resolverAddress) {
         const stakingContract = (context as LitContractContext).Staking;
@@ -642,7 +642,6 @@ export class LitContracts {
     provider: ethers.providers.JsonRpcProvider,
     contractNames?: (keyof LitContractContext)[]
   ): Promise<LitContractContext> {
-    const rpcUrl = DEFAULT_RPC;
     const resolverContract = new ethers.Contract(
       context.resolverAddress,
       context.abi,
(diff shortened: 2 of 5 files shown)


joshLong145 (commented):
Looks good, one question i have is if we want to reuse the `staking contract` instance in the static `contract-sdk` calls. This in theory will save us some cycles in new object creation but  curious your thoughts.

MaximusHaximus:
> Looks good, one question i have is if we want to reuse the `staking contract` instance in the static `contract-sdk` calls. This in theory will save us some cycles in new object creation but curious your thoughts.

I looked into that, actually -- but it looks like the only time actual Ethers contract instances are maintained is inside use the non-static version of the contract SDK instance objects -- the instance does take in a contractContext object, and creates actual ethers.js Contract instances and keeps them in scope.

I actually considered just changing `core` to use an instance of the contract SDK instead for this reason, because that's actually the behaviour we _want_ here -- but it would mean that `core` would then keep copies of _all_ contract instances in scope and all it _actually needs_ is the StakingContract, and the instance logic has a bunch of other complexity in it that seemed superfluous for the usage we need in `core`

I wanted to keep respecting the existing `rpcUrl` and `contractContext` logic, which is entangled inside of the `getValidators` `getStakingContract` and `getMinNodeCount` methods.

I thought about in-lining the logic from those methods into `lit-core` to ensure we only ever created one-and-only-one contract instance per core instance, but `getValidators` is actually a fair amount of code and I didn't want to duplicate it.

I think maybe the cleanest way to handle this would be to allow actual contract instances to be passed in `LitContractContext` instead of just the `LitContract` definitions of those Contract instances -- then update any method that takes a `LitContractContext` to use the Contract instance provided rather than creating its own internally if they exist.

I didn't undertake this as part of the PR because the creation of the extra contract objects was 'cheap' (I couldn't benchmark any difference at all in performance with 1 instance vs. having the methods create extras, though there is obviously some temporary memory usage overhead), and it seemed like a bit of a rabbit-hole to go down given our desire to ship the connect enhancements promptly.  WDYT?

joshLong145 (dismissed):

Ansonhkg:
has this been tested with Shiva?

joshLong145 (approved):

joshLong145:
> has this been tested with `localchain` & Shiva?

I pulled the branch and ran tests on `manzano` to see how the connection speeds where with `sev attestation` happening and with the added concurrency I can reproduce the described speed improvements.

PR #504, opened 2024-06-11
Add `signMessage` and `signTransaction` Lit Action exports for Wrapped Keys

spacesailor24:
**NOTE** Before this PR is merged, we need to pin the bundled Lit Action code via Lit's official Pinata account (or similar), as they are currently pinned via my personal Pinata account

---

Adds two exported `const`s from `wrapped-keys` package:

1. `signTransactionWithSolanaEncryptedKeyLitActionIpfsCid`
    - This CID refers to the bundled Lit Action code to sign and submit a Solana transaction using a decrypted private key within the Lit Action
    - This Lit Actions accepts the parameters:
        ```
        accessControlConditions,
        ciphertext,
        dataToEncryptHash,
        serializedTransaction,
        broadcast,
        solanaNetwork,
       ```
       - `serializedTransaction` by required a serialized transaction, we can support any Solana transaction signing (i.e. transfers or contract executions)
       - `broadcast` determines whether or not the signed transaction is submitted to the `solanaNetwork`
       - `solanaNetwork` is the Solana network to connect to
           - Options are: `mainnet-beta`, `testnet`, and `devnet`
3. `signMessageWithSolanaEncryptedKeyLitActionIpfsCid`
    - This CID refers to the bundled Lit Action code to sign a message using a decrypted private key within the Lit Action
    - This Lit Actions accepts the parameters:
        ```
        accessControlConditions,
        ciphertext,
        dataToEncryptHash,
        messageToSign
       ```
       - `messageToSign` can be any string and is not expected to be encoded any specific way (it will be converted to a `Buffer` before signing within the Lit Action)

---

I tested that the sign tx and submit Lit Action works manually by providing it a testnet private key to the `testSignTransactionWithSolanaEncryptedKey` test and enabled `broadcast` ([tx receipt](https://explorer.solana.com/tx/xh2oGC2hB2sDdjdTb1nDT6SVV7dapzVU43gtpzuKHgNsMbwYW4vEoqEvS69JTGjLhRPCzjYysdbyLDUGnMRuXua?cluster=devnet))

---

For organizational purposes, I renamed the Eth Lit Actions file to `ethereum.ts` and moved it under a new `litActions` directory. The Solana Lit Actions are located under `litActions/Solana`


DashKash54:
> solanaNetwork is the Solana network to connect to

Let's add it to the `SolanaLitTransaction` interface? Then it can be part of the unsigned transaction params?

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,70 @@
+const { Keypair } = require('@solana/web3.js');
+const bs58 = require('bs58');
+const nacl = require('tweetnacl');
```
DashKash54, line 3: Where have these been injected? It needs to be bundled in the Lit Action code before publishing to IPFS

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,70 @@
...
+
+(async () => {
+  const LIT_PREFIX = 'lit_';
+
```
DashKash54, line 7: Let's add some check for user params passes in the jsPrams like in `signTransactionWithEthereumEncryptedKeyLitAction`

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,70 @@
...
+    decryptedPrivateKey = await Lit.Actions.decryptToSingleNode({
+      accessControlConditions,
+      chain: 'ethereum',
+      ciphertext,
+      dataToEncryptHash,
+      authSig: null,
+    });
+  } catch (error) {
+    Lit.Actions.setResponse({
+      response: `Error decrypting data to private key: ${error.message}`,
```
DashKash54, line 19: Standardizing this since the tests depend on checking the extract string including the ":"
```suggestion
      response: `Error: When decrypting data to private key: ${error.message}`,
```

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,70 @@
...
+  if (!decryptedPrivateKey) {
+    // Exit the nodes which don't have the decryptedData
+    return;
+  }
+
+  const privateKey = decryptedPrivateKey.startsWith(LIT_PREFIX)
+    ? decryptedPrivateKey.slice(LIT_PREFIX.length)
+    : decryptedPrivateKey;
+  const solanaKeyPair = Keypair.fromSecretKey(
+    Uint8Array.from(Buffer.from(privateKey, 'hex'))
```
DashKash54, line 33: Why do we need a buffer? We don't need it if we just start by encrypting it as a hex?

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
```diff
@@ -0,0 +1,131 @@
...
+        comparator: '=',
+        value: await alice.wallet.getAddress(),
+      },
+    },
+  ];
+
+  const { ciphertext, dataToEncryptHash } = await encryptString(
+    {
+      accessControlConditions: decryptionAccessControlConditions,
+      dataToEncrypt: Buffer.from(solanaKeypair.secretKey).toString('hex'),
```
DashKash54, line 54: Same comment as in the above file, why buffer?

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
```diff
@@ -0,0 +1,131 @@
...
+        walletAddress: await alice.wallet.getAddress(),
+        nonce: await devEnv.litNodeClient!.getLatestBlockhash(),
+        litNodeClient: devEnv.litNodeClient,
+      });
+      return await generateAuthSig({
+        signer: alice.wallet,
+        toSign,
+      });
+    },
+  });
```
DashKash54, line 85: Same as in the above file

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const solanaTransaction = new Transaction();
+  solanaTransaction.add(
+    SystemProgram.transfer({
+      fromPubkey: solanaKeypair.publicKey,
+      toPubkey: new PublicKey(solanaKeypair.publicKey),
+      lamports: LAMPORTS_PER_SOL / 100, // Transfer 0.01 SOL
+    })
+  );
+  solanaTransaction.feePayer = solanaKeypair.publicKey;
```
DashKash54, line 95: What all should be part of the existing interface `SolanaLitTransaction`?

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const solanaConnection = new Connection(clusterApiUrl('devnet'), 'confirmed');
+  const { blockhash } = await solanaConnection.getLatestBlockhash();
+  solanaTransaction.recentBlockhash = blockhash;
```
DashKash54, line 99: Is it mandatory? What if we don't provide the blockhash?

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const serializedTransaction = solanaTransaction
+    .serialize({
+      requireAllSignatures: false,
+      verifySignatures: false,
+    })
+    .toString('base64');
```
DashKash54, line 106: Can we do this inside the `signTransactionWithSolanaEncryptedKey` litAction as the user is expected to provide only an unsigned transaction in the interface `SolanaLitTransaction`? The idea is to model is in a similar approach as `signTransactionWithEncryptedKey`

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,70 @@
...
+
+  let signature;
+  try {
+    signature = nacl.sign.detached(
+      new TextEncoder().encode(messageToSign),
+      solanaKeyPair.secretKey
+    );
+  } catch (error) {
+    Lit.Actions.setResponse({
+      response: `Error signing message: ${error.message}`,
```
DashKash54, line 44:
```suggestion
      response: `Error: When signing message: ${error.message}`,
```

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,70 @@
...
+    if (!isValid) {
+      Lit.Actions.setResponse({
+        response:
+          'Error: Signature did not verify to expected Solana public key',
+      });
+      return;
+    }
+  } catch (error) {
+    Lit.Actions.setResponse({
+      response: `Error checking if signed message is valid: ${error.message}`,
```
DashKash54, line 64:
```suggestion
      response: `Error: When validating signed message is valid: ${error.message}`,
```

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,60 @@
+const {
+  clusterApiUrl,
+  Connection,
+  Keypair,
+  Transaction,
+} = require('@solana/web3.js');
+
+(async () => {
+  const LIT_PREFIX = 'lit_';
+
```
DashKash54, line 10: Let's add some check for user params passes in the jsPrams like in signTransactionWithEthereumEncryptedKeyLitAction

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,60 @@
...
+    decryptedPrivateKey = await Lit.Actions.decryptToSingleNode({
+      accessControlConditions,
+      chain: 'ethereum',
+      ciphertext,
+      dataToEncryptHash,
+      authSig: null,
+    });
+  } catch (error) {
+    Lit.Actions.setResponse({
+      response: `Error decrypting data to private key: ${error.message}`,
```
DashKash54, line 22:
```suggestion
      response: `Error: When decrypting data to private key: ${error.message}`,
```

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,60 @@
...
+  const privateKey = decryptedPrivateKey.startsWith(LIT_PREFIX)
+    ? decryptedPrivateKey.slice(LIT_PREFIX.length)
+    : decryptedPrivateKey;
+  const solanaKeyPair = Keypair.fromSecretKey(
+    Uint8Array.from(Buffer.from(privateKey, 'hex'))
+  );
+
+  try {
+    const transaction = Transaction.from(
+      Buffer.from(serializedTransaction, 'base64')
```
DashKash54, line 41: Same as in the above file, why buffer to begin with?

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,60 @@
...
+    const signature = transaction.signature.toString('base64');
+    Lit.Actions.setResponse({ response: signature });
+
+    if (broadcast) {
+      const solanaConnection = new Connection(
+        clusterApiUrl(solanaNetwork),
+        'confirmed'
+      );
+      await solanaConnection.sendRawTransaction(transaction.serialize());
+    }
```
DashKash54, line 54: We should return the signedMessage when broadcast is false like we do in `signTransactionWithEthereumEncryptedKeyLitAction`:
```
Lit.Actions.setResponse({ response: signedTx });
```

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,60 @@
...
+  } catch (error) {
+    Lit.Actions.setResponse({
+      response: `Error during transaction signing and submission: ${error.message}`,
```
DashKash54, line 57:
```suggestion
      response: `Error: During transaction signing and submission: ${error.message}`,
```

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,60 @@
...
+    transaction.sign(solanaKeyPair);
+    const signature = transaction.signature.toString('base64');
+    Lit.Actions.setResponse({ response: signature });
+
+    if (broadcast) {
+      const solanaConnection = new Connection(
+        clusterApiUrl(solanaNetwork),
+        'confirmed'
+      );
+      await solanaConnection.sendRawTransaction(transaction.serialize());
```
DashKash54, line 53: We're not waiting for tx confirmation right? We shouldn't

DashKash54 (changes requested):
Left a bunch of comments. The idea is that I use these LitAction (ipfs_cid) in the functions: `signTransactionWithEncryptedKey`, `signMessageWithEncryptedKey`

DashKash54:
Also the Linter is failing, please run `yarn prettier`

local-tests/tests/testSignMessageWithSolanaEncryptedKey.ts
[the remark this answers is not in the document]
spacesailor24: Yes, this is what I was asking before. I wasn't sure if I should extend your existing helper functions, or if I should write new ones for Solana

local-tests/tests/testSignMessageWithSolanaEncryptedKey.ts
```diff
@@ -0,0 +1,107 @@
...
+        comparator: '=',
+        value: await alice.wallet.getAddress(),
+      },
+    },
+  ];
+
+  const { ciphertext, dataToEncryptHash } = await encryptString(
+    {
+      accessControlConditions: decryptionAccessControlConditions,
+      dataToEncrypt: Buffer.from(solanaKeypair.secretKey).toString('hex'),
```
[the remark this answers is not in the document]
spacesailor24: A `Uint8Array` is returned from `Keypair.generate();`, so I'm converting `solanaKeypair.secretKey` from a Buffer to a hex string

local-tests/tests/testSignMessageWithSolanaEncryptedKey.ts
```diff
@@ -0,0 +1,107 @@
...
+        walletAddress: await alice.wallet.getAddress(),
+        nonce: await devEnv.litNodeClient!.getLatestBlockhash(),
+        litNodeClient: devEnv.litNodeClient,
+      });
+      return await generateAuthSig({
+        signer: alice.wallet,
+        toSign,
+      });
+    },
+  });
```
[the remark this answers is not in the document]
spacesailor24: Thank you, wasn't sure what the correct helper function was

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
spacesailor24: A Uint8Array is returned from Keypair.generate();, so I'm converting solanaKeypair.secretKey from a Buffer to a hex string

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
spacesailor24: Solana txs appear to be very different from ETH txs. It seems there are specific instructions depending on what type of transaction you're trying to submit, and they have different requirements. Because of this, I don't think we should try to create a `SolanaLitTransaction`, and what I've opted for is for the user to just pass their serialized transaction that will be signed within the Lit Action (that way the Lit Action doesn't need to care about the type of tx being sent, it just signs it arbitrarily)

image

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
spacesailor24: Yes it's mandatory, the blockhash is used to ensure the transaction is processed within a specific timeframe and helps prevent replay attacks. Without it the tx is invalid

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
spacesailor24: No I don't think we should. As mentioned [here](https://github.com/LIT-Protocol/js-sdk/pull/504/files#r1640566254) Solana txs aren't just simple objects like ETH txs regardless of what you're trying to do on chain. I believe it's easiest for the user to use Solana's web3.js library to create their specific transaction instance (with all the Solana instructions), and then serialize it and provide it to the Lit Action for signing

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
spacesailor24: When you install the package there's a `postinstall` NPM script that executes the bundler (can also be called manually via `yarn bundle` within the `wrapped-keys` package). Afterwards, the bundled version of these files are located in `packages/wrapped-keys/src/lib/litActions/solana/dist` and are currently manually uploaded to IPFS

I think we should create a script to publish to IPFS, but I figured that could come after this PR is merged

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
spacesailor24: The user params required are:
```
jsParams: {
      accessControlConditions,
      ciphertext,
      dataToEncryptHash,
      messageToSign,
    },
```
do you want me to check for all of these? The `signMessageWithEthereumEncryptedKeyLitAction` doesn't check for any of them, should it be updated to?

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
spacesailor24: Updated to
```ts
const solanaKeyPair = Keypair.fromSecretKey(Buffer.from(privateKey, 'hex'));
```
It is encrypted as hex, but the Solana library requires it to be passed as a `Uint8Array`

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
spacesailor24: Similar question [here](https://github.com/LIT-Protocol/js-sdk/pull/504/files#r1640572166), but the required params are:

```
jsParams: {
      accessControlConditions: decryptionAccessControlConditions,
      ciphertext,
      dataToEncryptHash,
      serializedTransaction,
      broadcast: false,
      solanaNetwork: 'devnet',
    }
```

Probably makes sense to check `solanaNetwork` is one of the three possible options (`devnet`, `testnet`, `mainnet-beta`). Of course `serializedTransaction` is serialized, so there are no tx params to check for

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
spacesailor24: The Solana library pretty much only intakes `Uint8Array`s. `serializedTransaction` is being passed to the Lit Action as a hex string and is being converted to a Buffer here as required

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
spacesailor24: No we just submit the transaction and return the signature

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
spacesailor24: It's already being set as the response above  the `if (broadcast) {`:

```ts
transaction.sign(solanaKeyPair);
const signature = transaction.signature.toString('base64');
Lit.Actions.setResponse({ response: signature });
```

packages/wrapped-keys/src/index.ts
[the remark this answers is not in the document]
spacesailor24: We can un-minify the code deployed to IPFS, but that doubles the size of the files from < .5 MB to 1 MB

They can always view the un-bundled source code by looking at this repo, and if they really wanted to, could bundle the code from the `wrapped-keys` package and check the IPFS hash matched for each bundle file

Eventually we should add a script for bundling and uploading the files to IPFS and checking the hashes match to make the process easier, but I don't think there will be a straightforward solutions for this since we have to bundle

local-tests/test.ts
```diff
@@ -102,6 +102,8 @@ import { testFailImportWrappedKeysWithInvalidSessionSig } from './tests/testFail
 import { testFailImportWrappedKeysWithExpiredSessionSig } from './tests/testFailImportWrappedKeysWithExpiredSessionSig';
 import { testExportWrappedKey } from './tests/testExportWrappedKey';
 // import { testRemoveDecryptFailingOutsideLitAction } from './tests/testRemoveDecryptFailingOutsideLitAction';
+import { testSignMessageWithSolanaEncryptedKey } from './tests/testSignMessageWithSolanaEncryptedKey';
+import { testSignTransactionWithSolanaEncryptedKey } from './tests/testSignTransactionWithSolanaEncryptedKey';
```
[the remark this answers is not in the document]
spacesailor24: > Let's add another test for only signing and not broadcasting like signTransactionWithEncryptedKey

This is what the current `testSignTransactionWithSolanaEncryptedKey` is doing. It has `broadcast` set to `false` since sending a transaction would require testnet funds on Solana devnet

local-tests/tests/testSignMessageWithSolanaEncryptedKey.ts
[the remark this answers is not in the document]
spacesailor24: Addressed via [this commit](https://github.com/LIT-Protocol/js-sdk/pull/504/commits/c538a1e8d6011a0929e13e2d4c82ae7241d45cf1)

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
spacesailor24: Addressed via [this commit](https://github.com/LIT-Protocol/js-sdk/pull/504/commits/c538a1e8d6011a0929e13e2d4c82ae7241d45cf1)

local-tests/test.ts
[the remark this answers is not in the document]
DashKash54: Okay we aren't testing broadcasting in the tests? You just tested manually?

local-tests/tests/testSignMessageWithSolanaEncryptedKey.ts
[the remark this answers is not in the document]
DashKash54: It should be `getPkpSessionSigs` which gives you the PKP sessionSigs not EoaSessionSig as we only allow PKP sessionSigs to use Wrapped Key @spacesailor24

local-tests/tests/testSignMessageWithSolanaEncryptedKey.ts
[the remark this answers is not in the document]
DashKash54: You can use the same ones as I'm using Generics: `signTransactionWithEncryptedKey<T = LitTransaction>`. It accepts both Eth & Solana transaction types. So you should update the `SolanaLitTransaction` is required as well

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
DashKash54: Yeah we can write the IPFS uploading script later, out of scope for this PR

DashKash54: > postinstall NPM script that executes the bundle

Which packages are being bundled? We should only bundle the necessary ones as we're uploading it to IPFS so the file size should be as small as possible

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
DashKash54: Makes sense as these will be added internally when calling the function `signMessageWithEncryptedKey`. That's why we should use these functions as the functionality is already present for both Eth & Solana

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
DashKash54: As commented above, it should be `getPkpSessionSigs` which gives you the PKP sessionSigs not EoaSessionSig as we only allow PKP sessionSigs to use Wrapped Key

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
DashKash54: Sure Eth tx have other fields too: https://docs.ethers.org/v5/api/providers/types/#providers-TransactionRequest
But only some fields are mandatory in Solana just like Eth and the rest are autofilled like in Ethers?
What I meant was adding only the mandatory ones, which you're using above: `fromPubkey`, `toPubkey`, `lamports` and `feePayer` to `SolanaLitTransaction`
What do you think about it?

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
DashKash54: Based on my response: https://github.com/LIT-Protocol/js-sdk/pull/504#discussion_r1641997975
How about us setting the other params as default inside the Lit Action?

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
DashKash54: Yes let's add a check for `solanaNetwork`

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
DashKash54: Would it make sense to update the `signTransactionWithEncryptedKey` param to take Uint8Arrays?

local-tests/test.ts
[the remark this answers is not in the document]
spacesailor24: @DashKash54 Correct. We'd need to extend the testing infra to generate a Solana key with test funds

.prettierignore
```diff
@@ -11,4 +11,5 @@
 /tmp
 /packages/*/dist
 .nx
-tools
\ No newline at end of file
+tools
+**/*/dist
```
DashKash54, line 15: So we aren't adding the minified code that we publish?

local-tests/tests/testSignMessageWithSolanaEncryptedKey.ts
DashKash54, line 1: Need to use pkpSessionSig not eoaSessionSigs

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
DashKash54, line 1: Need to use pkpSessionSig not eoaSessionSigs

DashKash54 (commented):
Few more comments

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
spacesailor24:
```ts
const { Keypair } = require('@solana/web3.js');
const bs58 = require('bs58');
const nacl = require('tweetnacl');
```

@DashKash54  These imports are bundled and are being used by the LA. Unminified, the bundle is 1 MB, bundled it's about 400 KB

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
spacesailor24: > autofilled like in Ethers

The Solana SDK will autofill some of the properties, but you still have to instantiate one of the above `SystemProgram`s and these all require specific params. Those parameters I'm using in my tests are what's required to send a `transfer` `SystemProgram`, but  just using those same properties won't mean you can send the other types of txs

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
spacesailor24: Given how many different types of transactions there are in the Solana SDK, I still think we should just have the user create and serialize their tx, then pass it to the LA for signing

.prettierignore
spacesailor24: No, it bloats the package size for no real benefit since the bundled code can be reproduced with `yarn bundle`

packages/wrapped-keys/src/lib/litActions/solana/src/signMessageWithSolanaEncryptedKey.js
DashKash54: We can restrict the required function from nacl

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
DashKash54: `SolanaLitTransaction` should accept the serialized Solana tx provided by the user

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
DashKash54: `requireAllSignatures ` & `verifySignatures ` should be false to submit an unsigned tx. Otherwise, it needs to be signed by the user

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
DashKash54: `mainnet-beta`  is the actual mainnet

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
DashKash54: If the serialized Solana tx are uint8Arrays we should standardize to accept Uint8Array instead of string even for Eth tx

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
DashKash54: Using serialized Tx for Solana

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
DashKash54: We're passing it as a Json value so can't be a Uint8Array we need to convert it to either a string or native array

PR #505, opened 2024-06-12
fix(core): contract context

joshLong145:
Fixes a small bug from pr #502 where we would not respect the `contractResolver` address if provided which breaks some local development flows.


Ansonhkg (approved):
That makes sense. Thanks!

PR #506, opened 2024-06-12
Fix/get session sigs base provider block hash url

joshLong145:
- Makes `authNeededCallback` optional for `BaseProvider` in `LitAuthClient`
- Switches URL for block hash service to `https` over `http`


Ansonhkg (approved):
this makes sense!

PR #507, opened 2024-06-13
use eth blockhash as webauthn challenge

glitch003:
This PR makes a few changes related to webauthn:

* Use eth blockhash instead of chronicle, by default.  This is really important, since we check for the last N blockhashes and chronicle creates a new block every time a txn is made.  So it's possible for the blockhash to not be recent enough by the time the node sees the request.  Instead, we use eth blockhashes, which are already available.
* Remove nanoid.  We don't need a package to generate a random string, it's really easy in vanilla JS.  And the dynamic import causes problems.
* Remove the concept of `rpcUrl` from the Lit Auth Client.  It's only used to get the chronicle blockhash, and we don't need that anymore.
* Make the blockhash fetcher use HTTPS.  I had to do this to get my test to work in the browser, but we should do this anyway.  

I've already pushed fixes to cayenne to look for the eth blockhash, so this should be working on cayenne.  I will get it deployed to manzano asap too.

This should be fully working with the `rc-v0.1-2024-04-23` branch of lit-assets, which is currently deployed to cayenne.

This is published as [6.0.2-beta.6](https://www.npmjs.com/package/@lit-protocol/lit-node-client/v/6.0.2-beta.6)

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,5 +2,5 @@
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "useNx": true,
   "useWorkspaces": true,
-  "version": "6.0.2"
+  "version": "6.0.2-beta.6"
 }
--- a/lerna.json
+++ b/lerna.json
@@ -2,5 +2,5 @@
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "useNx": true,
   "useWorkspaces": true,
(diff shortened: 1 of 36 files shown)


Ansonhkg (approved):
Nice!

joshLong145 (approved):

PR #509, opened 2024-06-19
Feature/lit 3140 make pkp base a private instance variable in pkp

FedericoAmura:
# Description

This PR cleans PKPBase authContext and its ex-extending clases PKPEthersWallet, PKPCosmosWallet and PKPSuiWallet. Including changes:
- Removed extension from PKPBase, those classes now implement a shared interface and have a more restricted public interface
- Make `litNodeClient` a required param in PKP wallets and remove fields to create it, delegating that responsibility to user
- Removed redundant litNodeClient instance inside `authContext`
- Made PKPBase final (with a private constructor, there is no final in TS)
- Updated PKPClient due to these changes
- Fixed several type definitions
- Updated session sigs demo app to use all PKP wallet classes to test signing

This PR has BREAKING CHANGES due to changes in PKP*Wallet interfaces and removed things (WalletFactory)

## Type of change

# How Has This Been Tested?

- [x] Building and testing passing
- [x] New type validations SDK wide
- [x] session sigs demo updated to test signing using PKPEthersWallet, PKPCosmosWallet and PKPSuiWallet

Screenshot from 2024-06-21 13-09-42

# Checklist:


Ansonhkg (commented):
Nice! I like that fact that we are passing the `pkpBase` as an instance now instead of extending it, and also making `litNodeClient` a required param so we stop fiddling around the default configuration underneath.

Ansonhkg (approved):
Ran all tests using Tinny and all passed! Approved!

Command ran:
`NETWORK=cayenne yarn test:local --filter=testPkpEthers`

packages/pkp-ethers/src/lib/helper.ts
```diff
@@ -24,8 +24,10 @@ export function convertHexToUtf8(value: string): string {
  * @param {any} txParams - The original transaction parameters.
  * @returns {any} The transaction object with 'gas' and 'from' fields removed.
  */
-export const getTransactionToSign = (txParams: any): any => {
-  let formattedTx = Object.assign({}, txParams);
+export const getTransactionToSign = <T>(
+  txParams: T & { gas?: string; from?: string }
```
joshLong145, line 28: I think we can use the `UnsignedTransaction` type?

joshLong145 (approved):
🔥

packages/pkp-ethers/src/lib/helper.ts
FedericoAmura: Kind of, the source of the param that reaches this function is a `TransactionRequest` or an `any` nested inside a `ETHRequestSigningPayload`. So, it feels like yes, we should use that type, but in reality it does not comply with the function code. That is because `UnsignedTransaction` has `gasLimit` and `gasPrice`, not `gas`

Considering the function deletes `gas` when it is there, I would say the function has to be checked to have the right logic we want (deleting `gas` instead of `gasSomething`) and I think we better do those kind of things in other tickets, if needed, to not add extra delay and complexity

MaximusHaximus (commented):
:+1: This is awesome dude, nice work.

PR #512, opened 2024-06-24
check timestamps of BLS session sigs

glitch003:
A user found a bug where the SDK was sending an expired BLS session sig.  I discovered that we aren't checking the expiration when we are checking if we need to resign the session key.  We do this check for ECDSA, because it's built into SIWE, but we can't use SIWE verification for the BLS session sig, because SIWE doesn't understand BLS.  

So we're checking the sig already, and this PR just copy/pastes the timestamp verification code from the SIWE repo here: https://github.com/spruceid/siwe/blob/main/packages/siwe/lib/client.ts#L289

This makes it so that we check the expiration and issued at time, and then throw an error if they're not valid.  The calling function (`checkNeedToResignSessionKey()`) catches errors and will try to re-sign if the SIWE is expired.

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/helpers/validate-bls-session-sig.ts
+++ b/packages/lit-node-client-nodejs/src/lib/helpers/validate-bls-session-sig.ts
@@ -1,6 +1,7 @@
 import { AuthSig } from '@lit-protocol/types';
 import { uint8arrayToString } from '@lit-protocol/uint8arrays';
 import { ethers } from 'ethers';
+import { SiweError, SiweErrorType, SiweMessage } from 'siwe';
 
 const LIT_SESSION_SIGNED_MESSAGE_PREFIX = 'lit_session:';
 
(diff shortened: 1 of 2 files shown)


packages/lit-node-client-nodejs/src/lib/helpers/validate-bls-session-sig.ts
```diff
@@ -32,6 +34,34 @@ export const blsSessionSigVerify = (
...
+  }
+
+  const notBefore = new Date(authSigSiweMessage.notBefore);
+  if (checkTime.getTime() < notBefore.getTime()) {
+    throw new SiweError(
+      SiweErrorType.NOT_YET_VALID_MESSAGE,
+      `${checkTime.toISOString()} >= ${notBefore.toISOString()}`,
+      `${checkTime.toISOString()} < ${notBefore.toISOString()}`
+    );
+  }
```
joshLong145, line 63: Should we be checking on expiration in this scope? as this type of exp check is relevant for all sig types and not just BLS?

joshLong145 (approved):

packages/lit-node-client-nodejs/src/lib/helpers/validate-bls-session-sig.ts
glitch003: the other sig type (ecdsa) is already checked by the SIWE verify function so i think we're good

joshLong145: This makes sense. Will approve ❤️

joshLong145 (dismissed):

MaximusHaximus (approved):
:+1: LGTM -- :shipit: 🚢

PR #513, opened 2024-06-26
Wrapped Keys

MaximusHaximus:
# Description

- Updates wrapped keys to support multiple networks; Production for Habanero (and Datil in the future), TestNetworks for Cayenne and Manzano
- Adds new `keyType` and `publicKey` keys to the persisted metadata, and requires they are provided when consumers call the methods that require them
- Organizes types, constants, and logic by domain -- `lit-actions-client` contains all code specific to interacting with the lit actions, `service-client` contains all of the same for interactions with the wrapped keys backend, and `api` contains definitions of all top-level API methods that will be exposed to the consumer of the package
- Explicitly exporting types and methods from the top-level `index.ts` rather than blanket exporting everything in the package with * exports.  The goal here is to facilitate us being able to modify the LIT action APIs without making breaking changes to the top-level API methods that consumers are using as we need to.
- Added a bunch of JSDocs and explicit types for parameters for API-exposed methods and their return values
- Removed a lot of duplicated JSDoc comments by using @typedef and @extends
- Removed the `customXXXX` methods, in favour of us instead add a basic documentation blurb to the JSDocs for prospective integrators that shows them how easy it is to get the private key metadata from the wrapped keys service, and then submit it to _any_ custom LIT action, using the LitNodeClient instance that they _already have in scope_.  The `custom` behaviour in the SDK exposes internals of LIT actions to our consumer API, and would cause us to get tickets against the wrapped-keys SDK that are actually issues with custom LIT Action implementations on top of making the API worse for consumers of the package who want to use 'supported' networks.

e.g.:
```javascript
import { getPrivateKeyMetadata } from '@lit-protocol/wrapped-keys'

import { litNodeClient } from '@lit-protocol/lit-node-client'

await litNodeClient.connect({litNetwork: 'cayenne'});
const sessionSigs = await litNodeClient.getSessionSigs(...);
const keyMetadata = await getPrivateKeyMetadata({ litNodeClient, sessionSigs})
const customLitActionWithWrappedKeyResponse = await litNodeClient.executeJs({
  sessionSigs,
  ipfsId: litActionIpfsCid, // OR
  code: litActionCode, 
  jsParams: {
    accessControlConditions,
    keyMetadata,
  },
});

// Do whatever you want to with `customLitActionWithWrappedKeyResponse
```

## Type of change

# How Has This Been Tested?

- Ran all local tests against Cayenne.  The only ones that fail say they _intentionally_ fail (?!?!?!)

# Checklist:


spacesailor24:
The [local-tests/tests/testImportWrappedKey.ts](https://github.com/LIT-Protocol/js-sdk/blob/ac8f17372a2c0a204286515e35b6abeb26e1effc/local-tests/tests/testImportWrappedKey.ts) test is importing a Solana key but says the `keyType` is `K256`

This test also isn't really testing if the key was imported correctly, but I think we'd have to use either `getEncryptedKeyMetadata` or the sign methods to verify it was

packages/wrapped-keys/src/index.ts
```diff
@@ -1,3 +1,69 @@
...
+
+export const api = {
+  exportPrivateKey,
+  generatePrivateKey,
+  getEncryptedKeyMetadata,
+  importPrivateKey,
+  signMessageWithEncryptedKey,
+  signTransactionWithEncryptedKey,
+  storeEncryptedKeyMetadata,
+};
```
spacesailor24, line 48: Curious why we're wrapping this in an object, does this make it not tree shakeable if a user is only using one or a few of these methods?

packages/wrapped-keys/src/lib/types.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @extends BaseApiParams
+ *
+ * @property { string | Uint8Array } messageToSign The message to be signed
+ */
+export type SignMessageWithEncryptedKeyParams = BaseApiParams &
+  ApiParamsSupportedNetworks &
+  SignMessageParams;
+
+interface BaseLitTransaction {
+  chain: string;
```
spacesailor24, line 119: It would be nice if we could restrict this to the [supported EVM networks](https://github.com/LIT-Protocol/js-sdk/blob/cc95304922d0e0b00300e0198de2455586b858a4/packages/constants/src/lib/constants/constants.ts#L29-L601) for `EthereumLitTransaction`s and `mainnet-beta | testnet | devnet` for Solana txs

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
```diff
@@ -82,7 +82,7 @@ const {
         clusterApiUrl(unsignedTransaction.chain),
         'confirmed'
       );
-      await solanaConnection.sendRawTransaction(transaction.serialize());
+      await solanaConnection.sendRawTransaction(transaction.serialize()); // FIXME: Shouldn't this return the tx hash for consistency with the evm action?
```
spacesailor24, line 85: Yes it should, however, in my testing waiting for the transaction hash always resulted in the Lit Action timing out. I'm going to add the following code to the docs to show users how to obtain the tx hash/receipt outside of the Lit Action:

```ts
// Assuming signature is in base64, convert it back to byte array for querying
const byteSignature = Buffer.from(signature, 'base64');

// Wait for confirmation and fetch the transaction details
const confirmation = await solanaConnection.confirmTransaction(byteSignature);
console.log('Transaction confirmation status:', confirmation.value);

// Get the transaction receipt
const receipt = await solanaConnection.getTransaction(byteSignature.toString('base64'), { commitment: 'confirmed' });
console.log('Transaction receipt:', receipt);
```

spacesailor24:
Are there tests for `getEncryptedKeyMetadata` and `storeEncryptedKeyMetadata`?

packages/wrapped-keys/src/lib/api/store-encrypted-key-metadata.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/** Get a previously encrypted and persisted private key and its metadata.
+ * Note that this method does _not_ decrypt the private key; only the _encrypted_ key and its metadata will be returned to the caller.
+ *
+ * @param { StoreEncryptedKeyMetadataParams } params Parameters required to fetch the encrypted private key metadata
+ * @returns { Promise<StoredKeyMetadata> } The encrypted private key and its associated metadata
+ */
+export async function storeEncryptedKeyMetadata(
+  params: StoreEncryptedKeyMetadataParams
+): Promise<boolean> {
```
spacesailor24, line 13: This should be `Promise<boolean>` based on the below implementation?

DashKash54: Isn't it already boolean?

packages/wrapped-keys/src/index.ts
DashKash54: @MaximusHaximus ^

packages/wrapped-keys/src/lib/types.ts
DashKash54: Sure for later, can you cut a ticket for tracking this @spacesailor24 and comment here

packages/wrapped-keys/src/lib/api/export-private-key.ts
```diff
@@ -0,0 +1,55 @@
...
+} from '../utils';
+
+/** Exports a previously persisted private key from the wrapped keys service for direct use by the caller, along with the keys metadata
+ *
+ * @param { ExportPrivateKeyParams } params Parameters required to export the private key
+ */
+export async function exportPrivateKey(
+  params: ExportPrivateKeyParams
+): Promise<ExportPrivateKeyResult> {
+  const { pkpSessionSigs, litNodeClient } = params;
```
DashKash54, line 19: nit: How about restructuring it the function declaration? This way it makes clear which fields are exactly being used?

packages/wrapped-keys/src/lib/api/export-private-key.ts
```diff
@@ -0,0 +1,55 @@
...
+    litNetwork: litNodeClient.config.litNetwork,
+  });
+
+  const { ciphertext, dataToEncryptHash, ...privateKeyMetadataMinusEncrypted } =
+    privateKeyMetadata;
+
+  const decryptedPrivateKey = await decryptToString(
+    {
+      accessControlConditions: [allowPkpAddressToDecrypt],
+      chain: CHAIN_ETHEREUM, // FIXME: This won't always be true.
```
DashKash54, line 36: I think it will as it's just the PKP Eth Address?

packages/wrapped-keys/src/lib/api/export-private-key.ts
```diff
@@ -0,0 +1,55 @@
...
+      dataToEncryptHash,
+      sessionSigs: pkpSessionSigs,
+    },
+    litNodeClient
+  );
+
+  // It will be of the form lit_<privateKey>
+  if (!decryptedPrivateKey.startsWith(LIT_PREFIX)) {
+    throw new Error(
+      `PKey was not encrypted with salt; all wrapped keys must be prefixed with '${LIT_PREFIX}'`
```
DashKash54, line 47: How about for Custom Wrapped Keys. They may choose not to follow our convention so this results in them encrypting their key and not being able to access it

packages/wrapped-keys/src/lib/api/export-private-key.ts
DashKash54, line 1: Why isn't the filename camel case?

packages/wrapped-keys/src/lib/api/generate-private-key.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const firstSessionSig = getFirstSessionSig(pkpSessionSigs);
+  const pkpAddress = getPkpAddressFromSessionSig(firstSessionSig);
+  const allowPkpAddressToDecrypt = getPkpAccessControlCondition(pkpAddress);
+
+  const { ciphertext, dataToEncryptHash, publicKey } =
+    await generateKeyWithLitAction({
+      ...params,
+      pkpAddress,
+      litActionIpfsCid: getLitActionCid(network, 'generateEncryptedKey'),
```
DashKash54, line 37: Exact in a const:`generateEncryptedKey`?

packages/wrapped-keys/src/lib/api/generate-private-key.ts
```diff
@@ -0,0 +1,57 @@
...
+      accessControlConditions: [allowPkpAddressToDecrypt],
+    });
+
+  await storePrivateKeyMetadata({
+    sessionSig: firstSessionSig,
+    storedKeyMetadata: {
+      ciphertext,
+      publicKey,
+      keyType: LIT_CURVE.EcdsaK256, // FIXME: Should be returned by the LIT action; we won't know what it is unless it's provided.
```
DashKash54, line 46: We should provide it then as we're storing the same for Solana as well which is wrong

packages/wrapped-keys/src/lib/api/import-private-key.ts
```diff
@@ -0,0 +1,54 @@
...
+  params: ImportPrivateKeyParams
+): Promise<string> {
+  const { pkpSessionSigs, privateKey, publicKey, keyType, litNodeClient } =
+    params;
+
+  const firstSessionSig = getFirstSessionSig(pkpSessionSigs);
+  const pkpAddress = getPkpAddressFromSessionSig(firstSessionSig);
+  const allowPkpAddressToDecrypt = getPkpAccessControlCondition(pkpAddress);
+
+  const saltedPrivateKey = LIT_PREFIX + privateKey;
```
DashKash54, line 31: Nice naming

packages/wrapped-keys/src/lib/api/sign-message-with-encrypted-key.ts
```diff
@@ -0,0 +1,38 @@
...
+    litNetwork: litNodeClient.config.litNetwork,
+  });
+
+  const allowPkpAddressToDecrypt = getPkpAccessControlCondition(
+    storedKeyMetadata.pkpAddress
+  );
+
+  return signMessageWithLitAction({
+    ...params,
+    litActionIpfsCid: getLitActionCid(network, 'signMessage'),
```
DashKash54, line 33: Export `signMessage` to const?

packages/wrapped-keys/src/lib/api/sign-transaction-with-encrypted-key.ts
```diff
@@ -0,0 +1,38 @@
...
+    litNetwork: litNodeClient.config.litNetwork,
+  });
+
+  const allowPkpAddressToDecrypt = getPkpAccessControlCondition(
+    storedKeyMetadata.pkpAddress
+  );
+
+  return signTransactionWithLitAction({
+    ...params,
+    litActionIpfsCid: getLitActionCid(network, 'signTransaction'),
```
DashKash54, line 34: Export `signTransaction` to const?

packages/wrapped-keys/src/lib/api/store-encrypted-key-metadata.ts
```diff
@@ -0,0 +1,31 @@
+import { storePrivateKeyMetadata } from '../service-client';
+import { StoreEncryptedKeyMetadataParams } from '../types';
+import { getFirstSessionSig, getPkpAddressFromSessionSig } from '../utils';
+
+/** Get a previously encrypted and persisted private key and its metadata.
+ * Note that this method does _not_ decrypt the private key; only the _encrypted_ key and its metadata will be returned to the caller.
```
DashKash54, line 6: Should be stores instead of get

packages/wrapped-keys/src/lib/constants.ts
```diff
@@ -1,31 +1,2 @@
-const CHAIN_ETHEREUM = 'ethereum';
-const LIT_PREFIX = 'lit_';
-export const NETWORK_EVM = 'evm';
-export const NETWORK_SOLANA = 'solana';
```
DashKash54, line 4: Why has this been removed?

packages/wrapped-keys/src/lib/lit-actions-client/generate-key.ts
```diff
@@ -0,0 +1,42 @@
...
+    const result = await litNodeClient.executeJs({
+      sessionSigs: pkpSessionSigs,
+      ipfsId: litActionIpfsCid,
+      jsParams: {
+        pkpAddress,
+        accessControlConditions,
+      },
+    });
+
+    const response = postLitActionValidation(result);
```
DashKash54, line 35: Does `postLitActionValidation()` implementation looks good?

packages/wrapped-keys/src/lib/lit-actions-client/sign-transaction.ts
```diff
@@ -0,0 +1,53 @@
...
+  broadcast: boolean;
+}
+
+export async function signTransactionWithLitAction({
+  accessControlConditions,
+  broadcast,
+  litActionIpfsCid,
+  litNodeClient,
+  pkpSessionSigs,
+  storedKeyMetadata: { ciphertext, dataToEncryptHash, pkpAddress },
```
DashKash54, line 30: Why do it like this?

packages/wrapped-keys/src/lib/lit-actions-client/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Post processes the Lit Action result to ensure that the result is non-empty and a valid string
+ *
+ * @param result - The Lit Action result to be processes
+ *
+ * @returns { string } - The response field in the Lit Action result object
+ */
+export function postLitActionValidation(
+  result: ExecuteJsResponse | undefined
+): string {
+  // console.log(`Lit Action result: ${JSON.stringify(result)}`);
```
DashKash54, line 18: nit: Remove commented code. We should also ensure that we don't log decrypted private key when importing or exporting

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
DashKash54, line 1: The Ethereum Lit Action remain in their separate directory?

packages/wrapped-keys/src/lib/service-client/constants.ts
DashKash54, line 1: Do we provide both these info when we call the Lambda? How does it know which env var to use?

DashKash54 (commented):
Left a few comments

packages/types/src/lib/interfaces.ts
```diff
@@ -948,7 +948,7 @@ export interface DecryptZipFileWithMetadataProps extends SessionSigsOrAuthSig {
 export interface SessionKeySignedMessage {
   sessionKey: string;
   resources?: any[];
-  capabilities: string[];
+  capabilities: AuthSig[];
```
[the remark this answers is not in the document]
MaximusHaximus: Not sure what you mean by backwards compatible -- backwards compatible with what?  The value is definitely an `AuthSig` array, our local-tests all run using this code.

packages/types/src/lib/types.ts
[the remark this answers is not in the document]
MaximusHaximus: Nothing changed -- our ESLint rule configuration (and SDK-wide convention) enforces that we always use interfaces unless we are using features that require a type.  I ran `eslint --fix` on the file and it fixed the entries that should have been interfaces :). See first commit in the pr here: https://github.com/LIT-Protocol/js-sdk/pull/513/commits/21cfaf35bdf15f7add6ec2882fcad65963e15c72

packages/wrapped-keys/src/lib/api/generate-private-key.ts
MaximusHaximus: :+1: Updated this to derive the keyType based on the `network`

packages/wrapped-keys/src/lib/api/import-private-key.ts
MaximusHaximus: thx g

packages/wrapped-keys/src/lib/api/sign-transaction-with-encrypted-key.ts
MaximusHaximus: It's enforced by a string union type `LitActionType` here: https://github.com/LIT-Protocol/js-sdk/blob/c3dafd0a4eb4ccf7bb93848686bbd04a01882fd0/packages/wrapped-keys/src/lib/lit-actions-client/types.ts#L3

I defined a const to go with it, but discovered that TS complains about `Record<>` usage using a string const to reference a Record entry where the keys are defined from a type union.  I spent quite a bit of time trying to find a solution that worked for both, but in the end, I realised that since this is an entirely internal type and it's enforced by Typescript that it matches exact strings, making it a const is not really worth the effort; if you try to use any other string than is defined on `LitActionType`, TS won't compile and your IDE gives you auto-complete when you use this method.

packages/wrapped-keys/src/lib/api/sign-message-with-encrypted-key.ts
MaximusHaximus: See https://github.com/LIT-Protocol/js-sdk/pull/513#discussion_r1658783156

packages/wrapped-keys/src/lib/api/generate-private-key.ts
MaximusHaximus: See https://github.com/LIT-Protocol/js-sdk/pull/513#discussion_r1658783156

packages/wrapped-keys/src/lib/api/export-private-key.ts
MaximusHaximus: We use this file naming convention throughout the entire JS SDK; I named it the same way as the rest of the SDK names files.  Why should it be camel case? :)

packages/wrapped-keys/src/lib/api/export-private-key.ts
MaximusHaximus: So, we store the encrypted data at rest, but we can't actually see the data when the client posts it to know if it's salted.  However, we definitely don't want someone to be storing un-salted keys on our backend -- this seems like the _only place_ we could reasonably tell them that they have stored something that they shouldn't have on the backend.  If they want to _dangerously store un-salted keys on our backend_, IMO they should write their own _dangerously written `export-private-key` LIT action too_. They are already bypassing our SDK to get data into the backend without a salt on it, so there is no expectation that our 'stock' SDK logic will work at that point.  The LIT SDK internals should be opinionated about security IMO. What do you think?

packages/wrapped-keys/src/lib/constants.ts
MaximusHaximus: Same reason as https://github.com/LIT-Protocol/js-sdk/pull/513#discussion_r1658783156; consumers of the SDK get IDE auto-complete and TS compiler checking based on the string union type defined here: https://github.com/LIT-Protocol/js-sdk/blob/c3dafd0a4eb4ccf7bb93848686bbd04a01882fd0/packages/wrapped-keys/src/lib/types.ts#L10

packages/wrapped-keys/src/lib/lit-actions-client/generate-key.ts
MaximusHaximus: :+1: Yup, I think so - the errors I get from failures seem to be about as good as we can expect given how LIT actions throw.

packages/wrapped-keys/src/lib/lit-actions-client/sign-transaction.ts
MaximusHaximus: Can you elaborate on what you mean?

packages/wrapped-keys/src/lib/lit-actions-client/utils.ts
MaximusHaximus: :+1: Deleted the commented code, and removed the console.log left in the same method.
I audited `log()` usage in this package and found only 2 places where we log -- neither one logs privileged information.

References:
https://github.com/LIT-Protocol/js-sdk/blob/81409643d46c800834afe39d5cbc0048286345f8/packages/wrapped-keys/src/lib/utils.ts#L31
https://github.com/LIT-Protocol/js-sdk/blob/81409643d46c800834afe39d5cbc0048286345f8/packages/wrapped-keys/src/lib/utils.ts#L66

packages/wrapped-keys/src/lib/litActions/solana/src/signTransactionWithSolanaEncryptedKey.js
MaximusHaximus: Yep; I didn't change the structure of the litActions directory

packages/wrapped-keys/src/lib/service-client/constants.ts
MaximusHaximus: Sorry, your comment doesn't point at a specific line so I'm not entirely sure I understand what you mean.  The LIT actions aren't directly aware of our backend wrapped keys service, so I don't think they need to know about any of the constants in this file, right?

packages/wrapped-keys/src/lib/api/export-private-key.ts
MaximusHaximus: :+1: removed FIXME

packages/wrapped-keys/src/index.ts
MaximusHaximus: It's still tree-shakeable; bundlers will only include the values from the exported object that the consumer actually uses, and these are all specific functions, rather than classes with mixed dependencies.

I actually had considered adding a `createClient(litNodeClient, sessionSigs})` method to the module exports, which would eliminate those 2 arguments from _all of our API functions_, which would simplify all of our typescript types _and_ make our function signature more concise and easy to understand -- but decided not to because the return value of `createClient()` would _not_ have been tree-shakeable and would have always included all of the functions.

local-tests/tests/testEthereumBroadcastTransactionWrappedKey.ts
```diff
@@ -40,6 +37,8 @@ export const testEthereumBroadcastTransactionWrappedKey = async (
     pkpSessionSigs,
     privateKey: wrappedKeysWalletPrivateKey,
     litNodeClient: devEnv.litNodeClient,
+    publicKey: '0xdeadbeef',
```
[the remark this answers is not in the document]
MaximusHaximus: Yup -- it's not actually used for anything, and these are actually writing to Dynamodb and are only used for integration testing the API.  We don't validate the structure of the publicKey on the backend, so I set the integration tests to write this string so that we could safely clean out the DynamoDB backend of these records intermittently.

A better solution here would be to set up a CI instance of the backend wrapped keys API instead, so we can pave it over regularly and not spend $$$$$ on storing (eventually) thousands and millions of temporary keys that our CI and tests generate ;).

local-tests/tests/testSignTransactionWithSolanaEncryptedKey.ts
```diff
@@ -81,14 +80,14 @@ export const testSignTransactionWithSolanaEncryptedKey = async (
     })
     .toString('base64');
 
-  const unsignedTransaction: SolanaLitTransaction = {
+  const unsignedTransaction: SerializedTransaction = {
```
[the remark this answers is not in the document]
MaximusHaximus: Yep!

MaximusHaximus:
> Are there tests for `getEncryptedKeyMetadata` and `storeEncryptedKeyMetadata`?

Not directly; they are used by all of the other tests, so are extensively tested alreeady

packages/wrapped-keys/src/lib/api/store-encrypted-key-metadata.ts
MaximusHaximus: Good catch :+1: Fixed

local-tests/tests/testEthereumBroadcastTransactionGeneratedKey.ts
```diff
@@ -31,7 +28,7 @@ export const testEthereumBroadcastTransactionGeneratedKey = async (
 
   const { pkpAddress, generatedPublicKey } = await generatePrivateKey({
     pkpSessionSigs,
-    network: NETWORK_EVM,
+    network: 'evm',
```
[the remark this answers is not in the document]
MaximusHaximus: So once we started restricting `network` to be specific values using a string union type, I discovered that if someone used the `const` to call our method, the typescript checker would throw an error `TS2322: Type string is not assignable to type Network` -- because the type of the constant is `string`, decides that it is too generate -- it's not a _specific string literal_.

So then I tried defining these as:
```javascript
export const NETWORK_EVM: Network = 'evm';
export const NETWORK_SOLANA: Network = 'solana';
```
That _did_ fix the consumers code's typechecking, but then, _inside the SDK_ I started getting TS checker errors in the `lit-actions-client/constants.ts` file where we define the `LIT_ACTION_CID_REPOSITORY`:

`TS2739: Type
Readonly<{   [x: string]: 'QmdYUhPCCK5hpDWMK1NiDLNLG6RZQy61QE4J7dBm1Y2nbA' | 'QmSi9GL2weCFEP1SMAUw5PDpZRr436Zt3tLUNrSECPA5dT'; }>
is missing the following properties from type Readonly<Record<Network, string>>: evm, solana`

Even though the entry clearly has the correct keys in it, defined by their constants :(
```javascript
import { LitCidRepository } from './types';
import { NETWORK_EVM, NETWORK_SOLANA } from '../constants';

import { Network } from '../types';

export type LitActionType =
  | 'signTransaction'
  | 'signMessage'
  | 'generateEncryptedKey';

export type LitCidRepositoryEntry = Readonly<Record<Network, string>>;

export type LitCidRepository = Readonly<
  Record<LitActionType, LitCidRepositoryEntry>
>;

const LIT_ACTION_CID_REPOSITORY: LitCidRepository = {
  signTransaction: Object.freeze({
    [NETWORK_EVM]: 'QmdYUhPCCK5hpDWMK1NiDLNLG6RZQy61QE4J7dBm1Y2nbA',
    [NETWORK_SOLANA]: 'QmSi9GL2weCFEP1SMAUw5PDpZRr436Zt3tLUNrSECPA5dT',
  }),
  signMessage: Object.freeze({
    [NETWORK_EVM]: 'QmTMGcyp77NeppGaqF2DmE1F8GXTSxQYzXCrbE7hNudUWx',
    [NETWORK_SOLANA]: 'QmYQC6cd4EMvyB4XPkfEEAwNXJupRZWU5JsTCUrjey4ovp',
  }),
  generateEncryptedKey: Object.freeze({
    [NETWORK_EVM]: 'QmaoPMSqcze3NW3KSA75ecWSkcmWT1J7kVr8LyJPCKRvHd',
    [NETWORK_SOLANA]: 'QmdRBXYLYvcNHrChmsZ2jFDY8dA99CcSdqHo3p1ES3UThL',
  }),
};

export { LIT_ACTION_CID_REPOSITORY };
```

In the end, I discovered that I had to use _string literals_ in the LIT_ACTION_CID_REPOSITORY to make TS happy; I couldn't use our consts as keys.  Then I was faced with a conundrum; now external consumers would be using constants that we didn't use ourselves.  Since TS can't use runtime data as types, I went the opposite direction and exported the string union type. I did some digging around the net to try and identify a solution that would let me use the same consts that consumers used to define the keys on our repository, but couldn't find a solution and it was taking quite a lot of time at that point.

Given that these are user-facing arguments, maybe it'd be fine for our internal files to use string constants when defining actual records, but external consumers still have constants to use if they want to?  I've pushed a commit that makes that change -- take a look and LMK  what you think.  I also added exported constants for `keyType` using the same pattern.

MaximusHaximus:
Addressed feedback and responded to some questions with questions @DashKash54  :)

packages/wrapped-keys/src/index.ts
DashKash54: If both are tree-shakable then what's the reason behind `export const api = {..}` instead of directly exporting the individual functions? @MaximusHaximus

local-tests/tests/testEthereumBroadcastTransactionWrappedKey.ts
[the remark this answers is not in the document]
DashKash54: Interesting idea but for now just keeping the same publicKey seems easier and we can just periodically delete it from the DB

packages/types/src/lib/interfaces.ts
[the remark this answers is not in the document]
DashKash54: Should we also update the `SessionSigningTemplate` then? It uses `capabilities: any[]`. Maybe in a separate PR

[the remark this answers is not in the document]
DashKash54: Here's the ticket for it: https://linear.app/litprotocol/issue/LIT-3527/change-type-of-capabilities-to-authsig[]

packages/types/src/lib/types.ts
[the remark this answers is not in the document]
DashKash54: Then EsLint is not being run in the CI? @MaximusHaximus @Ansonhkg ?

packages/wrapped-keys/src/lib/lit-actions-client/sign-transaction.ts
DashKash54: Ig the idea is to club the DB data in a single var

DashKash54 (commented):
This looks good

DashKash54 (approved):
LGTM!

packages/types/src/lib/types.ts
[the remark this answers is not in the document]
MaximusHaximus: That's right, there is no mutative CI that changes files, and not everyone has their editors set to run `eslint --fix` on save, so these things happen :tm:.  It's been on the roadmap to implement for a few months.

PR #514, opened 2024-06-26
fix: if the user omits "uses" it gets default set to "1".  instead, i…

Ansonhkg:
if the user omits "uses" it gets default set to "1".  instead, it should just not be set at all (don't supply the uses key if the uses key isn't supplied).


joshLong145 (approved):
Approving but should have linter run before merge

glitch003 (approved):
beautiful, thank you sir!

PR #516, opened 2024-07-01
manually release private key in tests

glitch003:
We had a 10 second timeout on releasing a private key in tests.  This can cause issues if a test runs longer than 10s.  This PR is a first attempt at manually releasing keys after each test, so that it's not time based, but event based.


joshLong145:
This makes sense, I think this is a good place to iterate from. We might be able to make the release imlpicit. But this change moves us in a better direction with explicit rerelease of the key.

joshLong145 (approved):

Ansonhkg (approved):

PR #515, opened 2024-06-27
Feature/3404 datil dev support

Ansonhkg:
# Changelog

### 🚀 Features

- *(tinny)* Introduce chunked parallel execution
- *(tinny)* LIT_RPC enum
- *(tinny)*Add relayer tests
- *(sdk)* Add datil relayer support

### 🐛 Bug Fixes

- Import
- *(testing)* Correct test result reporting in runTestsParallel
- *(tinny)* Temp disabled wallet switching when minting pkp
- *(tinny)* Temp disabled `switchWallet` when minting capacity delegation authsig
- *(tinny)* Simplify and centralize RPC URL handling

### 📚 Documentation

- Update local-test README

### ⚙️ Miscellaneous Tasks

- Fix `testLIT` typo

### Dev

- *(test)* Add env file for local tests and dotenv
- *(test)* Add .env
- *(test)* Add lock file
- Add shiva types defs
- *(shiva)* Update shiva client

# Description

- Follow up from https://github.com/LIT-Protocol/js-sdk/pull/510

# Installation

```
yarn add @lit-protocol/lit-node-client@datil-dev
```
or

```
yarn add @lit-protocol/lit-node-client@6.1.0-beta.0
```

# Using the `datil-dev` network

```ts
const litNodeClient = new LitNodeClient({
  litNetwork: "datil-dev",
});

await litNodeClient.connect();
```

# Test command:

~~❗️❗️Note: I'm failing to run the tests in parallel right now, but they work when running individually. I added a `CHUNK_SIZE` feature which would run parallel execution in chunks and I thought that would fix it, but it seems like it might be RPC rate limit issue instead of nonce? not sure.. still to be investigated...❗️❗️~~

UPDATE: contract calls still being reverted, but if gasLimit is set to 5 mil, then parallel executions work fine

```
DEBUG=true NETWORK=datil-dev yarn test:local
```

# Related PRs:
- Relayer: https://github.com/LIT-Protocol/relay-server/pull/46
- General Worker: https://github.com/LIT-Protocol/lit-apps/pull/18
- DatilDev Faucet: https://github.com/LIT-Protocol/faucet-app/pull/10

PR #521, opened 2024-07-03
Feature/lit 3536 cleanup shiva changes

joshLong145:
# Description

Addresses comments from PR: #515 around Shiva Client changes.
- Fixes error handling and now `throws` over using `Promise.reject`
- Refactors polling to no longer double check request status
- Adds `ShivaError` as a convince method for transforming error results into an `Error` object type
- Swaps `dotenv` for nx dotenv loader

## Type of change

# How Has This Been Tested?

- Ran tests against shiva to assure compatibility

# Checklist:


local-tests/setup/shiva-client.ts
```diff
@@ -7,6 +7,19 @@ import {
   TestNetState,
 } from './shiva-client.d';
 
+class ShivaError extends Error {
+  constructor(shivaResponse: TestNetResponse<any>) {
+    let message = `An error occured on request to testnet with id: ${shivaResponse.testnetId}`;
```
Ansonhkg, line 12: typo
```suggestion
    let message = `An error occurred on request to testnet with id: ${shivaResponse.testnetId}`;
```

Ansonhkg (approved):
Nice & better error handling!

Ansonhkg (changes requested):
Need to fix CI build issue:

```
/opt/hostedtoolcache/node/20.15.0/x64/bin/node: .env: not found
```

joshLong145:
After re running e2e tests on other branches which previously passed it looks like our contract wrties are under what the current chain price is: https://github.com/LIT-Protocol/js-sdk/actions/runs/9751194755/job/27055335469?pr=516

So i do not think this pr is cause of failure. 
@Ansonhkg @MaximusHaximus

joshLong145:
@Ansonhkg @MaximusHaximus 
Can we merge this?

local-tests/setup/shiva-client.ts
```diff
@@ -152,10 +168,15 @@ export class TestnetClient {
       .then((res: Response) => {
         if (res.status === 200) {
           return res.json();
+        } else {
+          throw res; // throw as an error to the caller
```
MaximusHaximus, line 172: Can we make this throw an error with `res.text()` as it's message instead of throwing the entire Response object?

local-tests/setup/shiva-client.ts
```diff
@@ -220,21 +240,9 @@ export class ShivaClient {
...
-        pub epoch_length: i32,
-        pub custom_build_path: Option<String>,
-        pub lit_action_server_custom_build_path: Option<String>,
-        pub existing_config_path: Option<String>,
-        pub which: Option<String>,
-        pub ecdsa_round_timeout: Option<String>,
-        pub enable_rate_limiting: Option<String>,
-       }
+   * Used to start an instance of a lit network through the Lit Testnet Manager
+   * if an isntance exists, we will just take it as we optimistically assume it will not be shut down in the test life time.
```
MaximusHaximus, line 244: :nit: typo
```suggestion
   * if an instance exists, we will just take it as we optimistically assume it will not be shut down in the test life time.
```

local-tests/setup/shiva-client.ts
```diff
@@ -281,7 +289,11 @@ export class ShivaClient {
         }
       );
 
-      const createTestnet = await createTestnetResp.json();
+      const createTestnet =
+        (await createTestnetResp.json()) as TestNetResponse<void>;
+      if (createTestnetResp.status != 200) {
```
MaximusHaximus, line 294: I think this should go before trying to parse the result as json, so if we get an e.g. error `500` back with a string as its body, we don't throw a failure to parse the body as JSON before we check the response status?

MaximusHaximus (changes requested):
:+1: Nice work, this is looking good.  Just a few things I noticed added in-line

.env.sample
```diff
@@ -1,6 +1,6 @@
 #Tinny ENV Vars
 MAX_ATTEMTPS=1
```
Ansonhkg, line 2: typo

```suggestion
MAX_ATTEMPTS=1
```

Ansonhkg (changes requested):

local-tests/setup/shiva-client.ts
```diff
@@ -291,3 +270,23 @@ export class ShivaClient {
...
+
+async function _processTestnetResponse<T>(
+  response: Response
+): Promise<TestNetResponse<T>> {
+  let createTestnet: TestNetResponse<T>;
+  try {
+    createTestnet = (await response.json()) as TestNetResponse<T>;
+  } catch (err) {
+    let message = await response.text();
+    console.error('Error while creating testnet instance:', message);
```
MaximusHaximus, line 282: hmmm -- this isn't always used for creating a testnet instance anymore; could be due to a `stopTestnet` or a `transitionEpoch` call.  Since we're `throw`ing from this function in this catch() anyway, maybe just :hocho: the `console.error()` entirely?

local-tests/setup/shiva-client.ts
```diff
@@ -102,100 +110,82 @@ export class TestnetClient {
...
-    )
-      .then((res: Response) => {
-        return res.json();
-      })
-      .then((body: TestNetResponse<void>) => {
-        console.log('shutdown respone: ', body);
-      });
+    );
+
+    let testnetResp = _processTestnetResponse<boolean>(res);
```
MaximusHaximus, line 186: hmm -- stashing the promise returned by `_processTestnetResponse()` in a local variable here is a little unexpected; if you inspected it using a debugger on L188, it would not yet be resolved/rejected, and we don't `await` it here.  Maybe just `return _processTestnetResponse<bolean>(res);` here? Or `await` it if you want the local variable in scope for some purpose

local-tests/setup/shiva-client.ts
```diff
@@ -102,100 +110,82 @@ export class TestnetClient {
...
-          throw res;
-        }
-      })
-      .then((body: TestNetResponse<boolean>) => {
-        console.log('validator kick response: ', body);
-        return body;
-      });
+    );
+
+    let randomPeerStopRes = _processTestnetResponse<boolean>(res);
```
MaximusHaximus, line 172: Same remarks as the comment on `stopTestnet()` -- would be clearer to either `await` the result if you want to be able to inspect it on L172, or just `return` the function call.

local-tests/setup/shiva-client.ts
```diff
@@ -102,100 +110,82 @@ export class TestnetClient {
...
-        if (res.status === 200) {
-          return res.json();
-        }
-      })
-      .then((body: any) => {
-        console.log('Stopped random peer: ', body);
-      });
+    );
+
+    let transitionEpochAndWaitRes = _processTestnetResponse<boolean>(res);
```
MaximusHaximus, line 156: Same remarks as the comment on `stopTestnet()` -- would be clearer to either `await` the result if you want to be able to inspect it on L158, or just `return` the function call.

local-tests/setup/shiva-client.ts
```diff
@@ -102,100 +110,82 @@ export class TestnetClient {
...
-        return res.json();
-      })
-      .then((info: TestNetResponse<TestNetInfo>) => {
-        this._info = info.body;
-        this._currentState = info.lastStateObserved as TestNetState;
-        console.log('setting testnet info: ', this._info);
-      });
+    );
+
+    let testnetConfigRes = _processTestnetResponse<boolean>(res);
```
MaximusHaximus, line 140: Same remarks as the comment on `stopTestnet()` -- would be clearer to either `await` the result if you want to be able to inspect it on L140, or just `return` the function call.

local-tests/setup/shiva-client.ts
```diff
@@ -102,100 +110,82 @@ export class TestnetClient {
       const pollRes = await fetch(
         this._processEnvs.TESTNET_MANAGER_URL + '/test/poll/testnet/' + this._id
       );
-      const res = await pollRes.json();
+      const res: TestNetResponse<TestNetState> = await pollRes.json();
```
MaximusHaximus, line 113: Should this be above L113's attempt to parse the result as json? If we get an error 500 from a request, it probably won't be in JSON, no?  Maybe this should use `_processTestnetResponse()` also?

MaximusHaximus (changes requested):
Almost thrre I think -- I added a few things in in-line; several are just the same comment/question in different functions.

joshLong145:
@MaximusHaximus 
Thanks for the review! I had a pending commit that is now pushed that addresses the comments.

Ansonhkg (approved):
I think this is good to go!

PR #522, opened 2024-07-04
LIT-3688 - Handle more cases for talking to the lit nodes across epoch transitions

glitch003:
This PR essentially says:

if now < epoch start time + 30s, use `epoch number - 1`.  

if now >= epoch start time + 30s, then use `epoch number`

So, when an epoch transitions, we need to wait 30s before we start using it.  This is because, during that time, the nodes are in a transitionary state, and some may think they're on the next epoch, and some on the previous epoch.  By waiting 30s to send the new epoch, we give the nodes time to retrieve the new state.  Once 30s has elapsed, the nodes should all have seen and advanced to the new epoch, and it should be safe to send the new epoch.

So this PR removes the setTimeout that sets the new epoch number.  Instead, it looks at the epoch transition time from the chain, and starts to use the new epoch 30s after that has elapsed.  

This is better than the current system because it handles more cases.  

Suppose the user just connected via the SDK, and the epoch transitioned 1s ago.  If we use this new epoch number, and some nodes have not seen it yet, their signatures will fail.  This new strategy also works if the user is already connected via the SDK, and then the epoch transitions via the event listener.

### DIFF
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -860,7 +860,7 @@ export class LitCore {
     }
 
     try {
-      const epoch = this._stakingContract['epoch']();
+      const epoch = await this._stakingContract['epoch']();
 
       // when we transition to the new epoch, we don't store the start time.  but we
       // set the endTime to the current timestamp + epochLength.
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -15,6 +15,8 @@ import {
   validateUnifiedAccessControlConditionsSchema,
 } from '@lit-protocol/access-control-conditions';
 import {
+  HTTP,
+  HTTPS,
   LIT_CURVE,
   LIT_ENDPOINT,
   LIT_ERROR,
@@ -24,8 +26,6 @@ import {
   RPC_URL_BY_NETWORK,
   StakingStates,
   version,
-  HTTP,
-  HTTPS,
 } from '@lit-protocol/constants';
 import { LitContracts } from '@lit-protocol/contracts-sdk';
 import {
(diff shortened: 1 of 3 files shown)


GTC6244:
@glitch003  - the nodes themselves are on a 5s refresh timer - would suggest that the EPOCH_PROPAGATION_DELAY value could probably quite easily be 10s.

glitch003:
> @glitch003 - the nodes themselves are on a 5s refresh timer - would suggest that the EPOCH_PROPAGATION_DELAY value could probably quite easily be 10s.

ok let's do 15s just in case?

GTC6244:
> > @glitch003 - the nodes themselves are on a 5s refresh timer - would suggest that the EPOCH_PROPAGATION_DELAY value could probably quite easily be 10s.
> 
> ok let's do 15s just in case?

Sure, works for me!

Ansonhkg:
CI is failing - Putting a `console.log` in the `throwError` function allowed me to see this error:
```
'Unable to fetch current epoch number; no staking contract configured. Did you forget to `connect()`?'
```

cc @MaximusHaximus

packages/core/src/lib/lit-core.ts
```diff
@@ -117,15 +118,6 @@ const NETWORKS_REQUIRING_SEV: string[] = [
   LitNetwork.DatilTest,
 ];
 
-// The only network we consider entirely static, and thus ignore EPOCH changes for, is Cayenne
```
MaximusHaximus, line 120: Replaced with `CENTRALISATION_BY_NETWORK` since that is what we really needed here, and Anson added it earlier this week 🎉  💯

MaximusHaximus:
I updated this PR.
- Fixed the race condition between setting the staking contract and trying to use it to look up the current epoch data
- Added a try/catch to tinny's `init()` method that logs errors in a useful way
- Renamed some methods to make the setting of `epochCache` entirely private
- Separated mutating the `lit-core` instance object with new epoch data into a separate method from fetching the data from the chain for clarity and to allow us to do something like fetch the data but not mutate the current instance without refactoring the object when we need to do that.

MaximusHaximus (approved):
Per my comment, I refactored this some and I think this is now ready to land.

Ansonhkg (approved):
lgtm!

PR #523, opened 2024-07-05
Feature/lit 3549 js sdk add datiltest support

Ansonhkg:
# Description

Add support for `datil-test`

# npm tag

```
@datil-test
```

# Test

~~❗️ All tests passed except wrapped keys tests, we need to add `datil-test` support in the following services:~~
✅ (11 Jul 24) All tests related to these services now works!
```
  TestNetworks: 'https://test.wrapped.litprotocol.com/encrypted',
  Production: 'https://wrapped.litprotocol.com/encrypted',
```

```
NETWORK=datil-test yarn test:local
```


packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -986,6 +992,8 @@ export class LitContracts {
       'https://lit-general-worker.getlit.dev/habanero-contract-addresses';
     const DATIL_DEV_API =
       'https://lit-general-worker.getlit.dev/datil-dev/contracts';
+    const DATIL_TEST_API =
+      'https://staging.apis.getlit.dev/datil-test/contracts';
```
MaximusHaximus, line 996: Not necessarily a blocker, but I am curious why just this entry is using `staging.apis` when all the rest use general worker - do we need to ship another PR to update this URL at some point later, and if so, when do we know we can/should do that?

MaximusHaximus (approved):
I've shipped the update to wrapped-keys backend so it supports `datil-test` now 👍 

Looking at how 'give me an RPC URL for network X' logic is become more complex, and it spans across so many packages, I wonder if it would make sense to create a map of network names to RPC urls in `constants` and reference it there instead of using || statements checking specific network names -- basically the same pattern I used here: https://github.com/LIT-Protocol/js-sdk/blob/88cf00ea532b0c4d9193078cc51bc5bf0e4bc21f/packages/wrapped-keys/src/lib/service-client/constants.ts#L5-L17

WDYT?

I just added one question about the `staging.apis` domain vs. `lit-general-worker` -- otherwise this looks good to me.

packages/contracts-sdk/src/lib/contracts-sdk.ts
Ansonhkg: Yea, so the `staging.apis` is actually from a staging worker. I've been running it a few days to make sure that it doesn't break. I've now merged it to master https://github.com/LIT-Protocol/lit-apps/pull/22, so I will update this to use the general worker instead.

Ansonhkg:
> I've shipped the update to wrapped-keys backend so it supports `datil-test` now 👍
> 
> Looking at how 'give me an RPC URL for network X' logic is become more complex, and it spans across so many packages, I wonder if it would make sense to create a map of network names to RPC urls in `constants` and reference it there instead of using || statements checking specific network names -- basically the same pattern I used here:
> 
> https://github.com/LIT-Protocol/js-sdk/blob/88cf00ea532b0c4d9193078cc51bc5bf0e4bc21f/packages/wrapped-keys/src/lib/service-client/constants.ts#L5-L17
> 
> WDYT?
> 
> I just added one question about the `staging.apis` domain vs. `lit-general-worker` -- otherwise this looks good to me.

100%

Tracking this here https://linear.app/litprotocol/issue/LIT-3593/[js-sdk]-add-a-map-of-network-names-to-rpc-url

packages/lit-auth-client/src/lib/lit-auth-client.ts
```diff
@@ -146,7 +156,9 @@ export class LitAuthClient {
 
     // Set RPC URL
     this.rpcUrl =
-      options?.rpcUrl || this.litNodeClient.config.litNetwork === 'datil-dev'
+      options?.rpcUrl ||
+      this.litNodeClient.config.litNetwork === 'datil-dev' ||
+      this.litNodeClient.config.litNetwork === 'datil-test'
```
FedericoAmura, line 161: If `options.rpcUrl` has a value shouldn't `this.rpcUrl` utltimately be that value?

Screenshot from 2024-07-11 13-34-49

Same above [here](https://github.com/LIT-Protocol/js-sdk/pull/523/files#diff-2dcc2c94e7788b1bbaef5e6cf9158025a2580d238f807bd3cb923bbe355eb47bR619-R621)

FedericoAmura: I hope prettify doesn't do that! The code is fine, just that the operator resolution order does not match the developer intent

Anyway, the other PR is the way to go IMO so marking this as solved and approving 🚀

packages/constants/src/lib/constants/constants.ts
```diff
@@ -466,22 +466,50 @@ export const LIT_CHAINS: LITChain<LITEVMChain> = {
...
+  /**
+   * Chainlist entry for the Chronicle Testnet.
+   * https://chainlist.org/chain/175177
+   */
   chronicleTestnet: {
     contractAddress: null,
     chainId: 175177,
     name: 'Chronicle - Lit Protocol Testnet',
-    symbol: 'testLPX',
+    symbol: 'tstLPX',
```
MaximusHaximus, line 478: Looking at https://chainlist.org/?chain=2311&search=chronicle&testnets=true , I think this is backwards -- it shows vesuvius as `tstLPX` and our existing Chronicle as `tstLIT` ?

MaximusHaximus (changes requested):
I think our `symbol` is backwards from what I see on chainlist -- otherwise this is GTG.  Nice work on the constant maps <3 :)

packages/constants/src/lib/constants/constants.ts
Ansonhkg: Ohh nice catch! Fixed!

packages/misc/src/lib/misc.ts
```diff
@@ -666,25 +665,20 @@ export const defaultMintClaimCallback: MintCallback<
...
-      case LitNetwork.Manzano:
-        relayUrl = RELAY_URL_MANZANO + 'auth/claim';
-        break;
-      case LitNetwork.DatilDev:
-        relayUrl = RELAY_URL_DATIL_DEV + '/auth/claim';
-        break;
+    const AUTH_CLAIM_PATH = '/auth/claim';
+
+    const relayUrl: string =
+      RELAYER_URL_BY_NETWORK[network as LIT_NETWORK_VALUES];
```
MaximusHaximus, line 671: 1. Do we actually need to accept a string here instead of restricting the function to a specific string union of network keys (maybe `LIT_NETWORK_VALUES`?)?
3. If so, rather than coerce a possibly-invalid value to a type it may not be, then throw an error farther down based on a side-effect of that being invalid, it would be nice to get an explicit error back as a user of this API _and_ get type safety without a hard coercion on the data.  See this pattern here for the type-safe way to handle this using an assertion runtime function that throws a specific error if the data provided can't be coerced to that type.

https://github.com/LIT-Protocol/js-sdk/blob/304dca5a739d38b5a73a3c92715ed4878e6b04c4/packages/wrapped-keys/src/lib/service-client/utils.ts#L42

4. I think this code will no longer use the `params.relayUrl` when it is provided by the caller, because the ternary was removed on L687 (L679 in new code) -- I think for `custom` networks this should still be respected?

packages/constants/src/lib/constants/constants.ts
```diff
@@ -492,9 +520,12 @@ export const LIT_CHAINS: LITChain<LITEVMChain> = {
...
```
MaximusHaximus, line 523: I think this one is still off -- 
image

```suggestion
    symbol: 'tstLIT',
```

packages/constants/src/lib/constants/constants.ts
MaximusHaximus, line 478: I think this one is still off?
image
```suggestion
    symbol: 'tstLIT',
```

MaximusHaximus (changes requested):
Almost there I think -- tstLPX/tstLIT is a bit of a mindbender.  Just one other comment re: `relayerUrl` handling & types

local-tests/tests/wrapped-keys/util.ts
```diff
@@ -17,6 +17,8 @@ export function getChainForNetwork(network: LIT_NETWORKS_KEYS): {
       };
     case 'datil-dev':
       return { chain: 'vesuvius', chainId: LIT_CHAINS['datilDevnet'].chainId };
+    case 'datil-test':
+      return { chain: 'vesuvius', chainId: LIT_CHAINS['datilTestnet'].chainId };
```
MaximusHaximus, line 21: Should both the `datil-dev` and `datil-test` case now use `chronicleVesuviusTestnet`?  I don't think there is an entry for `datilTestnet` in `LIT_CHAINS`
```suggestion
      return { chain: 'vesuvius', chainId: LIT_CHAINS['chronicleVesuviusTestnet'].chainId };
    case 'datil-test':
      return { chain: 'vesuvius', chainId: LIT_CHAINS['chronicleVesuviusTestnet'].chainId };
```

packages/core/src/lib/lit-core.ts
```diff
@@ -110,6 +110,7 @@ const BLOCKHASH_SYNC_INTERVAL = 30_000;
 const NETWORKS_REQUIRING_SEV: string[] = [
   LitNetwork.Habanero,
   LitNetwork.Manzano,
+  LitNetwork.DatilTest,
 ];
 
 // The only network we consider entirely static, and thus ignore EPOCH changes for, is Cayenne
```
MaximusHaximus, line 116: Needs `LitNetwork.DatilTest` here also
Thinking about it, maybe it would make sencse to invert this logic entirel -- do we expect _every_ network going forwards to have epoch changes so that we don't need to update this moving forward.  WDYT?

packages/lit-auth-client/src/lib/lit-auth-client.ts
```diff
@@ -97,7 +94,13 @@ export class LitAuthClient {
         );
       }
 
-      const supportedNetworks = ['cayenne', 'habanero', 'manzano', 'datil-dev'];
+      const supportedNetworks = [
```
MaximusHaximus, line 97: :nit: Can we replace this check and `throw()` with `isSupportedLitNetwork()` from `misc` rather than adding to this array?

MaximusHaximus (changes requested):
2 last changes I think, and a nit.  We're almost there :D

packages/core/src/lib/lit-core.ts
Ansonhkg: `LitNetwork.DatilTest` is already in the array?

MaximusHaximus: I mean in `NETWORKS_WITH_EPOCH_CHANGES`, L117 - L122 -- it has `DatilDev` but not `DatilTest`

Ansonhkg: ahhh! Gotcha! [bb03270](https://github.com/LIT-Protocol/js-sdk/pull/523/commits/bb03270dff09e0ed06e24e0c9b6dcb34d1d04e55)

MaximusHaximus (approved):
:+1: LGTM, I think this is GTG 💖
Nice work <3

joshLong145 (approved):

PR #495, opened 2024-06-04
fix(tslib): standarised to use `private _` over `#` to prevent:

Ansonhkg:
# Description

Replace private methods `#methodFoo` to `private _methodFoo` 

This is a fix for this:
- https://github.com/LIT-Protocol/js-sdk/pull/492

Related PRs:
- https://github.com/microsoft/TypeScript/issues/36841

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Published `6.0.0-testing.7` to test on this repo:
https://github.com/LIT-Protocol/developer-guides-code/tree/master/custom-auth/browser

PR #526, opened 2024-07-10
fix(tslib): standarised to use private _ over # to prevent:

Ansonhkg:
Replacement for this:
https://github.com/LIT-Protocol/js-sdk/pull/495


yarn.lock
```diff
@@ -2,6 +2,11 @@
 # yarn lockfile v1
 
 
+"@aashutoshrathi/word-wrap@^1.2.3":
```
joshLong145, line 5: Is there a reason we need to update the lock with this pr? I dont see any changes to our `package.json` so just curious.

FedericoAmura (approved):
There is one more # call in [DiscordProvider](https://github.com/LIT-Protocol/js-sdk/blob/master/packages/lit-auth-client/src/lib/providers/DiscordProvider.ts#L210) 

Then our sdk would be #-calls free =D

joshLong145 (commented):
One small comment but looks good

yarn.lock
Ansonhkg: `package.json` file is updated in the `master` branch but the `yarn.lock` is old, and i `checkout -b` from `master` that's why `yarn.lock` is added here.

jtary (approved):
LGTM

PR #527, opened 2024-07-11
Feature/lit 3593 js sdk add a map of network names to rpc url

Ansonhkg:
# Description

This PR introduces a mapping of network names to their respective RPC URLs, relayer URLs, general worker URLs, metamask chain infos. It consolidates and organizes the URL endpoints to improve code maintainability and readability.

Context: https://github.com/LIT-Protocol/js-sdk/pull/523#pullrequestreview-2170217103

### DIFF
--- a/packages/lit-auth-client/src/lib/lit-auth-client.ts
+++ b/packages/lit-auth-client/src/lib/lit-auth-client.ts
@@ -2,22 +2,22 @@ import { ethers } from 'ethers';
 
 import {
   AuthMethodType,
-  LIT_RPC,
   ProviderType,
   RELAYER_URL_BY_NETWORK,
+  RPC_URL_BY_NETWORK,
 } from '@lit-protocol/constants';
 import { LitNodeClient } from '@lit-protocol/lit-node-client';
-import { bootstrapLogManager, getLoggerbyId, log } from '@lit-protocol/misc';
+import { bootstrapLogManager, log } from '@lit-protocol/misc';
 import {
+  AuthMethod,
   EthWalletProviderOptions,
   IRelay,
   LitAuthClientOptions,
-  OAuthProviderOptions,
-  StytchOtpProviderOptions,
-  ProviderOptions,
-  WebAuthnProviderOptions,
-  AuthMethod,
   MintRequestBody,
+  OAuthProviderOptions,
+  ProviderOptions,
+  StytchOtpProviderOptions,
+  WebAuthnProviderOptions,
 } from '@lit-protocol/types';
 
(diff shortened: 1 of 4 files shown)


packages/lit-auth-client/src/lib/lit-auth-client.ts
```diff
@@ -157,10 +135,7 @@ export class LitAuthClient {
     // Set RPC URL
     this.rpcUrl =
       options?.rpcUrl ||
-      this.litNodeClient.config.litNetwork === 'datil-dev' ||
-      this.litNodeClient.config.litNetwork === 'datil-test'
-        ? LIT_RPC.VESUVIUS
-        : LIT_RPC.CHRONICLE;
+      RPC_URL_BY_NETWORK[this.litNodeClient.config.litNetwork];
```
joshLong145, line 138: nit: though we have type safety there is a chance a user passes an invalid `network` param and this look up will fail. We might want to add a check after lookup to assure `this.rpcUrl` is defined.

packages/misc/src/lib/misc.ts
```diff
@@ -667,25 +665,15 @@ export const defaultMintClaimCallback: MintCallback<
...
-        relayUrl = RELAY_URL_DATIL_DEV + '/auth/claim';
-        break;
-        break;
-      case LitNetwork.DatilTest:
-        relayUrl = RELAY_URL_DATIL_TEST + '/auth/claim';
-        break;
+    const AUTH_CLAIM_PATH = '/auth/claim';
+
+    const relayUrl: string =
+      RELAYER_URL_BY_NETWORK[network as LIT_NETWORK_VALUES] + AUTH_CLAIM_PATH;
```
joshLong145, line 671: Same as above

joshLong145 (approved):

packages/misc/src/lib/misc.ts
```diff
@@ -667,25 +665,15 @@ export const defaultMintClaimCallback: MintCallback<
...
+
+    if (!params.relayUrl && !relayUrl) {
```
FedericoAmura, line 673: This will always be `false` because of `AUTH_CLAIM_PATH`

Screenshot from 2024-07-11 15-52-13

However, I think that the best error we can return here is that the URL is broken. Letting this pass and failing at the request, which will also catch the cases where user passes an invalid `params.relayUrl`

Ansonhkg: Nice catch! You mean always return `true` because `relayUrl` is always a valid string right?

FedericoAmura (approved):

PR #496, opened 2024-06-05
Feature/lit 2591 js sdk remove all redundantunused build processes & refactor: delete `cypress` directory

Ansonhkg:
# Description

Delete the `cypress` directory - This is to fix the merge conflict for this PR:
https://github.com/LIT-Protocol/js-sdk/pull/485

PR #529, opened 2024-07-11
fix: match version before building

Ansonhkg:
# Description

This PR added back `yarn tools --match-versions` command before building the packages. It was removed accidentally from this PR https://github.com/LIT-Protocol/js-sdk/pull/496

cc @MaximusHaximus when you're ready for individual versioning, we can remove this command again


joshLong145 (approved):
Good catch!

MaximusHaximus (approved):
:+1: rock on man

PR #533, opened 2024-07-12
feat: add `TEST_TIMEOUT` for tests

Ansonhkg:
# Description

This PR ensures that a test will time out if it ever gets stuck, allowing the test runner to continue running and produce a test report.

Note: We are in the process of migrating to Jest, but this fix is required to ensure smooth testing in the meantime.


local-tests/setup/tinny-utils.ts
```diff
@@ -45,3 +45,18 @@ export function randomSolanaPrivateKey() {
...
+
+/**
+ * Wraps a promise with a timeout.
+ * If the promise does not resolve or reject within the specified time, it will be rejected with a "Timed out" error.
+ *
+ * @param promise - The promise to wrap with a timeout.
+ * @param ms - The timeout duration in milliseconds.
+ * @returns A new promise that resolves or rejects based on the original promise or the timeout.
+ */
+export function withTimeout<T>(promise: Promise<T>, ms: number): Promise<T> {
```
joshLong145, line 57: I dont think the return type is `Promise<T>` Since the race is not guaranteed to resolve `promise` and therefore might be `void` if the timeout triggers.

joshLong145 (commented):
Left A comment on return type. But otherwise looks good

local-tests/setup/tinny-utils.ts
Ansonhkg: the `Promise.race` method returns a promise that resolves or rejects as soon as one of the promises in the itterable resolves or rejects, means that the return type will always be `Promise<T>` if the input promise resolves or `Promise<void>` if the timeout triggers, but both are covered under the type `Promise<T>`.

joshLong145: Im not sure that 'T' covers the case where the timeout promise resolves first and the value returned is void and `T` is not guaranteed to include `void` thinking the return type should look something like 
```js
withTimeout<T>(promise: Promise<T>, ms: number): Promise<T | void>
```
as to denote that the promise returned may resolve the given type specified by the promise provided, or return `void` meaning the timeout promise resolved first. Perhaps I am missing something.

local-tests/setup/tinny-utils.ts
```diff
@@ -45,3 +45,21 @@ export function randomSolanaPrivateKey() {
...
+ * If the promise does not resolve or reject within the specified time, it will be rejected with a "Timed out" error.
+ *
+ * @param promise - The promise to wrap with a timeout.
+ * @param ms - The timeout duration in milliseconds.
+ * @returns A new promise that resolves or rejects based on the original promise or the timeout.
+ */
+export function withTimeout<T>(
+  promise: Promise<T>,
+  ms: number
+): Promise<T | void> {
```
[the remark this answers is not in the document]
joshLong145: No i see what you mean, since it rejects it wont effect the return value of the `Promise` as the `timeout` promise will not return a value hence `never` but may `throw` and exception. I assumed `never` was assignable to `void` but it is actually a distinct type which is the omission of any return value. 

Going to resolve the previous comment. Thanks for the breakdown.

.env.sample
```diff
@@ -11,6 +11,7 @@ PRIVATE_KEYS="0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d
 NO_SETUP=false
 USE_SHIVA=false
 NETWORK_CONFIG=./networkContext.json
+TEST_TIMEOUT=30000
```
MaximusHaximus, line 14: Can we make this longer than the timeout time on the nodes just to give them a chance to actually time out server-side before we give up?  How about 45 seconds to give a little bit of time for setup costs plus 30s for the nodes to time out?  If the reason why it's taking >30s for the test to finish, it would be better if we got an error _from the nodes timing out on the lit action_ than a generic 'local' test timeout from our test suite.

MaximusHaximus (changes requested):
Just one thing -- would like the default timeout to be a bit longer than the timeout for running a LIT action on the nodes.

.env.sample
Ansonhkg: Yes sir! 🫡

MaximusHaximus (approved):
:+1: Rock on man, nice work 💖

PR #534, opened 2024-07-14
Feat: Update Wrapped keys Export to use Lit Actions

DashKash54:
# Description

Currently Wrapped Keys export feature simply decrypts the private key on the client. This PR updates the export to happen via a Lit Action and returns the decrypted key from inside the Lit Action. Exporting through a Lit Action also allows us to add more restrictions on the export feature like allow decryption only inside certain permitted Lit Actions which is not currently the case.

Also fixed the following:
1. `signMessageWithEthereumEncryptedKey` Lit Action was **not** initializing the Ethers wallet with the decrypted private key. This was accidentally left out in this commit: https://github.com/LIT-Protocol/js-sdk/pull/515/commits/2fbd86136198fc194133f805ccdd51feb4345a6f#diff-cfa4907cf70adeff456099de1858fb36fb1d13060b43e6357f6c932c679f815fL41
2. All the Lit Actions are using `removeSaltFromDecryptedKey()` which throws an error if the private key isn't prefixed with "lit_". But when invoked inside a Lit Action we don't support it with a try/catch block which causes the Lit Action to crash and  doesn't show the error to the user. Updated all the Lit Actions to show this error to the client.
    1. Published all the Lit Actions with the above fix and updated the Ipfs Cids in the `constants` file

## Type of change

# How Has This Been Tested?

It has been tested with three Tinny tests

- [X] `testExportWrappedKey`: Import an existing key, export it and compares the private keys
- [X] `testGenerateSolanaWrappedKey`: Generate a Solana key, export it and compare the derived publicKeys
- [X] `testGenerateEthereumWrappedKey`: Generate a Solana key, export it and compare the derived publicKeys

# Checklist:

### DIFF
--- a/local-tests/tests/wrapped-keys/testExportWrappedKey.ts
+++ b/local-tests/tests/wrapped-keys/testExportWrappedKey.ts
@@ -52,6 +52,7 @@ export const testExportWrappedKey = async (devEnv: TinnyEnvironment) => {
   const { decryptedPrivateKey } = await exportPrivateKey({
     pkpSessionSigs: pkpSessionSigsExport,
     litNodeClient: devEnv.litNodeClient,
+    network: 'solana',
   });
 
   if (decryptedPrivateKey !== privateKey) {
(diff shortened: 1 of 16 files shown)


MaximusHaximus (approved):
:+1: Nice work, this looks good to me <3

Ansonhkg (approved):

PR #537, opened 2024-07-16
fix: LIT-3607 - Use StaticJsonRpcProvider to improve performance of RPC requests

MaximusHaximus:
Reference issue - https://github.com/ethers-io/ethers.js/issues/901 Cuts our RPC request time costs approximately in half by eliminating a sequential call to get `chainId` from every(!) contract call we make

# Description

Replaced ethers JSONRPCProvider instances with StaticJSONRPCProvider
This eliminates unnecessary RPC calls to get `chainId` every time we use one of our contract methods.

## Type of change

# How Has This Been Tested?

- [X] Ran localtests against datil-dev, datil-test and cayenne

# Checklist:

### DIFF
--- a/packages/contracts-sdk/src/lib/addresses.ts
+++ b/packages/contracts-sdk/src/lib/addresses.ts
@@ -1,12 +1,12 @@
+import { rawSecp256k1PubkeyToRawAddress } from '@cosmjs/amino';
+import { Secp256k1 } from '@cosmjs/crypto';
+import { toBech32 } from '@cosmjs/encoding';
 import * as bitcoinjs from 'bitcoinjs-lib';
 import { Contract, ethers } from 'ethers';
 import { computeAddress } from 'ethers/lib/utils';
-import { PKPNFTData } from '../abis/PKPNFT.sol/PKPNFTData';
-import { toBech32 } from '@cosmjs/encoding';
-import { Secp256k1 } from '@cosmjs/crypto';
 
-import { rawSecp256k1PubkeyToRawAddress } from '@cosmjs/amino';
-export type TokenInfo = {
+import { PKPNFTData } from '../abis/PKPNFT.sol/PKPNFTData';
+export interface TokenInfo {
   tokenId: string;
   publicKey: string;
   publicKeyBuffer: Buffer;
@@ -14,7 +14,7 @@ export type TokenInfo = {
   btcAddress: string;
   cosmosAddress: string;
   isNewPKP: boolean;
-};
+}
 
 export const derivedAddresses = async ({
   publicKey,
@@ -64,7 +64,9 @@ export const derivedAddresses = async ({
         if (cachedPkpJSON[pkpTokenId]) {
           publicKey = cachedPkpJSON[pkpTokenId];
         } else {
-          const provider = new ethers.providers.JsonRpcProvider(defaultRPCUrl);
+          const provider = new ethers.providers.StaticJsonRpcProvider(
+            defaultRPCUrl
+          );
 
           const contract = new Contract(
             pkpContractAddress,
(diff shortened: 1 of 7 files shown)


packages/contracts-sdk/src/lib/addresses.ts
```diff
@@ -1,20 +1,20 @@
...
 import * as bitcoinjs from 'bitcoinjs-lib';
 import { Contract, ethers } from 'ethers';
 import { computeAddress } from 'ethers/lib/utils';
-import { PKPNFTData } from '../abis/PKPNFT.sol/PKPNFTData';
-import { toBech32 } from '@cosmjs/encoding';
-import { Secp256k1 } from '@cosmjs/crypto';
 
-import { rawSecp256k1PubkeyToRawAddress } from '@cosmjs/amino';
-export type TokenInfo = {
+import { PKPNFTData } from '../abis/PKPNFT.sol/PKPNFTData';
```
MaximusHaximus, line 8: eslint --fix

packages/contracts-sdk/src/lib/contracts-sdk.spec.ts
```diff
@@ -1,10 +1,7 @@
 import { LitContracts } from './contracts-sdk';
-import { ethers } from 'ethers';
-import { LitContract } from '../../../types/src/lib/types';
 
 describe('contractsSdk', () => {
   let litContracts: LitContracts;
```
MaximusHaximus, line 4: Went to update this reference to static rpc provider, but discovered it wasn't actually being used in this test file at all. Removed it

packages/core/src/lib/lit-core.ts
```diff
@@ -488,15 +484,14 @@ export class LitCore {
     if (!this.config.contractContext) {
       this.config.contractContext = await LitContracts.getContractAddresses(
         this.config.litNetwork,
-        new ethers.providers.JsonRpcProvider(
-          this.config.rpcUrl || LIT_CHAINS['lit'].rpcUrls[0]
+        new ethers.providers.StaticJsonRpcProvider(
+          this.config.rpcUrl || RPC_URL_BY_NETWORK[this.config.litNetwork]
```
MaximusHaximus, line 488: Fixed a bug here; we were always using `LIT_CHAINS['lit']` here.  It just so happened that due to internal override logic inside of the contracts SDK this actually worked.

packages/pkp-ethers/src/lib/pkp-ethers.ts
```diff
@@ -88,16 +87,12 @@ export class PKPEthersWallet
       prop.rpc || RPC_URL_BY_NETWORK[prop.litNodeClient.config.litNetwork];
 
     if (!rpcUrl) {
-      throw new Error('No RPC URL provided');
```
MaximusHaximus, line 91: Removed some duplicate code here that probably ended up here due to merge conflict shenanigans

Ansonhkg (approved):
This LGTM! Should we merge this https://github.com/LIT-Protocol/js-sdk/pull/538 into this? WDYT.?

packages/core/src/lib/lit-core.ts
Ansonhkg: Should we do this in the constructor instead? so that `this.config.rpcUrl` is always defined

FedericoAmura:
Just tested this with the pkp mint demo and not only this saves the time on every contract call, it was also making that request in a loop every a few seconds. There is a massive bandwidth and rpc load save as every client would be making those constantly

Huge improvement! 🔥

FedericoAmura (approved):

packages/core/src/lib/lit-core.ts
Ansonhkg: cc @MaximusHaximus  ^^

PR #538, opened 2024-07-16
fix: Workaround for ethers.js bug in webpack environments

joshdchang:
# Description

I was getting the following issue when trying to connect in a nodejs server-side webpack environment (in Next.js):
```
Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.7.2)
```

After doing some searching and testing, I figured out it was an bug from ethers.js with webpack. I tried the workaround from [https://github.com/ethers-io/ethers.js/issues/3536#issuecomment-1323302238](https://github.com/ethers-io/ethers.js/issues/3536#issuecomment-1323302238) and it fixed the issue. This PR only changes one line of code.

## Type of change

# How Has This Been Tested?

- [x] Try running it in Next.js server side with and without this change. You will see the difference.

# Checklist:

### DIFF
--- a/packages/contracts-sdk/src/lib/contracts-sdk.ts
+++ b/packages/contracts-sdk/src/lib/contracts-sdk.ts
@@ -584,8 +584,8 @@ export class LitContracts {
     if (context && 'provider' in context!) {
       provider = context.provider;
     } else {
-      provider = new ethers.providers.JsonRpcProvider({
-        rpcUrl,
+      provider = new ethers.providers.StaticJsonRpcProvider({
+        url: rpcUrl,
         skipFetchSetup: true,
       });
     }
--- a/packages/contracts-sdk/src/lib/contracts-sdk.ts
+++ b/packages/contracts-sdk/src/lib/contracts-sdk.ts
@@ -584,7 +584,10 @@ export class LitContracts {
     if (context && 'provider' in context!) {
       provider = context.provider;
     } else {
-      provider = new ethers.providers.JsonRpcProvider(rpcUrl);
+      provider = new ethers.providers.JsonRpcProvider({
+        rpcUrl,
+        skipFetchSetup: true,
+      });
     }
 
     if (network === 'datil-dev' || network === 'datil-test') {


Ansonhkg (commented):
Thanks for digging into this! 

Could you change the `JsonRpcProvider` to `StaticJsonRpcProvider`? 

We have just got this PR up (https://github.com/LIT-Protocol/js-sdk/pull/537) which could half to RPC request time by simply changing it to static

joshdchang:
> Could you change the `JsonRpcProvider` to `StaticJsonRpcProvider`?
> 
> We have just got this PR up (#537) which could half to RPC request time by simply changing it to static

Sure let me try

joshdchang:
@Ansonhkg Okay I changed it to `StaticJsonRpcProvider` and it seems to work well.

Ansonhkg (approved):
This LGTM. @MaximusHaximus WDYT?

MaximusHaximus (changes requested):
Thanks so much for the contribution @joshdchang :D We really appreciate it.

FYI, this change is a little bit different than you described @Ansonhkg -- this is literally controlling how `fetch()` is called rather than controlling whether `chainId` is checked (which was the difference between `JSONRPCProvider` and `StaticJSONRPCProvider`)

From the docs:

### **connection.skipFetchSetup ⇒ boolean**
_Normally a connection will specify the default values for a connection such as CORS-behavior and caching policy, to ensure compatibility across platforms. On some services, such as Cloudflare Workers, specifying any value (inclluding the default values) will cause failure. Setting this to true will prevent any values being passed to the underlying API.```_

I think it should be fine to add this argument to the `JSONRPCProvider` construction, but we create these providers in a bunch of different places, so while this PR as currently written would fix using the contracts SDK _directly_, consumers of the SDK will still run into this error in other places in the SDK where we construct the `JSONRPCProvider` on their behalf - likely just a few minutes later down their exploration of the JS SDK. 

For your reference, every place that we need the new argument added are the same exact lines that were changed from `JSONRPCProvider` to `StaticJSONRPCProvider` in #537, plus one additional place that I didn't change in that PR because we couldn't use the StaticJSONRPCProvider for generic chains where we wouldn't necessarily be able to skip the `chainId` pre-check behaviour [here](https://github.com/LIT-Protocol/js-sdk/blob/89b24c889b0b14830684fe581095ca9d8455595a/packages/misc/src/lib/misc.ts#L630)

Are you OK to make that change @joshdchang?

joshdchang:
> Are you OK to make that change @joshdchang?

Hi, thanks for this! Do you want me to change it to StaticJSONRPCProvider in all the locations as well? This would overlap substantially with #537 and might lead to merge conflicts. Maybe this is best done in one PR?

MaximusHaximus:
> > Are you OK to make that change @joshdchang?
> 
> Hi, thanks for this! Do you want me to change it to StaticJSONRPCProvider in all the locations as well? This would overlap substantially with #537 and might lead to merge conflicts. Maybe this is best done in one PR?

Sorry for the delayed response, and thanks so much for contributing!
We've landed the JsonStaticRPCProvider PR now, so you're GTG without needing to worry about conflicts :)

FedericoAmura:
Replacing this with https://github.com/LIT-Protocol/js-sdk/pull/646 which has the changes needed on all the providers. Thanks @joshdchang !!!

PR #540, opened 2024-07-17
fix: throw error with stack

Ansonhkg:
# Description

This will show error stack when running tinny/local-tests

```
NETWORK=datil-test yarn test:local  --filter=testUseEoaSessionSigsToPkpSign
```

## Before
image

## After

image

### DIFF
--- a/packages/misc/src/lib/misc.ts
+++ b/packages/misc/src/lib/misc.ts
@@ -162,6 +162,7 @@ export const throwErrorV1 = ({
   errorCode,
   requestId,
 }: NodeClientErrorV1): never => {
+  // TODO: to refactor such that we do not need a wrapper function to construct the error, since it gets added to all stack traces it does technically pollute.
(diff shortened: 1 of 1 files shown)


packages/misc/src/lib/misc.ts
```diff
@@ -162,30 +162,15 @@ export const throwErrorV1 = ({
...
-  };
-
-  throw new (errConstructorFunc as any)(
-    errorKind,
-    status,
-    details,
-    message,
-    errorCode,
-    requestId
+  throw new Error(
```
joshLong145, line 165: Can we add a TODO here to refactor such that we do not need a wrapper function to construct the error? Snice it gets added to all stack traces it does technically pollute.

joshLong145 (dismissed):

Ansonhkg:
Will do a proper change in v7.0, but we can use this code locally to debug

PR #543, opened 2024-07-17
feat: use short names

Ansonhkg:
# Description

Decision was made to use short names in this PR https://github.com/LIT-Protocol/lit-assets/pull/1328#discussion_r1680840672

### DIFF
--- a/packages/constants/src/lib/constants/constants.ts
+++ b/packages/constants/src/lib/constants/constants.ts
@@ -26,6 +26,45 @@ export const LIT_AUTH_SIG_CHAIN_KEYS: string[] = [
 export const AUTH_SIGNATURE_BODY =
   'I am creating an account to use Lit Protocol at {{timestamp}}';
 
+const oldChronicleChain = {
+  contractAddress: null,
+  chainId: 175177,
+  name: 'Chronicle - Lit Protocol Testnet',
+  symbol: 'tstLIT',
+  decimals: 18,
+  rpcUrls: [
+    'https://lit-protocol.calderachain.xyz/replica-http',
+    'https://chain-rpc.litprotocol.com/http',
+  ],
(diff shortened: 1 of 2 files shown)


packages/constants/src/lib/constants/constants.ts
```diff
@@ -471,83 +510,39 @@ export const LIT_CHAINS: LITChain<LITEVMChain> = {
...
-    blockExplorerUrls: ['https://vesuvius-explorer.litprotocol.com/'],
-    type: null,
-    vmType: 'EVM',
-  },
+  vesuvius: vesuviusChain,
+
+  /**
+   * @deprecated Will be removed in version 7.x. - Use `yellowstone` instead.
+   */
+  chronicleYellowstoneTestnet: yellowstoneChain,
```
MaximusHaximus, line 535: May as well :hocho: this entire entry since it was never merged into the shipped SDK, no?

packages/constants/src/lib/constants/constants.ts
```diff
@@ -688,32 +683,31 @@ export const metamaskChainInfo = {
...
-    blockExplorerUrls: LIT_CHAINS['chronicleVesuviusTestnet'].blockExplorerUrls,
+    rpcUrls: LIT_CHAINS['vesuvius'].rpcUrls,
+    blockExplorerUrls: LIT_CHAINS['vesuvius'].blockExplorerUrls,
     iconUrls: ['future'],
   },
 
   /**
    * Information about the "chronicleYellowstone" chain.
    */
   chronicleYellowstone: {
```
MaximusHaximus, line 701: As above, since this was never merged into the live/shipped SDK, should we just :hocho: it and only ship `yellowstone`?

MaximusHaximus (changes requested):
I think we can remove the longer versions of `yellowstone` and just ship the short one?

PR #544, opened 2024-07-18
new wasm for new blsful version

glitch003:
We had a peculiar issue - if you encrypted data in a lit action, and tried to decrypt it with the SDK, it would fail in the WASM.

If you encrypted via the SDK, and decrypted via a lit action, it works.  
And obviously, encrypting then decrypting using the SDK works.  And encrypting then decrypting using a lit action works.  So what's wrong?

The issue was blsful got a version bump, and this new version was on the node but not in the WASM.  Something changed in the new version - when i deserialized the `ciphertext` from a new version (from the node) using the old bls wasm crate, i saw that the `scheme` in the ciphertext struct was set to `MessageAugmentation` but it should be `ProofOfPossession`.  I tried tweaking the scheme manually to be `ProofOfPossession` and the decryption worked!  So, something changed about the ciphertext serialization between blsful versions.  

This PR was created by me just running build.sh in this repo: https://github.com/LIT-Protocol/lit-bls-wasm which pulled down the latest blsful crate.  I tested this using this build here https://www.npmjs.com/package/@lit-protocol/lit-node-client/v/6.2.1-dev.3 and it worked!  I was able to decrypt something using the JS SDK that was encrypted in a lit action.

I think this should be safe and backwards compatible.  Remember - the nodes could decrypt items decrypted with the SDK.  This means that encryptions created with the old blsful crate can be decrypted by the new crate.  But the opposite is not true.  So by bringing the JS SDK up to the latest blsful version, we should be all good.

This PR builds lit-bls-wasm from the latest blsful 2.5.7.  Previously it was using 2.4.2.  Note that we need to do this anytime blsful gets a new version - we need to keep the version on the node in step with the one on the SDK.  I believe the one on the node got bumped automatically by cargo or when someone ran `cargo update` and that's how we ended up here.

### DIFF
--- a/packages/bls-sdk/src/lib/bls-sdk.ts
+++ b/packages/bls-sdk/src/lib/bls-sdk.ts
@@ -1,6 +1,152 @@
 // @ts-nocheck
 import * as pako from 'pako';
 
+// Encoding conversions
+
+// modified from https://stackoverflow.com/a/11058858
+function asciiToUint8Array(a) {
+  let b = new Uint8Array(a.length);
+  for (let i = 0; i < a.length; i++) {
+    b[i] = a.charCodeAt(i);
+  }
+  return b;
+}
+// https://stackoverflow.com/a/19102224
+// TODO resolve RangeError possibility here, see SO comments
+function uint8ArrayToAscii(a) {
+  return String.fromCharCode.apply(null, a);
+}
+// https://stackoverflow.com/a/50868276
+function hexToUint8Array(h) {
+  if (h.length == 0) {
+    return new Uint8Array();
+  }
+  return new Uint8Array(h.match(/.{1,2}/g).map((byte) => parseInt(byte, 16)));
+}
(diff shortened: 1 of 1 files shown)


joshLong145:
Thanks Chris! Could you link to the bls combiner PR where these changes where made?

glitch003:
> Thanks Chris! Could you link to the bls combiner PR where these changes where made?

sure!  https://github.com/LIT-Protocol/lit-bls-wasm/pull/4

All I did was rebuild the wasm, but, critically, i removed cargo.lock first.  cargo.lock had locked blsful to the old version, from the last time i built it on my system, and cargo.lock isn't in the repo.  actually i will add it to the repo now.

joshLong145 (approved):

FedericoAmura (approved):
As far as I can tell, LGTM. The file is consistent with the one at nodes

PR #545, opened 2024-07-18
fix: determin correct protocol by network and port number

Ansonhkg:
# Description

Add mappers that decides whether to use http or https. It now looks at both the network type and port number to make this decision. This makes sure we're using the right protocol for each different network in our system.

## Notable changes

- Moved mapper constants to `mappers.ts` for better readability and upgradability.
- Added `HTTP_BY_NETWORK` to determine network values to corresponding http protocol.
- Added `CENTRALISATION_BY_NETWORK` to determine network values to corresponding centralisation status.


joshLong145:
@Ansonhkg 
Looks like handshake is failing on `datil-test` with some ssl errors so perhaps the selection algo needs tweaking?

PR #547, opened 2024-07-19
Feature/lit 3582 js sdk 70 breakdown lit auth client

FedericoAmura:
# Description

This PR removes LitAuthClient and every place it is used. It keeps the package, users will still need to use LitRelay (now also exported) and the different auth method providers they want

## Type of change

# How Has This Been Tested?

- [x] Tests included in CI still pass
- [x] Locally run an [updated Lit PKP Auth Demo version](https://github.com/LIT-Protocol/lit-pkp-auth-demo/pull/14) with the required updates (migrating away of LitAuthClient to individual providers)

# Checklist:


packages/lit-auth-client/src/lib/lit-auth-client.ts
```diff
@@ -1,395 +0,0 @@
...
-   * If no permissions are provided then each auth method will be assigned `1` for sign anything
-   * If no flags are provided then `sendPkpToitself` will be false, and `addPkpEthAddressAsPermittedAddress` will be true
-   * It is then up to the implementor to transfer the pkp nft to the pkp address.
-   * **note** When adding permissions, each permission should be added in the same order the auth methods are ordered
-   * @throws {Error} - Throws an error if no AuthMethods are given
-   * @param {AuthMethod[]} - AuthMethods authentication methods to be added to the pkp
-   * @param {{ pkpPermissionScopes?: number[][]; sendPkpToitself?: boolean; addPkpEthAddressAsPermittedAddress?: boolean;}}
-   * @returns {Promise<{pkpTokenId?: string; pkpEthAddress?: string; pkpPublicKey?: string}>} pkp information
-   */
-  public async mintPKPWithAuthMethods(
```
joshLong145, line 319: I think we need to port this method to `relay.ts` since we will need an upgrade path for users who relied on this method.

local-tests/setup/tinny-person.ts
```diff
@@ -49,6 +49,10 @@ export class TinnyPerson {
     this.wallet = new ethers.Wallet(privateKey, this.provider);
   }
 
+  async getAuthMethodId(): Promise<string> {
+    return await EthWalletProvider.authMethodId(this.authMethod);
+  }
+
```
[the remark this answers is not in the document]
FedericoAmura: We could, however modern JS handles this correctly without the extra microtask of unwrapping&wrapping the value in a promise. And also this keeps the stacktrace which, in my opinion, is more than enough to justify that cost, if it were still a thing

Even more, the [eslint rule to prevent this](https://eslint.org/docs/latest/rules/no-return-await) was deprecated because of that

Made an example [here](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBANiA5jAvDUkRwKYDoGIDcAUMRtDAO4CGAlrGgBQC2EAXDGAK7MBG2AJwCUqAHydslGAAUBIZrQjYAPADcQtACajGjAdghZV2ESnFKoAFVrNsILlF2nx+w3GOMhAGhishQkmJqCABPMGAYADMucKhacBgoAAs5SgBRAA9gbAAHOPAAQUioQQKaek8YAG9iGBhqcthGxgBGAAY2gNrElJApMEkYNIE5AUYAcksACQAlAHkFxaWlgHU1lYBCLY3xroBfUmCwiOjY+LAomOB8sEsk6igAYWo4OAgCsALQ8LvU2jBEAAxK43RiNWTyRTYDi8EBYbDUMAiGp1WiRGBguhQCEKJTI7p1fRQLgCC4NLE9VKZbJ5c5FEoCMpYzwkOp7GDYN7YaoEmBEkkXZJUrK5G700rNLpskgHILfE4g86+OhgSoo+BICYzGJgf7IZgq3asxICEI8up1cn0S5ncB3B7PV7vT7y359PXA22qqACLgmY3s4APYBJDEmc0W8iwQSjVAc+oQGAxADWYD6ZMTw1GxrqBAmTPoepgOTkuOwPhjIAEMEUbCNvLzlYEuGg1GAyalMFl3R9ZvVlsaNuu53tTxebw+X2ObsoHsV4EYkReeID6GDocY4f76HAFCbce5wSTYFT6YTQxGVZzGsQEwAciBYFa4gDi6WoRXL9Xa-WLTfN1+LZQG2HYBsQsoGv8ni4EGUAhow5DwrgTZdEAA) 😃

packages/lit-auth-client/src/lib/lit-auth-client.ts
FedericoAmura: Good call, restored

packages/lit-auth-client/src/lib/relay.ts
```diff
@@ -1,16 +1,36 @@
...
  */
 export class LitRelay implements IRelay {
+  /** URL for Lit's relay server */
+  static getRelayUrl(litNetwork: LitNetwork): string {
+    const networkMap: Record<LitNetwork, string | undefined> = {
+      [LitNetwork.Cayenne]: RELAY_URL_CAYENNE,
+      [LitNetwork.Manzano]: RELAY_URL_MANZANO,
+      [LitNetwork.Habanero]: RELAY_URL_HABANERO,
+      [LitNetwork.DatilDev]: RELAY_URL_DATIL_DEV,
+      [LitNetwork.Custom]: undefined,
```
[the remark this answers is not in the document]
FedericoAmura: Good point. I've added a `throw` for the case where `relayUrl` is not defined instead of returning the cayenne value as that could hide runtime typos or other errors

I'll keep `LitNetwork.Custom` so we can keep the `networkMap: Record<LitNetwork, string | undefined>` and have it require every `LitNetwork` without complex types. It does not change the behaviour anyway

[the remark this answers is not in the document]
Ansonhkg: Let's merge `staging/v7` in  (I just merged master into staging/v7) and use this mapper as we are deprecating the olds:
https://github.com/LIT-Protocol/js-sdk/blob/1a843e951ff5cb1d9203075fe829434bd7f471d4/packages/constants/src/lib/constants/constants.ts#L730-L740

local-tests/tests/testRelayer.ts
```diff
@@ -21,18 +14,17 @@ import { ProviderType } from '@lit-protocol/constants';
...
+    relay: litRelay,
     litNodeClient: devEnv.litNodeClient,
   });
 
-  // -- test fetch pkps
-  const ethWalletProvider = litAuthClient.initProvider<EthWalletProvider>(
-    ProviderType.EthWallet
-  );
-
   const pkps = await ethWalletProvider.fetchPKPsThroughRelayer(
```
Ansonhkg, line 28: Let's deprecate the `fetchPKPsThroughRelayer` function? It's just a contract read I don't see the reason to read it from the relayer (cc @MaximusHaximus )

fyi here's the relayer flow:

---

1. endpoint:

```ts
app.post("/fetch-pkps-by-auth-method", fetchPKPsHandler);
```

3. call `fetchPKPsHandler` function

```ts
// Fetch PKPs for verified Discord account
export async function fetchPKPsHandler(
	req: Request<
		{},
		AuthMethodVerifyToFetchResponse,
		FetchRequest,
		ParsedQs,
		Record<string, any>
	>,
	res: Response<AuthMethodVerifyToFetchResponse, Record<string, any>, number>,
) {
	// get Discord access token from body
	const { authMethodType, authMethodId } = req.body;

	try {
		const pkps = await getPKPsForAuthMethod({
			authMethodType: authMethodType,
			idForAuthMethod: authMethodId,
		});
		console.info("Fetched PKPs with auth type", authMethodType, {
			pkps: pkps,
		});
		return res.status(200).json({
			pkps: pkps,
		});
	} catch (err) {
		console.error(
			`Unable to fetch PKPs for given auth type ${authMethodType}`,
			{
				err,
			},
		);
		return res.status(500).json({
			error: `Unable to fetch PKPs for given auth method type: ${authMethodType}`,
		});
	}
}
```

4. call `getPKPsForAuthMethod` function:

```ts
export async function getPKPsForAuthMethod({
	authMethodType,
	idForAuthMethod,
}: {
	authMethodType: AuthMethodType;
	idForAuthMethod: string;
}) {
	if (!authMethodType || !idForAuthMethod) {
		throw new Error(
			"Auth method type and id are required to fetch PKPs by auth method",
		);
	}

	const pkpPermissions = await getPermissionsContract();
	if (pkpPermissions) {
		try {
			const tokenIds = await pkpPermissions.getTokenIdsForAuthMethod(
				authMethodType,
				idForAuthMethod,
			);
			const pkps: PKP[] = [];
			for (let i = 0; i < tokenIds.length; i++) {
				const pubkey = await pkpPermissions.getPubkey(tokenIds[i]);
				if (pubkey) {
					const ethAddress = ethers.utils.computeAddress(pubkey);
					pkps.push({
						tokenId: tokenIds[i],
						publicKey: pubkey,
						ethAddress: ethAddress,
					});
				}
			}
			return pkps;
		} catch (err) {
			throw new Error("Unable to get PKPs for auth method");
		}
	} else {
		throw new Error("Unable to connect to PKP Permissions contract");
	}
}
```

packages/lit-auth-client/src/lib/utils.ts
```diff
@@ -277,14 +286,14 @@ export function parseAuthenticatorData(
 // **`unparse()` - Convert UUID byte array (ala parse()) into a string**
 export function unparse(buf: any) {
```
Ansonhkg, line 287: nit: do we have a unit test on this?

Ansonhkg (changes requested):

packages/lit-auth-client/src/lib/utils.ts
FedericoAmura: We don't, tried adding it but it was a hell to set. Anyway those changes should not impact logic as are just naming or declaration keywords

packages/lit-auth-client/src/lib/relay.ts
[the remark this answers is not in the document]
FedericoAmura: Done

packages/lit-auth-client/src/lib/providers/BaseProvider.ts
```diff
@@ -112,6 +114,80 @@ export abstract class BaseProvider {
...
+    if (!authMethodType || !authMethodId) {
+      throw new Error(
+        'Auth method type and id are required to fetch PKPs by auth method'
+      );
+    }
+
+    const litContracts = new LitContracts({
+      // Using a random wallet, we just need to read
+      signer: ethers.Wallet.createRandom(),
+    });
```
Ansonhkg, line 141: the `LitContracts` instance is by default using the `cayenne` network, so we are actually fetching PKPs from the wrong intended network. 

```suggestion
    const litContracts = new LitContracts({
      randomPrivatekey: true,
      network: this.litNodeClient.config.litNetwork,
    });
```

Ansonhkg (changes requested):

Ansonhkg (approved):
Thanks for getting this fixed! LGTM now! :))

PR #548, opened 2024-07-22
feat: add node error inside sdk error v1 wrapper

FedericoAmura:
# Description

This PR adds the node error data inside the error thrown in SDK wrappers. This way it is not lost in context and can be easily seen in console, not having to also search in network tab
Screenshot from 2024-07-22 16-53-18

## Type of change

# How Has This Been Tested?

- [x] Throw lit action error in a demo app and checking its console

# Checklist:

### DIFF
--- a/packages/misc/src/lib/misc.ts
+++ b/packages/misc/src/lib/misc.ts
@@ -144,10 +144,38 @@ export const throwErrorV0 = ({
 };
 
 // Map for old error codes to new ones
-const oldErrorToNewErrorMap: { [key: string]: string } = {
+const oldErrorToNewErrorMap: Record<string, string> = {
   not_authorized: 'NodeNotAuthorized',
   storage_error: 'NodeStorageError',
 };
+
+class GenericError extends Error {
+  readonly requestId?: string;
+  readonly errorCode: string;
+  readonly errorKind: string;
+
+  constructor(message: string, requestId?: string) {
+    super(message);
+    this.requestId = requestId;
+    this.errorCode = LIT_ERROR.UNKNOWN_ERROR.code;
+    this.errorKind = LIT_ERROR.UNKNOWN_ERROR.name;
+  }
+}
+
+class ErrorV1 extends GenericError {
+  override readonly errorCode: string;
+  override readonly errorKind: string;
+  readonly status: number | undefined;
+  readonly details: string[] | undefined;
+
+  constructor(message: string, errorKind: string, errorCode: string, status?: number, details?: string[], requestId?: string) {
+    super(message, requestId);
+    this.status = status;
+    this.details = details;
+    this.errorCode = errorCode;
+    this.errorKind = errorKind;
+  }
+}
 
 /**
  *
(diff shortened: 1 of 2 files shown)


Ansonhkg (approved):
lgtm!

nice of switching from `void` to `never` cus `void` would mean something still works completely normally, whereas `never` indicates that the function either throws an error or has an infinite loop.

packages/misc/src/lib/misc.ts
```diff
@@ -144,11 +144,46 @@ export const throwErrorV0 = ({
 };
 
 // Map for old error codes to new ones
-const oldErrorToNewErrorMap: { [key: string]: string } = {
+const oldErrorToNewErrorMap: Record<string, string> = {
   not_authorized: 'NodeNotAuthorized',
   storage_error: 'NodeStorageError',
 };
 
+class GenericError extends Error {
```
MaximusHaximus, line 152: 1. So there's a bit of a weird situation here, because we've been `throw`ing a lot of plain javascript objects which serialise to JSON -- but the native Error object _doesn't_. Note that if you `console.log(JSON.stringify(new Error('wat')));` you'll see `{}` is the output. This is because standard Errors don't include a toJSON -- historically, this was because the `stack` property is 'special'.

So in order to keep existing serialisation behaviour somewhat close (e.g. people will check if JSON.stringify(err) `contains` a string, but `{}` never will contain any of the strings they are checking, it would be good to go ahead and make these errors extend from `openagenda/verror`, which defines a solid `toJSON()` implementation.

For context for the reader:
- VError is Joyent's canonical error package for NodeJS.
- `openagenda/verror` is a browser-friendly version of `verror` that is much smaller than verror and has no dependencies on nodeJS internals.

2. Note that `@openagenda/verror` supports injecting any number of properties without a customized constructor signature, by providing `meta` in the constructor; this way constructing our errors will be consistent everywhere.

3. `GenericError` is not that generic; it actually is quite opinionated that it will have a requestId, for example.  Let's call it `LitNodeRequestError` instead for clarity?

4. `LitNodeRequestErrorV1` would be something like:
```javascript
class LitNodeRequestErrorV1 extends LitNodeRequestError {
  readonly status?: number
  readonly details?: string[]
}
```

Constructing it would be:
```
throw new LitNodeRequestErrorV1({
  name: 'LitNodeRequestErrorV1'
  meta: {
    requestId,
    errorKind,
    errorCode,
    status,
    details,
  },
  constructorOpt: throwErrorV1
});
```

Note:
- The use of `constructorOpt` will eliminate the `throwErrorV1` function from the stack trace on the error
- `meta` keys are set on the error as properties without needing to make a custom constructor signature. We can use a function for that purpose, and avoid constructor signature mismatches and `super()` order of execution irritations
- To see examples of how to subclass VError with pre-set values injected for `meta`, but supporting the caller to also override those, and ensuring the entire VError constructor API is maintained, check out this example where http errors are being shipped in `@openagenda/verror`: https://github.com/OpenAgenda/verror/blob/master/src/http.js#L8

packages/misc/src/lib/misc.ts
```diff
@@ -210,16 +213,13 @@ export const throwGenericError = (e: any): never => {
 export const isNodeClientErrorV1 = (
   nodeError: NodeClientErrorV0 | NodeClientErrorV1
 ): nodeError is NodeClientErrorV1 => {
-  return (
-    nodeError.hasOwnProperty('errorCode') &&
-    nodeError.hasOwnProperty('errorKind')
-  );
+  return 'errorCode' in nodeError && 'errorKind' in nodeError;
```
MaximusHaximus, line 216: Using VError, we would use `VError.meta()` to get a flattened object, then continue to use `hasOwnProperty` https://www.npmjs.com/package/@openagenda/verror#verrormetaerr`.  Or even better, we could just check `name` of the error.  Generally, all errors should have a `name` and that it a reasonable way to identify a particular type of error that is less brittle than using `instanceof`

packages/misc/src/lib/misc.ts
```diff
@@ -210,16 +213,13 @@ export const throwGenericError = (e: any): never => {
...
 };
 
 export const isNodeClientErrorV0 = (
   nodeError: NodeClientErrorV0 | NodeClientErrorV1
 ): nodeError is NodeClientErrorV0 => {
-  return nodeError.hasOwnProperty('errorCode');
+  return 'nodeError' in nodeError;
```
MaximusHaximus, line 222: Should this be checking `errorCode`?  As w/ above comment, rather than ducktype it, it'd be better if wherever we throw these types of errors set the `name` to a specific value and we check that instead of duck-typing based on a single property.

MaximusHaximus (changes requested):
<3 thanks for starting down this path -- I know we had spoken about @openagenda/verror directly, but I added some notes here for your (and other readers) reference

packages/misc/src/lib/misc.ts
FedericoAmura: This is removed in the PR https://github.com/LIT-Protocol/js-sdk/pull/576 so it would not apply anymore
Lucky us, it was only used in `throwError`

packages/misc/src/lib/misc.ts
FedericoAmura: This is removed in the PR https://github.com/LIT-Protocol/js-sdk/pull/576 so it would not apply anymore
Lucky us, it was only used in `throwError`

packages/misc/src/lib/misc.ts
FedericoAmura: This is solved with https://github.com/LIT-Protocol/js-sdk/pull/576

As the plan is to merge and release v7 now, I'm closing this PR. The other one is way better and this one will only create conflicts

FedericoAmura:
This PR is replaced by https://github.com/LIT-Protocol/js-sdk/pull/548 which brings a much better solution to this problem

PR #549, opened 2024-07-22
Feature/lit 2921 js sdk migrate to jest

joshLong145:
- Migration of `local-tests` from custom test runner to `Jest` for standardized test runner implementations.
- Creates a package for `tinny` which has been migrated from the `setup` directory in `local-tests`
- Has migrated explicit test cases for each permutation of `session` generation to an implicit implementations for code reuse.

At this time we still rely on the global installation of `packages` into `node_modules` for binding to other local packages through package import statements. 

To reduce the scope of this PR the migration away from global module installation has not been implemented. This will be done in a followup PR.

### New Test Command
For running new `jest` test suite the `test:e2e` command can be used. CI has been migrated to this command in this PR.


packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+    });
+  });
+
+  it('Invalid lit action Custom Auth SessionSigs', async () => {
+    devEnv.setUnavailable(LIT_TESTNET.MANZANO);
+
+    const alice = await devEnv.createRandomPerson();
+
+    try {
+      await getInvalidLitActionSessionSigs(devEnv, alice);
```
MaximusHaximus, line 240: We can use `expect().toThrow(...)` here, which will let us :hocho: the try/catch entirely and get rid of the `console.logs`. The log on L242 is particularly confusing because we don't yet _know_ that it's actually expected, and if it _was_ expected we see a big red X -- ideally all of this would be condensed into a single `.toThrow()` check

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+    }
+  });
+
+  it('Invalid Lit Action Custom Auth IPFS SessionSigs', async () => {
+    const alice = await devEnv.createRandomPerson();
+
+    try {
+      await getInvalidLitActionIpfsSessionSigs(devEnv, alice);
+    } catch (e: any) {
+      console.log('❌ THIS IS EXPECTED: ', e);
```
MaximusHaximus, line 264: As above, We can use `expect().toThrow(...)` here, which will let us :hocho: the try/catch entirely and get rid of the `console.logs`. The log on L242 is particularly confusing because we don't yet _know_ that it's actually expected, and if it _was_ expected we see a big red X -- ideally all of this would be condensed into a single `.toThrow()` check

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+
+  const res = await devEnv.litNodeClient?.executeJs({
+    sessionSigs: litActionSessionSigs,
+    code: `(async () => {
+    Lit.Actions.claimKey({keyId: "foo"});
+    Lit.Actions.claimKey({keyId: "bar"});
+  })();`,
+  });
+
+  devEnv.releasePrivateKeyFromUser(alice);
```
MaximusHaximus, line 307: Need this to be in a `finally()` block to ensure it is released if any of the code from L288 until this line throws an error

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+
+const executeJsClaimKey = async (
+  devEnv: TinnyEnvironment,
+  generator: (
+    devEnv: TinnyEnvironment,
+    person: TinnyPerson,
+    resources?: LitResourceAbilityRequest[]
+  ) => Promise<SessionSigsMap | undefined>
+): Promise<void> => {
+  devEnv.setUnavailable(LIT_TESTNET.MANZANO);
```
MaximusHaximus, line 387: 🔪 ideally this would be replaced with wrapping the test definition function call itself in a conditional check, see above comment.

That being said, do we _really_ never run any test that uses this function on manzano -- and should it be disabled on `datil-test` if it was disabled on `manzano`?

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+
+const pkpSign = async (
+  devEnv: TinnyEnvironment,
+  generator: (
+    devEnv: TinnyEnvironment,
+    person: TinnyPerson,
+    resources?: LitResourceAbilityRequest[]
+  ) => Promise<SessionSigsMap | undefined>
+): Promise<void> => {
+  devEnv.setUnavailable(LIT_TESTNET.MANZANO);
```
MaximusHaximus, line 450: 🔪 ideally this would be replaced with wrapping the test definition function call itself in a conditional check, see above comment.

That being said, do we _really_ never run any test that uses this function on manzano -- and should it be disabled on `datil-test` if it was disabled on `manzano`?

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+  const alice = await devEnv.createRandomPerson();
+  const litActionSessionSigs = await generator(devEnv, alice);
+
+  const res = await devEnv.litNodeClient?.pkpSign({
+    toSign: alice.loveLetter,
+    pubKey: alice.authMethodOwnedPkp?.publicKey!,
+    sessionSigs: litActionSessionSigs!,
+  });
+
+  devEnv.releasePrivateKeyFromUser(alice);
```
MaximusHaximus, line 461: Need this to be in a finally{} block so that if code throws on L453-459, we release the key

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+    console.log('hello world')
+
+    LitActions.setResponse({
+      response: JSON.stringify({hello: 'world'})
+    });
+
+  })();`,
+  });
+
+  devEnv.releasePrivateKeyFromUser(alice);
```
MaximusHaximus, line 504: Need this to be in a finally{} block so that if code throws on L490-502, we release the key

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+
+const executeJsJSONResponse = async (
+  devEnv: TinnyEnvironment,
+  generator: (
+    devEnv: TinnyEnvironment,
+    person: TinnyPerson,
+    resources?: LitResourceAbilityRequest[]
+  ) => Promise<SessionSigsMap | undefined>
+): Promise<void> => {
+  devEnv.setUnavailable(LIT_TESTNET.MANZANO);
```
MaximusHaximus, line 487: 🔪 ideally this would be replaced with wrapping the test definition function call itself in a conditional check, see above comment.

That being said, do we _really_ never run any test that uses this function on manzano -- and should it be disabled on `datil-test` if it was disabled on `manzano`?

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+
+const executeJsSigningParallel = async (
+  devEnv: TinnyEnvironment,
+  generator: (
+    devEnv: TinnyEnvironment,
+    person: TinnyPerson,
+    resources?: LitResourceAbilityRequest[]
+  ) => Promise<SessionSigsMap | undefined>
+): Promise<void> => {
+  devEnv.setUnavailable(LIT_TESTNET.MANZANO);
```
MaximusHaximus, line 563: 🔪 ideally this would be replaced with wrapping the test definition function call itself in a conditional check, see above comment.

That being said, do we _really_ never run any test that uses this function on manzano -- and should it be disabled on `datil-test` if it was disabled on `manzano`?

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+
+const decryptString = async (
+  devEnv: TinnyEnvironment,
+  generator: (
+    devEnv: TinnyEnvironment,
+    person: TinnyPerson,
+    resources?: LitResourceAbilityRequest[]
+  ) => Promise<SessionSigsMap | undefined>
+): Promise<void> => {
+  devEnv.setUnavailable(LIT_TESTNET.MANZANO);
```
MaximusHaximus, line 606: 🔪 ideally this would be replaced with wrapping the test definition function call itself in a conditional check, see above comment.

That being said, do we _really_ never run any test that uses this function on manzano -- and should it be disabled on `datil-test` if it was disabled on `manzano`?

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+
+const broadcastAndCollect = async (
+  devEnv: TinnyEnvironment,
+  generator: (
+    devEnv: TinnyEnvironment,
+    person: TinnyPerson,
+    resources?: LitResourceAbilityRequest[]
+  ) => Promise<SessionSigsMap | undefined>
+) => {
+  devEnv.setUnavailable(LIT_TESTNET.MANZANO);
```
MaximusHaximus, line 676: 🔪 ideally this would be replaced with wrapping the test definition function call itself in a conditional check, see above comment.

That being said, do we _really_ never run any test that uses this function on manzano -- and should it be disabled on `datil-test` if it was disabled on `manzano`?

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+
+const decryptAndCombine = async (
+  devEnv: TinnyEnvironment,
+  generator: (
+    devEnv: TinnyEnvironment,
+    person: TinnyPerson,
+    resources?: LitResourceAbilityRequest[]
+  ) => Promise<SessionSigsMap | undefined>
+) => {
+  devEnv.setUnavailable(LIT_TESTNET.MANZANO);
```
MaximusHaximus, line 714: 🔪 ideally this would be replaced with wrapping the test definition function call itself in a conditional check, see above comment.

That being said, do we _really_ never run any test that uses this function on manzano -- and should it be disabled on `datil-test` if it was disabled on `manzano`?

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+        sigName: "sig",
+      });
+    })();`,
+    jsParams: {
+      dataToSign: alice.loveLetter,
+      publicKey: alice.authMethodOwnedPkp?.publicKey,
+    },
+  });
+
+  devEnv.releasePrivateKeyFromUser(alice);
```
MaximusHaximus, line 547: Need this to be in a finally{} block so that if code above throws, we release the key

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+    jsParams: {
+      accessControlConditions: accs,
+      dataToEncryptHash: encryptRes.dataToEncryptHash,
+      ciphertext: encryptRes.ciphertext,
+    },
+  });
+
+  devEnv.releasePrivateKeyFromUser(alice);
+
+  expect(res?.response).toEqual('Hello world');
```
MaximusHaximus, line 773: Can get a nicer output in cases where this fails by using `expect(res).toHaveProperty('response', 'Hello world')
https://jestjs.io/docs/expect#tohavepropertykeypath-value

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+  );
+
+  // -- Expected output:
+  // {
+  //   ciphertext: "pSP1Rq4xdyLBzSghZ3DtTtHp2UL7/z45U2JDOQho/WXjd2ntr4IS8BJfqJ7TC2U4CmktrvbVT3edoXJgFqsE7vy9uNrBUyUSTuUdHLfDVMIgh4a7fqMxsdQdkWZjHign3JOaVBihtOjAF5VthVena28D",
+  //   dataToEncryptHash: "64ec88ca00b268e5ba1a35678a1b5316d212f4f366b2477232534a8aeca37f3c",
+  // }
+
+  // -- assertions
+  if (!encryptRes.ciphertext) {
```
MaximusHaximus, line 742: should use `expect()` here

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+    throw new Error(`Expected "ciphertext" in encryptRes`);
+  }
+
+  if (!encryptRes.dataToEncryptHash) {
```
MaximusHaximus, line 746: should use `expect()` here

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+            });
+          })();`,
+    jsParams: {
+      accessControlConditions: accs,
+      dataToEncryptHash: encryptRes.dataToEncryptHash,
+      ciphertext: encryptRes.ciphertext,
+    },
+  });
+
+  devEnv.releasePrivateKeyFromUser(alice);
```
MaximusHaximus, line 771: Need this to be in a finally{} block so that if code above throws, we release the key

packages/e2e-tests/src/tests/SessionSigs.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+                response: sigShare
+            });
+          })();`,
+    jsParams: {
+      dataToSign: alice.loveLetter,
+      publicKey: alice.pkp?.publicKey,
+    },
+  });
+
+  devEnv.releasePrivateKeyFromUser(alice);
```
MaximusHaximus, line 806: Need this to be in a finally{} block so that if code above throws, we release the key

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+  return result;
+}
+
+try {
+  jest.setTimeout(100_0000);
+} catch (e) {
+  // ... continue execution
+}
+
+describe('Wrapped Keys', () => {
```
MaximusHaximus, line 72: Missing calls `releasePrivateKey` here.  Since _all_ of the tests in this suite _only_ use alice, this is probably easiest done by moving `alice` into the describe() block as a local mutable variable, and setting it and releasing the key in `beforeEach()` and `afterEach()`

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+    const alice = await devEnv.createRandomPerson();
+
+    const pkpSessionSigs = await getPkpSessionSigs(
+      devEnv,
+      alice,
+      undefined,
+      new Date(Date.now() + 1000 * 60 * 10).toISOString()
+    ); // 10 mins expiry
+
+    console.log(pkpSessionSigs);
```
MaximusHaximus, line 232: :hocho: We removed this in upstream as it was excessively spammy

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
```
MaximusHaximus, line 260: :hocho: We removed this in upstream as it was excessively spammy

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+    });
+
+    const signatureIsValidForPublicKey = nacl.sign.detached.verify(
+      Buffer.from(messageToSign),
+      bs58.decode(signature),
+      bs58.decode(generatedPublicKey)
+    );
+
+    if (!signatureIsValidForPublicKey)
+      throw new Error(
```
MaximusHaximus, line 296: We should use `expect()` here

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+        pkpSessionSigs: pkpSessionSigs!,
+        privateKey,
+        litNodeClient: devEnv.litNodeClient!,
+        publicKey: '0xdeadbeef',
+        keyType: 'K256',
+        memo: 'Test Key',
+      });
+    } catch (e: any) {
+      if (
+        e.message.includes(
```
MaximusHaximus, line 390: We can use expect.toThrow() here instead, and :hocho: the extra `console.log`
See also https://jestjs.io/docs/expect#rejects :)

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+    const alice = await devEnv.createRandomPerson();
+
+    const pkpSessionSigs = await getPkpSessionSigs(
+      devEnv,
+      alice,
+      undefined,
+      new Date(Date.now() + 1000 * 60 * 10).toISOString()
+    ); // 10 mins expiry
+
+    console.log(pkpSessionSigs);
```
MaximusHaximus, line 411: :hocho: we removed this from upstream because it was so spammy

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+      }
+    }
+  });
+
+  it('Fail Max Expiration Session Sig', async () => {
+    const alice = await devEnv.createRandomPerson();
+
+    const pkpSessionSigs = await getPkpSessionSigs(devEnv, alice);
+
+    console.log(pkpSessionSigs);
```
MaximusHaximus, line 455: :hocho: we removed this from upstream because it was so spammy

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+      await importPrivateKey({
+        pkpSessionSigs: pkpSessionSigs!,
+        privateKey,
+        litNodeClient: devEnv.litNodeClient!,
+        publicKey: '0xdeadbeef',
+        keyType: 'K256',
+        memo: 'Test Key',
+      });
+    } catch (e: any) {
+      if (e.message.includes('Expires too far in the future')) {
```
MaximusHaximus, line 469: We can use expect.toThrow() here instead, and :hocho: the extra `console.log`
See also https://jestjs.io/docs/expect#rejects :)

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+      await importPrivateKey({
+        pkpSessionSigs: tamperPkpSessionSigs(pkpSessionSigs!),
+        privateKey,
+        litNodeClient: devEnv.litNodeClient!,
+        publicKey: '0xdeadbeef',
+        keyType: 'K256',
+        memo: 'Test Key',
+      });
+    } catch (e: any) {
+      if (e.message.includes('bad public key size')) {
```
MaximusHaximus, line 512: We can use expect.toThrow() here instead, and :hocho: the extra `console.log`
See also https://jestjs.io/docs/expect#rejects :)

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+      await importPrivateKey({
+        pkpSessionSigs,
+        privateKey,
+        litNodeClient: devEnv.litNodeClient!,
+        publicKey: '0xdeadbeef',
+        keyType: 'K256',
+        memo: 'Test Key',
+      });
+    } catch (e: any) {
+      if (e.message.includes('Invalid sessionSig: Expired')) {
```
MaximusHaximus, line 560: We can use expect.toThrow() here instead, and :hocho: the extra `console.log`
See also https://jestjs.io/docs/expect#rejects :)

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+        unsignedTransaction: {
+          ...getChainForNetwork(devEnv.litNodeClient?.config?.litNetwork!),
+          // @ts-expect-error This test is intentionally using the type incorrectly.
+          serializedTransaction: 'random-value',
+        },
+        broadcast: false,
+        litNodeClient: devEnv.litNodeClient!,
+      });
+    } catch (e: any) {
+      if (
```
MaximusHaximus, line 639: We can use expect.toThrow() here instead, and :hocho: the extra `console.log`
See also https://jestjs.io/docs/expect#rejects :)

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+    const alice = await devEnv.createRandomPerson();
+
+    const pkpSessionSigs = await getPkpSessionSigs(
+      devEnv,
+      alice,
+      undefined,
+      new Date(Date.now() + 1000 * 60 * 10).toISOString()
+    ); // 10 mins expiry
+
+    console.log(pkpSessionSigs);
```
MaximusHaximus, line 661: :hocho: we removed this from upstream because it was so spammy

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+    const signature = await signMessageWithEncryptedKey({
+      pkpSessionSigs: pkpSessionSigsSigning!,
+      network: 'evm',
+      messageToSign: unsignedStringMessage,
+      litNodeClient: devEnv.litNodeClient!,
+      id,
+    });
+
+    console.log('signature');
+    console.log(signature);
```
MaximusHaximus, line 897: :hocho: we removed this from upstream because it was so spammy (and `expect()` will print the value if it doesn't match what we expect!)

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+      return {
+        chain: 'chronicleTestnet',
+        chainId: LIT_CHAINS['chronicleTestnet'].chainId,
+      };
+    case 'datil-dev':
+      return {
+        chain: 'yellowstone',
+        chainId: LIT_CHAINS['yellowstone'].chainId,
+      };
+    case 'datil-test':
```
MaximusHaximus, line 1164: Need case for `datil` now?

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1211 @@
...
+    case 'cayenne':
+    case 'habanero':
+    case 'manzano':
+      return {
+        gasPrice: '0.001',
+        gasLimit: 30000,
+      };
+    case 'datil-dev':
+      return { gasLimit: 5000000 };
+    case 'datil-test':
```
MaximusHaximus, line 1188: Need case for `datil` now?

packages/e2e-tests/src/tests/connection.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(devEnv.litNodeClient).toBeDefined();
+    expect(devEnv.litNodeClient?.ready).toBe(true);
+    expect(devEnv.litNodeClient?.config.litNetwork).toBe(NETWORKS[0]);
+    expect(devEnv.litNodeClient?.networkPubKey).toBeDefined();
+    expect(devEnv.litNodeClient?.hdRootPubkeys).toBeDefined();
+    expect(devEnv.litNodeClient?.connectedNodes?.size).toBeGreaterThanOrEqual(
+      devEnv.litNodeClient?.config?.minNodeCount!
+    );
+
+    devEnv.litNodeClient?.disconnect();
```
MaximusHaximus, line 38: Needs `await`

packages/e2e-tests/src/tests/connection.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+const NETWORKS: string[] = ['manzano', 'datil-dev', 'cayenne'];
+
+describe('Connections', () => {
+  beforeEach(() => {
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    // TODO: can be removed once v7 is merged with wasm refactors
+    delete globalThis.wasmExports;
```
MaximusHaximus, line 19: This should be happening in `disconnect()` if we await it (it calls `unloadModules`)

packages/e2e-tests/src/tests/connection.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(devEnv.litNodeClient).toBeDefined();
+    expect(devEnv.litNodeClient?.ready).toBe(true);
+    expect(devEnv.litNodeClient?.config.litNetwork).toBe(NETWORKS[1]);
+    expect(devEnv.litNodeClient?.networkPubKey).toBeDefined();
+    expect(devEnv.litNodeClient?.hdRootPubkeys).toBeDefined();
+    expect(devEnv.litNodeClient?.connectedNodes?.size).toBeGreaterThanOrEqual(
+      devEnv.litNodeClient?.config?.minNodeCount!
+    );
+
+    devEnv.litNodeClient?.disconnect();
```
MaximusHaximus, line 53: Needs await

packages/e2e-tests/src/tests/connection.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(devEnv.litNodeClient).toBeDefined();
+    expect(devEnv.litNodeClient?.ready).toBe(true);
+    expect(devEnv.litNodeClient?.config.litNetwork).toBe(NETWORKS[2]);
+    expect(devEnv.litNodeClient?.networkPubKey).toBeDefined();
+    expect(devEnv.litNodeClient?.hdRootPubkeys).toBeDefined();
+    expect(devEnv.litNodeClient?.connectedNodes?.size).toBeGreaterThanOrEqual(
+      devEnv.litNodeClient?.config?.minNodeCount!
+    );
+
+    devEnv.litNodeClient?.disconnect();
```
MaximusHaximus, line 68: Needs await

packages/e2e-tests/src/tests/connection.spec.ts
```diff
@@ -0,0 +1,70 @@
+import { expect, jest, test } from '@jest/globals';
```
MaximusHaximus, line 1: Needs datil and datil-test ?

MaximusHaximus (changes requested):
Added some stuff in-line - this is looking _awesome_.

1. I'm curious -- what is the difference between the `networkContext.example.json` and `networkContext.json`?  If we're going to commit an actual `networkContext.json`, can we remove the example anyway since the actual file is there to serve as a real-world example?

2. Looks like we're missing our `.eslintrc.json` file for this package -- see https://github.com/LIT-Protocol/js-sdk/blob/4c8034a2d084201cf2a2d2a461c80a7f794a2b4c/packages/crypto/.eslintrc.json for an example of what we should have in the root of this package

After you've added it, you can run `eslint --fix` on the entire package to get most of the eslint issues resolved without manually editing each file, but keep in mind that there could be useful errors that pop up in the new code you wrote since currently our eslint rule set is effectively disabled for this package.

3. Is the `packages/tinny` package a duplication of the code in `e2e-tests/setup`? I'm not sure which to add comments on; it appears they are duplicates. If we're going to introduce the tinny package, we should update `e2e-tests` to _use that one package_ instead of having 2 copies of tinny in the codebase at once -- i added comments to the files that are _inside e2e-tests_ before I realised they were duplicated in the new package, and didn't read through the packages/tinny files to verify the same feedback applies for those

nice work on this stuff 💖 , it's a beast of an undertaking

joshLong145:
> Added some stuff in-line - this is looking _awesome_.
> 
> 1. I'm curious -- what is the difference between the `networkContext.example.json` and `networkContext.json`?  If we're going to commit an actual `networkContext.json`, can we remove the example anyway since the actual file is there to serve as a real-world example?
> 2. Looks like we're missing our `.eslintrc.json` file for this package -- see https://github.com/LIT-Protocol/js-sdk/blob/4c8034a2d084201cf2a2d2a461c80a7f794a2b4c/packages/crypto/.eslintrc.json for an example of what we should have in the root of this package
> 
> After you've added it, you can run `eslint --fix` on the entire package to get most of the eslint issues resolved without manually editing each file, but keep in mind that there could be useful errors that pop up in the new code you wrote since currently our eslint rule set is effectively disabled for this package.
> 
> 3. Is the `packages/tinny` package a duplication of the code in `e2e-tests/setup`? I'm not sure which to add comments on; it appears they are duplicates. If we're going to introduce the tinny package, we should update `e2e-tests` to _use that one package_ instead of having 2 copies of tinny in the codebase at once -- i added comments to the files that are _inside e2e-tests_ before I realised they were duplicated in the new package, and didn't read through the packages/tinny files to verify the same feedback applies for those
> 
> nice work on this stuff 💖 , it's a beast of an undertaking

Yes good points. I think I agree with having a single `networkContext.json` as the first deployment of contracts to `anvil` will be deterministic addresses. So we can be safe for now with a static configuration declaration.

You make a great point on Tinny being duplicated. I will add a reference to `packages/tinny` to `local-tests`

Will finally add the lint files and do a run through.

jest.preset.js
```diff
@@ -1,3 +1,34 @@
-const nxPreset = require('@nx/jest/preset').default;
-
-module.exports = { ...nxPreset };
+/**
+ * Taken from https://github.com/nrwl/nx/blob/master/packages/jest/preset/jest-preset.ts
```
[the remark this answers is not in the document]
joshLong145: I wanted to inline as to not be effected if defaults update to something that can possibly cause changes to our test runs without us realizing. I can switch back but this felt safer.

packages/e2e-tests/setup.config.js
```diff
@@ -0,0 +1,34 @@
...
+ * Jest Node Env docs: https://jestjs.io/docs/configuration#testenvironment-string
+ * Global setup file for jest e2e tests
+ * Loads all global context and provides to each test runner as needed
+ * env loading has been moved into
+ */
+
+const NodeEnvironment = require('jest-environment-node').TestEnvironment;
+const TinnyEnvironment = require('@lit-protocol/tinny').TinnyEnvironment;
+
+require('dotenv').config();
```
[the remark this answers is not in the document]
joshLong145: I should actually remove the nx loader as we can no longer rely on the tests being in process. In my testing the actual values from the file load are not able to be passed to the test context. to loading them in scope of the setup operation does support loading in a worker context.

packages/e2e-tests/setup/session-sigs/get-eoa-session-sigs.ts
```diff
@@ -0,0 +1,156 @@
...
+      ? { capabilityAuthSigs: [devEnv.superCapacityDelegationAuthSig] }
+      : {}),
+  } as any); // todo: remove any cast
+
+  log('[getEoaSessionSigs]: ', getEoaSessionSigs);
+
+  return sessionSigs;
+};
+
+export const getEoaSessionSigsWithCapacityDelegations = async (
```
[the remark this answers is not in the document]
joshLong145: it depends on the case. If the resources are not important to the test case then `getEoaSessionSigs` can be used.

packages/e2e-tests/setup/tinny-person.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    const capacityTokenId = (
+      await this.contractsClient?.mintCapacityCreditsNFT({
+        requestsPerKilosecond:
+          this.envConfig.processEnvs.REQUEST_PER_KILOSECOND,
+        daysUntilUTCMidnightExpiration: 2,
+      })
+    )?.capacityTokenIdStr;
+
+    this.contractsClient!.signer! = this.wallet;
```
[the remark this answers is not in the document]
joshLong145: this was added if we are overriding with a new wallet which was created after the spawn of the `Person` instance. so its a mutative op to provide the new wallet which was just in time created.

packages/e2e-tests/src/tests/Delegation.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+    devEnv = global.devEnv;
+  });
+
+  beforeEach(() => {
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
+  });
+
+  afterAll(async () => {
+    //@ts-ignore
+    await global.devEnv.litNodeClient?.disconnect();
```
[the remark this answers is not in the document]
joshLong145: ah no reason I have updated to use the local var to avoid the global access after initial var set.

packages/e2e-tests/src/tests/Delegation.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+                });
+            })();`,
+      jsParams: {
+        dataToSign: alice.loveLetter,
+        publicKey: bob.authMethodOwnedPkp?.publicKey,
+      },
+    });
+
+    devEnv.releasePrivateKeyFromUser(alice);
+    devEnv.releasePrivateKeyFromUser(bob);
```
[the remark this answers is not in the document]
joshLong145: would using the `finally` implementation on `Promise` achieve the same result? Feels a bit cleaner and would handle the case of if the operation which might fails results in a `rejection`

packages/e2e-tests/src/tests/PKPEthers.spec.ts
```diff
@@ -0,0 +1,861 @@
...
+  const alice = await devEnv.createRandomPerson();
+  const pkpSessionSigs = await generator(devEnv, alice);
+
+  const pkpEthersWallet = new PKPEthersWallet({
+    litNodeClient: devEnv.litNodeClient!,
+    pkpPubKey: alice.pkp?.publicKey!,
+    controllerSessionSigs: pkpSessionSigs,
+  });
+
+  await pkpEthersWallet.init();
```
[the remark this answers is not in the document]
joshLong145: If that is the case wouldn't we not want to catch that error and let the test runner report the error? Since we would just have to wait for a new key and restart the test essentially?

packages/e2e-tests/src/tests/PKPEthers.spec.ts
```diff
@@ -0,0 +1,861 @@
...
+  const alice = await devEnv.createRandomPerson();
+  const eoaSessionSigs = await generator(devEnv, alice);
+
+  const pkpEthersWallet = new PKPEthersWallet({
+    litNodeClient: devEnv.litNodeClient!,
+    pkpPubKey: alice.pkp?.publicKey!,
+    controllerSessionSigs: eoaSessionSigs,
+  });
+
+  await pkpEthersWallet.init();
```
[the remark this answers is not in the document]
joshLong145: Wouldn't we want those ops to fail in an obvious way? It seems dangerous to catch those errors as it would then cause the test to enter a separate code path for error handling.

packages/e2e-tests/src/tests/PKPEthers.spec.ts
```diff
@@ -0,0 +1,861 @@
...
+  const alice = await devEnv.createRandomPerson();
+  const eoaSessionSigs = await generator(devEnv, alice);
+
+  const pkpEthersWallet = new PKPEthersWallet({
+    litNodeClient: devEnv?.litNodeClient!,
+    pkpPubKey: alice.pkp?.publicKey!,
+    controllerSessionSigs: eoaSessionSigs,
+  });
+
+  await pkpEthersWallet.init();
```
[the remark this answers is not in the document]
joshLong145: same as above

packages/e2e-tests/src/tests/PKPEthers.spec.ts
```diff
@@ -0,0 +1,861 @@
...
+    expect(recoveredAddress.toLowerCase()).toEqual(
+      alice.pkp?.ethAddress.toLowerCase()
+    );
+  } catch (e) {
+    if ((e as Error).message.includes('insufficient FPE funds')) {
+      console.log(
+        `🧪 PKPEthersWallet should be able to send tx (insufficient FPE funds ❗️)`
+      );
+    } else {
+      throw new Error(`❌ Error: ${(e as Error).toString()}`);
```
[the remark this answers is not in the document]
joshLong145: I have refactored this to use `toThrow` on the `promise rejection` so now we will be expecting a rejected promise with a specifically formed error object. if we do not see the specific `message` in the error from the `Promise` the test will repot a diff in the error `message`

packages/e2e-tests/src/tests/PKPEthers.spec.ts
```diff
@@ -0,0 +1,861 @@
...
+  const eoaSessionSigs = await generator(devEnv, alice);
+
+  const pkpEthersWallet = new PKPEthersWallet({
+    litNodeClient: devEnv?.litNodeClient!,
+    pkpPubKey: alice.pkp?.publicKey!,
+    controllerSessionSigs: eoaSessionSigs,
+  });
+
+  await pkpEthersWallet.init();
+
```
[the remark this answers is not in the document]
joshLong145: Same as above.

Ansonhkg:
LGTM - should be GTG after conflicts are resolved

packages/e2e-tests/src/tests/connection.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+try {
+  jest.setTimeout(60000);
+} catch (e) {
+  // ... continue execution
+}
+
+describe('Connections', () => {
+  beforeEach(() => {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
```
FedericoAmura, line 14:
```suggestion
    jest.spyOn(console, 'warn').mockImplementation(jest.fn());
```

packages/e2e-tests/src/tests/WrappedKeys.spec.ts
```diff
@@ -0,0 +1,1125 @@
...
+    alice = await devEnv.createRandomPerson();
+  });
+
+  afterEach(() => {
+    alice && devEnv.releasePrivateKeyFromUser(alice);
+  });
+
+  beforeEach(() => {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
```
FedericoAmura, line 90:
```suggestion
    jest.spyOn(console, 'warn').mockImplementation(jest.fn());
```

packages/e2e-tests/src/tests/Delegation.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+  });
+
+  afterEach(() => {
+    devEnv.releasePrivateKeyFromUser(alice);
+    devEnv.releasePrivateKeyFromUser(bob);
+  });
+
+  beforeEach(() => {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
```
FedericoAmura, line 35:
```suggestion
    jest.spyOn(console, 'warn').mockImplementation(jest.fn());
```

packages/e2e-tests/src/tests/Delegation.spec.ts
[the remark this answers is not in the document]
FedericoAmura: I think these two calls have to be removed now that `afterEach` is calling these methods

packages/e2e-tests/src/tests/Delegation.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+    // }
+
+    // -- assertions
+    expect(res?.signatures?.sig.r).toBeDefined();
+    expect(res?.signatures?.sig.s).toBeDefined();
+    expect(res?.signatures?.sig.dataSigned).toBeDefined();
+    expect(res?.signatures?.sig.publicKey).toBeDefined();
+
+    // -- signatures.sig.signature must start with 0x
+    expect(res?.signatures.sig.signature.startsWith('0x')).toBeDefined();
```
FedericoAmura, line 127:
```suggestion
    expect(res?.signatures.sig.signature.startsWith('0x')).toBe(true);
```

`false` is also defined

packages/e2e-tests/src/tests/Delegation.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+          })();`,
+        jsParams: {
+          dataToSign: alice.loveLetter,
+          publicKey: bob.pkp?.publicKey,
+        },
+      })
+      .finally(() => {
+        devEnv.releasePrivateKeyFromUser(alice);
+        devEnv.releasePrivateKeyFromUser(bob);
+      });
```
FedericoAmura, line 160: Same as above, we have `afterEach`

packages/e2e-tests/src/tests/Delegation.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+    // }
+
+    // -- assertions
+    expect(res?.signatures?.sig.r).toBeDefined();
+    expect(res?.signatures?.sig.s).toBeDefined();
+    expect(res?.signatures?.sig.dataSigned).toBeDefined();
+    expect(res?.signatures?.sig.publicKey).toBeDefined();
+
+    // -- signatures.sig.signature must start with 0x
+    expect(res?.signatures.sig.signature.startsWith('0x')).toBeDefined();
```
FedericoAmura, line 187:
```suggestion
    expect(res?.signatures.sig.signature.startsWith('0x')).toBe(true);
```

packages/e2e-tests/src/tests/Delegation.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+        sessionSigs: bobsSessionSigs!,
+        toSign: alice.loveLetter,
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-non-null-asserted-optional-chain
+        pubKey: bob.pkp?.publicKey!,
+      })
+      .finally(() => {
+        devEnv.releasePrivateKeyFromUser(alice);
+        devEnv.releasePrivateKeyFromUser(bob);
+      });
+
```
FedericoAmura, line 216: Same as above

packages/e2e-tests/src/tests/PKPEthers.spec.ts
```diff
@@ -0,0 +1,814 @@
...
+    alice = await devEnv.createRandomPerson();
+  });
+
+  afterEach(() => {
+    alice && devEnv.releasePrivateKeyFromUser(alice);
+  });
+
+  beforeEach(() => {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
```
FedericoAmura, line 57:
```suggestion
    jest.spyOn(console, 'warn').mockImplementation(jest.fn());
```

packages/e2e-tests/src/tests/Relayer.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+describe('Relayer', () => {
+  let devEnv: TinnyEnvironment;
+  beforeAll(async () => {
+    //@ts-expect-error global defined
+    devEnv = global.devEnv;
+  });
+
+  beforeEach(() => {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
```
FedericoAmura, line 23:
```suggestion
    jest.spyOn(console, 'warn').mockImplementation(jest.fn());
```

packages/e2e-tests/src/tests/SOLAuthSig.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    alice = await devEnv.createRandomPerson();
+  });
+
+  afterEach(() => {
+    alice && devEnv.releasePrivateKeyFromUser(alice);
+  });
+
+  beforeEach(() => {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
```
FedericoAmura, line 33:
```suggestion
    jest.spyOn(console, 'warn').mockImplementation(jest.fn());
```

packages/tinny/src/lib/tinny-config.ts
```diff
@@ -86,7 +85,7 @@ export interface ProcessEnvs {
    * The URL of Lit RPC server.
    * - If it's running locally on Anvil, it should be 'http://127.0.0.1:8545'
    * - If it's running on Chronicle, it should be 'https://chain-rpc.litprotocol.com/http'
-   * - If it's running on Yellowstone, it should be 'https://yellowstone-rpc.litprotocol.com'
+   * - If it's running on Vesuvius, it should be 'https://vesuvius-rpc.litprotocol.com'
```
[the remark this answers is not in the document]
Ansonhkg: yea it should be `yellowstone`, `vesuvius` is gone!

packages/e2e-tests/setup.config.js
```diff
@@ -0,0 +1,39 @@
...
+
+const NodeEnvironment = require('jest-environment-node').TestEnvironment;
+
+const TinnyEnvironment = require('@lit-protocol/tinny').TinnyEnvironment;
+
+require('dotenv').config();
+console.log('loaded configuration from .env', __dirname);
+
+class CustomEnvironment extends NodeEnvironment {
+  _hasLoadedTinny = false;
```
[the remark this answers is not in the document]
Ansonhkg: It's a `.js` file 😅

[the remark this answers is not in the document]
FedericoAmura: LOL missed that detail!!! 😅

packages/e2e-tests/src/tests/Delegation.spec.ts
joshLong145: Good catch!

packages/tinny/src/lib/tinny-person.ts
```diff
@@ -181,12 +181,12 @@ export class TinnyPerson {
...
+      await this.contractsClient?.mintCapacityCreditsNFT({
         requestsPerKilosecond:
           this.envConfig.processEnvs.REQUEST_PER_KILOSECOND,
         daysUntilUTCMidnightExpiration: 2,
       })
-    ).capacityTokenIdStr;
+    )?.capacityTokenIdStr;
 
     return capacityTokenId;
   }
```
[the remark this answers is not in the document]
joshLong145: our linter flags `!` as dangerous and suggests the `?` to allow for the null coalesce. I have refactored the implementation to use two variables. One for the `capacity increase nft promise` and another for the `token id`. The promise now has a catch which will throw an error if the minting fails. This removes the need for either `!` or `?` syntax.

PR #551, opened 2024-07-22
Feature/lit 3619 js sdk add datil mainnet support

Ansonhkg:
add `datil` production support


local-tests/setup/session-sigs/get-eoa-session-sigs.ts
```diff
@@ -29,9 +29,12 @@ export const getEoaSessionSigs = async (
   person: TinnyPerson,
   resourceAbilityRequests?: LitResourceAbilityRequest[]
 ) => {
-  if (devEnv.litNodeClient.config.litNetwork === LitNetwork.Manzano) {
+  const centralisation =
+    CENTRALISATION_BY_NETWORK[devEnv.litNodeClient.config.litNetwork];
+
+  if (centralisation === 'decentralised') {
```
MaximusHaximus, line 35: Not a blocker for us to land this since it's internal to our SDK, but I think this is more of a coincidental truth than an actual truth; really we probably want an entirely new map tracking e.g. `RLI_TOKEN_REQUIRED_BY_NETWORK`.  Also, when using `LOCALCHAIN` this could be different depending on how the nodes on the localchain are configured so I think we'll need to be able to set it from the environment cc: @joshLong145

local-tests/setup/tinny-environment.ts
```diff
@@ -33,7 +34,10 @@ export class TinnyEnvironment {
     NETWORK: (process.env['NETWORK'] as LIT_TESTNET) || LIT_TESTNET.LOCALCHAIN,
     DEBUG: process.env['DEBUG'] === 'true',
     REQUEST_PER_KILOSECOND:
-      parseInt(process.env['REQUEST_PER_KILOSECOND']) || 200,
+      parseInt(process.env['REQUEST_PER_KILOSECOND']) ||
+      (process.env['NETWORK'] as LIT_TESTNET) === 'datil-dev'
+        ? 1
```
MaximusHaximus, line 39: 1-per-every-10-seconds seems excessively low, no?

local-tests/setup/tinny-environment.ts
```diff
@@ -494,47 +518,53 @@ export class TinnyEnvironment {
...
+      this.superCapacityDelegationAuthSig = (
+        await this.litNodeClient.createCapacityDelegationAuthSig({
+          dAppOwnerWallet: wallet,
+          capacityTokenId: capacityTokenId,
+          // Sets a maximum limit of 200 times that the delegation can be used and prevents usage beyond it
+          uses: '200',
+        })
+      ).capacityDelegationAuthSig;
+    } catch (e: any) {
+      if (e.message.includes(`Can't allocate capacity beyond the global max`)) {
```
MaximusHaximus, line 560: :nit: Is it really ok to always ignore all errors from this code? If this code is necessary, and it fails, then why not exit immediately so we know exactly what went wrong? Ideally we wouldn't continue only to have a later test fail.

local-tests/tests/testUseCustomAuthSessionSigsToPkpSignExecuteJs.ts
```diff
@@ -115,7 +123,7 @@ export const testUseCustomAuthSessionSigsToPkpSignExecuteJs = async (
   } finally {
     devEnv.releasePrivateKeyFromUser(alice);
   }
-
+  process.exit();
```
MaximusHaximus, line 126: Leftover process.exit()?

packages/constants/src/lib/constants/constants.ts
```diff
@@ -979,7 +985,6 @@ export const AUTH_METHOD_TYPE_IDS = {
 
 // ========== PKP Client ==========
 export const PKP_CLIENT_SUPPORTED_CHAINS = ['eth', 'cosmos'];
-export const TELEM_API_URL = 'https://lit-general-worker.getlit.dev';
```
MaximusHaximus, line 982: 💖 woooo

MaximusHaximus (approved):
Just one question that might not be a blocker about 1-req-per-kilosecond being overly low for RLI NFT -- otherwise I don't see any blockers here :+1: NJ <3

local-tests/setup/tinny-environment.ts
Ansonhkg: It doesn't get used anyway for centralised testnet - it's just there to be pretty :))

local-tests/setup/tinny-environment.ts
Ansonhkg: yea - i think it makes sense to just throw the actual error now!

PR #552, opened 2024-07-22
chore: update yarn.lock

Ansonhkg:
# Fix

Move away from `vesuvius` to `yellowstone` for `datil-dev`


Ansonhkg:
switching to https://github.com/LIT-Protocol/js-sdk/pull/550

PR #553, opened 2024-07-23
LIT-3351 - Update wrapped-keys client to support multiple wrapped keys per PKP

MaximusHaximus:
# Description

- Updates the `wrapped-keys` package to allow developers to store and retrieve more than one wrapped key per PKP.
- Added `memo` property to wrapped keys, which is expected to be a user-meaningful label for the key being stored so that the user can differentiate between multiple wrapped keys
- Added `id` property to wrapped keys, which is the unique identifier for a particular wrapped key
    - This property is returned from API methods that are responsible for writing keys to the backend wrapped-keys service
- Added new method to `api` - `listEncryptedKeyMetadata()`.  This method returns an array of wrapped key metadata for the PKP associated with the provided sessionSigs.  Note that the list method does _not_ return `ciphertext` or `dataToEncryptHash` in the array.  Fetching the data required to decrypt a particular key requires calling `getEncryptedKeyMetadata()` with the `id` of the specific key; this is likely only necessary for manual decryption of the key -- the rest of the APIs fetch the metadata and decrypt it where necessary internally, based exclusively on the `id`.
- Existing APIs now require the `id` of the key to be used to perform those API tasks be provided as arguments.  For example, `signMessageWithEncryptedKey()` requires `id` is provided -- it will internally fetch the appropriate encrypted key metadata for the associated wrapped key and provide it to the LIT action for decryption.

### Now, the expected flow for a dapp is:
1. The dapp runs the `listEncryptedKeyMetadata()`, and gets back the array with `{publicKey, memo, id, keyType, litNetwork}[]`

2. The dapp either prompts the user which key to use (`memo` is intended to be user-recognizable identifier for the key), or they could use whatever logic they might have written based on any other arbitrary code -- e.g. If an app is managing _all_ of the wrapped keys for the PKP, it might keep its _own_ hash of ids and purposes, or intuit which key to use based on `keyType`, `publicKey`, or other logic.

3. The dapp then calls the appropriate wrapped keys method (sign message, sign transaction), providing the `id` of the key to be used for that operation.  Internally, the encrypted key data for the given `id` is fetched, and is passed to the LIT action performing the action.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- [X] Updated localtests to use the new function signatures. Ran wrapped keys localtests.

# Checklist:

- [X] I have made corresponding changes to the documentation -- JSDocs updated


DashKash54:
> `listEncryptedKeyMetadata ()`  does not return ciphertext or dataToEncryptHash in the array

Does the Lambda itself not return these info or it's returned but filtered out by the SDK? @MaximusHaximus

DashKash54 (approved):
LGTM!

packages/wrapped-keys/src/lib/constants.ts
```diff
@@ -1,10 +1,10 @@
...
 
-export const CHAIN_ETHEREUM = 'ethereum';
-export const LIT_PREFIX = 'lit_';
+export const CHAIN_ETHEREUM = 'ethereum' as const;
+export const LIT_PREFIX = 'lit_' as const;
 
-export const NETWORK_EVM: Network = 'evm';
-export const NETWORK_SOLANA: Network = 'solana';
+export const NETWORK_EVM: Network = 'evm' as const;
+export const NETWORK_SOLANA: Network = 'solana' as const;
```
Ansonhkg, line 8: What's the benefits of asserting a string a literal type? eg. the type `CHAIN_ETHEREUM` has to be exactly `ethereum`? does it help with passing this const value as a key to other constant values?

Ansonhkg (approved):

MaximusHaximus:
> > `listEncryptedKeyMetadata ()`  does not return ciphertext or dataToEncryptHash in the array
> 
> Does the Lambda itself not return these info or it's returned but filtered out by the SDK? @MaximusHaximus

We run a dynamo Query with an ProjectionExpression that only returns a subset of the data, so it doesn't send it at all.

packages/wrapped-keys/src/lib/constants.ts
MaximusHaximus: so the TS type of NETWORK_EVM as Network is `string`, because Network is a union of `string`s... but the TS type of `NETWORK_EVM` as Network `as const`, is a string _literal_ of `evm`.  `as const` is basically telling TS 'Use the narrowest possible type for this thing'.  It matters when you try to compose objects using these strings or use these strings as indexes into object hashes that use them as keys.

DashKash54:
Just one feedback on testing. Can we add tests for the following scenarios:
1. Trying to add the same Wrapped Key multiple times with **separate** PKP should fail
2. Trying to add multiple Wrapped Keys with the same PKP should pass

@MaximusHaximus

MaximusHaximus:
> Just one feedback on testing. Can we add tests for the following scenarios:
> 
> 1. Trying to add the same Wrapped Key multiple times with **separate** PKP should fail
> 2. Trying to add multiple Wrapped Keys with the same PKP should pass
> 
> @MaximusHaximus

:+1: #1 is already being tested by `testFailImportWrappedKeysWithSamePrivateKey`, I created a ticket to add a test for case #2 - LIT-3689

PR #554, opened 2024-07-23
feat: add custom `nodeProtocol` and enable `https` for port `443`

Ansonhkg:
# Description

- Introducing `nodeProtocol` to customise the HTTP protocol for node connections
- Added a check to determine if port is 443, if so use HTTPS

### DIFF
--- a/packages/constants/src/lib/constants/constants.ts
+++ b/packages/constants/src/lib/constants/constants.ts
@@ -789,8 +789,8 @@ export const METAMASK_CHAIN_INFO_BY_NETWORK: Record<
   localhost: metamaskChainInfo.yellowstone,
 };
 
-const HTTP = 'http://';
-const HTTPS = 'https://';
+export const HTTP = 'http://';
+export const HTTPS = 'https://';
 
 /**
  * Mapping of network values to corresponding http protocol.
@@ -804,8 +804,9 @@ export const HTTP_BY_NETWORK: Record<
   habanero: HTTPS,
   'datil-dev': HTTPS,
   'datil-test': HTTPS,
-  custom: HTTP,
-  localhost: HTTP,
+  internalDev: HTTPS, 
+  custom: HTTP, // default, can be changed by config
+  localhost: HTTP, // default, can be changed by config
 };
 
 /**
--- a/packages/constants/src/lib/constants/constants.ts
+++ b/packages/constants/src/lib/constants/constants.ts
@@ -804,7 +804,7 @@ export const HTTP_BY_NETWORK: Record<
   habanero: HTTPS,
   'datil-dev': HTTPS,
   'datil-test': HTTPS,
-  internalDev: HTTPS, 
+  internalDev: HTTPS,
   custom: HTTP, // default, can be changed by config
   localhost: HTTP, // default, can be changed by config
 };
--- a/packages/contracts-sdk/src/lib/contracts-sdk.ts
+++ b/packages/contracts-sdk/src/lib/contracts-sdk.ts
@@ -56,6 +56,8 @@ import {
   HTTP_BY_NETWORK,
   CENTRALISATION_BY_NETWORK,
   LIT_NETWORK,
+  HTTP,
+  HTTPS,
 } from '@lit-protocol/constants';
 import { LogManager, Logger } from '@lit-protocol/logger';
 import { computeAddress } from 'ethers/lib/utils';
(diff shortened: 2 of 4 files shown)


packages/constants/src/lib/constants/constants.ts
```diff
@@ -804,8 +804,9 @@ export const HTTP_BY_NETWORK: Record<
   habanero: HTTPS,
   'datil-dev': HTTPS,
   'datil-test': HTTPS,
-  custom: HTTP,
-  localhost: HTTP,
+  internalDev: HTTPS,
+  custom: HTTP, // default, can be changed by config
+  localhost: HTTP, // default, can be changed by config
```
joshLong145, line 809: Nice, do we think wrapping the `PROTO` type in an object for future config values is worth it at this time?

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -888,7 +890,8 @@ export class LitContracts {
   public static getValidators = async (
     network: LIT_NETWORKS_KEYS,
     context?: LitContractContext | LitContractResolverContext,
-    rpcUrl?: string
+    rpcUrl?: string,
+    nodeProtocol?: typeof HTTP | typeof HTTPS | null
```
joshLong145, line 894: Should we pick a proto to default to instead of allowing null?

joshLong145 (approved):

packages/contracts-sdk/src/lib/contracts-sdk.ts
Ansonhkg: we already have a default here:
https://github.com/LIT-Protocol/js-sdk/pull/554/files#diff-2dcc2c94e7788b1bbaef5e6cf9158025a2580d238f807bd3cb923bbe355eb47bR958

`nodeProtocol` should be null by default, as we don't really want other to set except the internal devs

packages/constants/src/lib/constants/constants.ts
Ansonhkg: the only type right now is `443` right? not sure if we have `80` though as they all range from 7xxx to 8xxx

joshLong145: We might want to add a port override for local testing. This could become handy with future tooling.

joshLong145: Actually think just having the single property is okay for now. But I think we will need to think about overrides and how we handle them broadly speaking.

packages/core/src/lib/lit-core.ts
```diff
@@ -99,6 +99,8 @@ export type LitNodeClientConfigWithDefaults = Required<
     Pick<LitNodeClientConfig, 'storageProvider' | 'contractContext' | 'rpcUrl'>
   > & {
     bootstrapUrls: string[];
+  } & {
+    nodeProtocol?: `http://` | `https://` | null;
```
FedericoAmura, line 103: Can we use constants here?

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -941,13 +944,24 @@ export class LitContracts {
...
+      let port = item.port;
+
+      // Determine the protocol to use based on various conditions
+      const protocol =
+        // If nodeProtocol is defined, use it
+        nodeProtocol ||
+        // If port is 443, use HTTPS, otherwise use network-specific HTTP
+        (port === 443 ? HTTPS : HTTP_BY_NETWORK[network]) ||
+        // Fallback to HTTP if no other conditions are met
+        HTTP;
```
FedericoAmura, line 960: There is no port+protocol check? Is it valid that a validator uses https on port 80 or http on port 443?
Maybe forcing https on port 443 to at least avoid that

FedericoAmura (approved):

packages/contracts-sdk/src/lib/contracts-sdk.ts
Ansonhkg: It’s a good point. AFAIK. a validator might use `https` on port `80`, and I actually wanted to force `https` on port `443`. However, I think we should still provide an option to override the default value using nodeProtocol (from an internal developer’s perspective).

packages/core/src/lib/lit-core.ts
Ansonhkg: yea, gotta be `typeof <constant_value>`

PR #563, opened 2024-07-25
feat: extract `formatPKPResource` and add unit tests

Ansonhkg:
# Description

Extract the function out and add unit tests for this https://github.com/LIT-Protocol/js-sdk/pull/562

---
1. Takes out '0x' and makes the string 64 characters long.
2. Adds zeros to make short strings 64 characters.
3. Don't change valid 64-character hex strings.
4. Returns '*' as is.
5. Returns the original if it has bad hex characters.
6. Doesn't change 64-character strings.
7. Adds zeros to make short hex strings 64 characters.
8. Returns the original if it partly matches hex.
9. Throws an error if the string is too long.

### DIFF
--- a/packages/auth-helpers/src/lib/utils.ts
+++ b/packages/auth-helpers/src/lib/utils.ts
@@ -10,7 +10,7 @@
  * - Adds zeros to make short hex strings 64 characters.
  * - Returns the original if it partly matches hex.
  * - Throws an error if the string is too long.
- * 
+ *
  * @param resource The identifier for the resource. This should be the PKP token ID.
  * @returns A 32-byte hex string representing the resource ID.
  * @throws Will throw an error if the input exceeds 64 characters.
@@ -24,10 +24,10 @@ export function formaPKPResource(resource: string): string {
     throw new Error('Resource ID exceeds 64 characters (32 bytes) in length.');
   }
 
-/**
- * The pattern matches any sequence of 6 characters that are
- * either digits (0-9) or letters A-F (both uppercase and lowercase).
- */
+  /**
+   * The pattern matches any sequence of 6 characters that are
+   * either digits (0-9) or letters A-F (both uppercase and lowercase).
+   */
   const hexRegex = /^[0-9A-Fa-f]+$/;
 
   // Ensure the resource is a valid hex string and not a wildcard '*'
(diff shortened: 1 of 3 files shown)


glitch003:
very fast sir

packages/auth-helpers/src/lib/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+export function formaPKPResource(resource: string): string {
+  // Remove the '0x' prefix if present
+  let fixedResource = resource.startsWith('0x') ? resource.slice(2) : resource;
+
+  // Throw an error if the resource length exceeds 64 characters
+  if (fixedResource.length > 64) {
+    throw new Error('Resource ID exceeds 64 characters (32 bytes) in length.');
+  }
+
+  // Regex to validate hex strings
```
Ansonhkg, line 16:
```suggestion
/**
 * The pattern matches any sequence of 6 characters that are
 * either digits (0-9) or letters A-F (both uppercase and lowercase).
 */
```

packages/auth-helpers/src/lib/utils.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Formats the resource ID to a 32-byte hex string.
+ * @param resource The identifier for the resource. This should be the PKP token ID.
+ * @returns A 32-byte hex string representing the resource ID.
+ * @throws Will throw an error if the input exceeds 64 characters.
+ */
```
Ansonhkg, line 6:
```suggestion
/**
 * Formats the resource ID to a 32-byte hex string.
 *
 * - Takes out '0x' and makes the string 64 characters long.
 * - Adds zeros to make short strings 64 characters.
 * - Doesn't change valid 64-character hex strings.
 * - Returns '*' as is.
 * - Returns the original if it has bad hex characters.
 * - Doesn't change 64-character strings.
 * - Adds zeros to make short hex strings 64 characters.
 * - Returns the original if it partly matches hex.
 * - Throws an error if the string is too long.
 * 
 * @param resource The identifier for the resource. This should be the PKP token ID.
 * @returns A 32-byte hex string representing the resource ID.
 * @throws Will throw an error if the input exceeds 64 characters.
 */
```

PR #567, opened 2024-07-29
feat: replace hardcoded gas limit with calculated one including margin

FedericoAmura:
# Description

This PR removes the hardcoded `GAS_LIMIT` constant in contracts sdk used by default when user did not provide one and replaced it with one that increases the limit by certain margin when it is a minting transaction

## Type of change

# How Has This Been Tested?

- [x] Used Ansh demo app that to mint a PKP with contracts-sdk directly (not using relay)
Screenshot from 2024-07-29 21-09-01
 

# Checklist:

### DIFF
--- a/packages/contracts-sdk/src/lib/contracts-sdk.ts
+++ b/packages/contracts-sdk/src/lib/contracts-sdk.ts
@@ -13,7 +13,7 @@ import {
   MintWithAuthResponse,
 } from '@lit-protocol/types';
 import bs58 from 'bs58';
-import { BytesLike, ContractReceipt, ethers } from 'ethers';
+import { BigNumberish, BytesLike, ContractReceipt, ethers } from 'ethers';
 import { decToHex, hexToDec, intToIP } from './hex2dec';
 
 // ----- autogen:import-data:start  -----
@@ -103,7 +103,7 @@ declare global {
   }
 }
 
-const GAS_LIMIT = ethers.utils.hexlify(5000000); // Adjust as needed
+const GAS_LIMIT_INCREASE_PERCENTAGE = 10;
 
 // This code defines a LitContracts class that acts as a container for a collection of smart contracts. The class has a constructor that accepts an optional args object with provider and rpc properties. If no provider is specified, the class will create a default provider using the specified rpc URL. If no rpc URL is specified, the class will use a default URL.
 // The class has a number of properties that represent the smart contract instances, such as accessControlConditionsContract, litTokenContract, pkpNftContract, etc. These smart contract instances are created by passing the contract address, ABI, and provider to the ethers.Contract constructor.
@@ -948,7 +948,7 @@ export class LitContracts {
 
       // Convert the integer IP to a string format
       const ip = intToIP(item.ip);
-      let port = item.port;
+      const port = item.port;
 
       // Determine the protocol to use based on various conditions
       const protocol =
@@ -1014,6 +1014,7 @@ export class LitContracts {
    * @param scopes - The permission scopes.
    * @param pubkey - The public key.
    * @param authMethodId - (optional) The authentication ID.
+   * @param gasLimit - (optional) The gas limit.
    * @returns An object containing the PKP information and the transaction receipt.
    * @throws Error if the contracts are not connected, the contract is not available, authMethodType or accessToken is missing, or permission scopes are required.
    */
@@ -1061,7 +1062,7 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
     const _pubkey = pubkey ?? '0x';
 
     // if scopes are list of strings, turn them into numbers
-    scopes = scopes.map((scope: any) => {
+    const _scopes = scopes.map((scope) => {
       if (typeof scope === 'string') {
         return ethers.BigNumber.from(scope);
       }
@@ -1077,18 +1078,37 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
     // -- go
     const mintCost = await this.pkpNftContract.read.mintCost();
 
+    let _gasLimit: BigNumberish | undefined = gasLimit;
+    if (!_gasLimit) {
+      const txData = await this.pkpHelperContract.write.populateTransaction.mintNextAndAddAuthMethods(
+        2, // key type
+        [authMethod.authMethodType],
+        [_authMethodId],
+        [_pubkey],
+        [[..._scopes]],
+        true,
+        true,
+        { value: mintCost }
+      );
+
+      const gasEstimation = await this.signer.estimateGas(txData);
+      const adjustedGasLimit = gasEstimation.mul(100 + GAS_LIMIT_INCREASE_PERCENTAGE).div(100);
+
+      _gasLimit = adjustedGasLimit;
+    }
+
     // -- start minting
     const tx = await this.pkpHelperContract.write.mintNextAndAddAuthMethods(
       2, // key type
       [authMethod.authMethodType],
       [_authMethodId],
       [_pubkey],
-      [[...scopes]],
+      [[..._scopes]],
       true,
       true,
       {
         value: mintCost,
-        gasLimit: gasLimit || GAS_LIMIT, // Adjust as needed
+        gasLimit: _gasLimit,
       }
     );
     const receipt = await tx.wait();
@@ -1117,7 +1137,7 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
         break;
       }
       tries++;
-      await new Promise((resolve, _reject) => {
+      await new Promise((resolve) => {
         setTimeout(resolve, 10_000);
       });
     }
@@ -1361,7 +1381,7 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
     try {
       const res = await this.rateLimitNftContract.write.mint(expiresAt, {
         value: mintCost,
-        gasLimit: gasLimit || GAS_LIMIT,
+        gasLimit: gasLimit,
       });
 
       const txHash = res.hash;
@@ -1616,35 +1636,35 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
           throw new Error('Could not get mint cost');
         }
 
-        let sentTx;
-
         if (this.isPKP) {
           this.log(
             "This is a PKP wallet, so we'll use the PKP wallet to sign the tx"
           );
-
-          this.log('...populating tx');
-          const tx =
-            await this.pkpNftContract.write.populateTransaction.mintNext(2, {
-              value: mintCost,
-              gasLimit: param?.gasLimit || GAS_LIMIT,
-            });
-          this.log('tx:', tx);
-
-          this.log('...signing tx');
-          const signedTx = await this.signer.signTransaction(tx);
-          this.log('signedTx:', signedTx);
-
-          this.log('sending signed tx...');
-          sentTx = await this.signer.sendTransaction(
-            signedTx as ethers.providers.TransactionRequest
-          );
-        } else {
-          sentTx = await this.pkpNftContract.write.mintNext(2, {
-            value: mintCost,
-            gasLimit: param?.gasLimit || GAS_LIMIT,
-          });
         }
+
+        this.log('...populating tx');
+        const tx =
+          await this.pkpNftContract.write.populateTransaction.mintNext(2, {
+            value: mintCost,
+            gasLimit: param?.gasLimit,
+          });
+
+        if (!param?.gasLimit) {
+          const gasEstimation = await this.signer.estimateGas(tx);
+          const adjustedGasLimit = gasEstimation.mul(100 + GAS_LIMIT_INCREASE_PERCENTAGE).div(100);
+          tx.gasLimit = adjustedGasLimit;
+        }
+
+        this.log('tx:', tx);
+
+        this.log('...signing tx');
+        const signedTx = await this.signer.signTransaction(tx);
+        this.log('signedTx:', signedTx);
+
+        this.log('sending signed tx...');
+        const sentTx = await this.signer.sendTransaction(
+          signedTx as ethers.providers.TransactionRequest
+        );
 
         this.log('sentTx:', sentTx);
 
@@ -1667,7 +1687,7 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
             break;
           }
           tries++;
-          await new Promise((resolve, _reject) => {
+          await new Promise((resolve) => {
             setTimeout(resolve, 10_000);
           });
         }
@@ -2394,6 +2414,24 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
       }: MintNextAndAddAuthMethods): Promise<any> => {
         // first get mint cost
         const mintCost = await this.pkpNftContract.read.mintCost();
+
+        let _gasLimit: BigNumberish | undefined = gasLimit;
+        if (!_gasLimit) {
+          const txData = await this.pkpHelperContract.write.populateTransaction.mintNextAndAddAuthMethods(
+            keyType,
+            permittedAuthMethodTypes,
+            permittedAuthMethodIds as BytesLike[],
+            permittedAuthMethodPubkeys as BytesLike[],
+            permittedAuthMethodScopes,
+            addPkpEthAddressAsPermittedAddress,
+            sendPkpToItself,
+            { value: mintCost }
+          );
+          const gasEstimation = await this.signer.estimateGas(txData);
+          const adjustedGasLimit = gasEstimation.mul(100 + GAS_LIMIT_INCREASE_PERCENTAGE).div(100);
+          _gasLimit = adjustedGasLimit;
+        }
+
         const tx = await this.pkpHelperContract.write.mintNextAndAddAuthMethods(
           keyType,
           permittedAuthMethodTypes,
@@ -2404,7 +2442,7 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
           sendPkpToItself,
           {
             value: mintCost,
-            gasLimit: gasLimit || GAS_LIMIT,
+            gasLimit: _gasLimit,
           }
         );
         return tx;
--- a/packages/contracts-sdk/src/lib/contracts-sdk.ts
+++ b/packages/contracts-sdk/src/lib/contracts-sdk.ts
@@ -105,7 +105,9 @@ declare global {
 
 // Due to the usage of arbitrum stylus contracts the gas limit is increased by 10% to avoid reverts due to out of gas errors
 const GAS_LIMIT_INCREASE_PERCENTAGE = 10;
-const GAS_LIMIT_ADJUSTMENT = ethers.BigNumber.from(100).add(GAS_LIMIT_INCREASE_PERCENTAGE);
+const GAS_LIMIT_ADJUSTMENT = ethers.BigNumber.from(100).add(
+  GAS_LIMIT_INCREASE_PERCENTAGE
+);
 
 // This code defines a LitContracts class that acts as a container for a collection of smart contracts. The class has a constructor that accepts an optional args object with provider and rpc properties. If no provider is specified, the class will create a default provider using the specified rpc URL. If no rpc URL is specified, the class will use a default URL.
 // The class has a number of properties that represent the smart contract instances, such as accessControlConditionsContract, litTokenContract, pkpNftContract, etc. These smart contract instances are created by passing the contract address, ABI, and provider to the ethers.Contract constructor.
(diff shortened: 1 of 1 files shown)


Ansonhkg (changes requested):
Looking good! We also need to add gasLimit to pkpPermissions & rateLimit contracts. Let's create a helper function to do the adjusted gas limit calculation? as we are re-using it and multiple places now, and add an unit test on it.

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -2436,4 +2463,17 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
       // },
     },
   };
+
+  private estimateGasWithMargin = async (
```
Ansonhkg, line 2467: we have a new naming convention for our code space to have all private methods prefixed with `_`

```suggestion
  private _estimateGasWithMargin = async (
```

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1077,18 +1078,34 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
...
+          2, // key type
+          [authMethod.authMethodType],
+          [_authMethodId],
+          [_pubkey],
+          [[..._scopes]],
+          true,
+          true,
+          { value: mintCost }
+        );
+      _gasLimit = await this.estimateGasWithMargin(txData);
```
Ansonhkg, line 1094:
```suggestion
      _gasLimit = await this._estimateGasWithMargin(txData);
```

Ansonhkg (changes requested):

FedericoAmura:
> Looking good! We also need to add gasLimit to pkpPermissions & rateLimit contracts. Let's create a helper function to do the adjusted gas limit calculation? as we are re-using it and multiple places now, and add an unit test on it.

gas limit is set automatically by ethers when sending the transaction, so unless it needs the adjustment, it is not needed. Only minting needed this adjustment

Adding a unit test for it defeats the purpose as it would make the function `public` (undesirable if we want to reduce our public api) and it ultimately needs to communicate with blockchain so it would need an integration test to be really tested automatically and will still depend on factors we cannot control

Ansonhkg:
> > Looking good! We also need to add gasLimit to pkpPermissions & rateLimit contracts. Let's create a helper function to do the adjusted gas limit calculation? as we are re-using it and multiple places now, and add an unit test on it.
> 
> gas limit is set automatically by ethers when sending the transaction, so unless it needs the adjustment, it is not needed. Only minting needed this adjustment
> 
> Adding a unit test for it defeats the purpose as it would make the function `public` (undesirable if we want to reduce our public api) and it ultimately needs to communicate with blockchain so it would need an integration test to be really tested automatically and will still depend on factors we cannot control

re: gas limit is set automatically by ethers
yea, but the problem is not ethers, but we might have some issues from the Arbitrum Stylus

re: unit test
I was referring to this little portion of the code, but not the communication layer, but we're fine without it. 

```
const adjustedGasLimit = gasEstimation
      .mul(100 + porcentualIncrease)
      .div(100);
``

File: packages/contracts-sdk/src/lib/contracts-sdk.ts [hunk 3]
```diff
@@ -1361,7 +1378,7 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
     try {
       const res = await this.rateLimitNftContract.write.mint(expiresAt, {
         value: mintCost,
-        gasLimit: gasLimit || GAS_LIMIT,
+        gasLimit: gasLimit,
```
MaximusHaximus, line 1381: Should this call also be using our new `estimateGasWithMargin()`? Do we not want dynamic gas limit calculation on every call? If we're intentionally not using that logic here, a comment would be a good idea describing _why_ we explicitly left this as-is

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1616,36 +1633,29 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
...
+        if (!_gasLimit) {
           const tx =
             await this.pkpNftContract.write.populateTransaction.mintNext(2, {
               value: mintCost,
-              gasLimit: param?.gasLimit || GAS_LIMIT,
             });
-          this.log('tx:', tx);
-
-          this.log('...signing tx');
-          const signedTx = await this.signer.signTransaction(tx);
```
MaximusHaximus, line 1635: Is the new code functionally still the same after removing the explicit code here that calls an explicit `signTransaction` and `sendTransaction()` using `this.signer` and skip straight to calling `mintNext()` without signing it explicitly using `this.signer`?

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1667,7 +1677,7 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
             break;
           }
           tries++;
-          await new Promise((resolve, _reject) => {
+          await new Promise((resolve) => {
```
MaximusHaximus, line 1680: Not a blocker for this PR, as I know this wasn't changed in this PR, but it occurs to me that on the new chain we might be able to change this to retry a lot sooner than 10 seconds 🤔

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1077,18 +1078,34 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
     // -- go
     const mintCost = await this.pkpNftContract.read.mintCost();
 
+    let _gasLimit: BigNumberish | undefined = gasLimit;
```
MaximusHaximus, line 1081: Rather than needing to duplicate all of the transaction params and also needing a local mutable variable to store the `_gasLimit` value, it would be nice to get this defined as a self-contained function -- otherwise we are likely to end up changing the tx parameters in one spot but not the other by mistake, when we really need them to always be the exact same for the estimateGas() to be accurate

```javascript
async function _mintNextAndAddAuthMethodsDynamicGasLimit({ gasLimit, mintCost, txParams }) {
  return this.pkpHelperContract.write.mintNextAndAddAuthMethods(...txParams, {
    gasLimit:
      gasLimit ||
      (await this.pkpHelperContract.write.populateTransaction.mintNextAndAddAuthMethods(
        ...txParams,
        { value: mintCost }
      )),
    value: mintCost,
  });
}

```

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -2394,6 +2404,23 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
       }: MintNextAndAddAuthMethods): Promise<any> => {
         // first get mint cost
         const mintCost = await this.pkpNftContract.read.mintCost();
+
+        let _gasLimit: BigNumberish | undefined = gasLimit;
+        if (!_gasLimit) {
+          const txData =
+            await this.pkpHelperContract.write.populateTransaction.mintNextAndAddAuthMethods(
```
MaximusHaximus, line 2411: As above, it would be good to get this defined as its own function rather than having to define all of the transaction arguments twice and use a local variable to store gasLImit

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -2436,4 +2463,17 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
...
+    porcentualIncrease: number = GAS_LIMIT_INCREASE_PERCENTAGE
+  ) => {
+    const gasEstimation = await this.pkpNftContract.write.provider.estimateGas(
+      txData
+    );
+    const adjustedGasLimit = gasEstimation
+      .mul(100 + porcentualIncrease)
+      .div(100);
+    return adjustedGasLimit;
+  };
```
MaximusHaximus, line 2478: Can we change this to use english (percentage) for consistency w/ the constant? :)

```suggestion
  private _estimateGasWithMargin = async (
    txData: ethers.PopulatedTransaction,
    percentageIncrease: number = GAS_LIMIT_INCREASE_PERCENTAGE
  ) => {
    const gasEstimation = await this.pkpNftContract.write.provider.estimateGas(
      txData
    );
    const adjustedGasLimit = gasEstimation
      .mul(100 + percentageIncrease)
      .div(100);
    return adjustedGasLimit;
  };
```

MaximusHaximus (changes requested):
Rock on, this will be a big reliability improvement in dynamic load situations.  I added a few questions and suggestions in-line :+1:

packages/contracts-sdk/src/lib/contracts-sdk.ts
FedericoAmura: I added the margin to all calls as @Ansonhkg suggested
In particular does not seem to be a bad thing as any excess gas will be returned

packages/contracts-sdk/src/lib/contracts-sdk.ts
FedericoAmura: Seems like this could be moved into a chain configuration constant

packages/contracts-sdk/src/lib/contracts-sdk.ts
FedericoAmura: Great suggestion. I  updated the generic function to receive the contract, method, params, overrides and all needed things, I believe that approach is more readable

packages/contracts-sdk/src/lib/contracts-sdk.ts
FedericoAmura: Responded above

packages/contracts-sdk/src/lib/contracts-sdk.ts
FedericoAmura: Not that it is a new code, it is unifying both paths. Now that there might be a intermediate step, the code ended up being identical for both so there was no more need for the `if`

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -2436,4 +2442,26 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
...
+  ) => {
+    if (!overrides.gasLimit) {
+      const txData = await contract.populateTransaction[method](
+        ...args,
+        overrides
+      );
+      const gasEstimation = await contract.provider.estimateGas(txData);
+      const adjustedGasLimit = gasEstimation
+        .mul(100 + percentageIncrease)
+        .div(100);
```
MaximusHaximus, line 2461: (Nit, since this is functional as written)

The addition-to-100 followed by divide-by-100 is a bit of a mind-bender.  If we're going to store the percentage as an integer rather than a ratio, something a bit more explicit would be easier to follow

```suggestion
      // Gas estimation is not reliable so we add X% to the stimate to ensure the tx is successful
      const extraGas = gasEstimation.div(percentageIncrease); 
      const adjustedGasLimit = gasEstimation.add(extraGas);
```

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -2436,4 +2442,26 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
...
+      overrides.gasLimit = adjustedGasLimit;
```
MaximusHaximus, line 2462: This is kind of a :nit: based on how the function is _currently_ being used, but in general it would be preferable to avoid mutating arguments passed due to the law of unintended consequences/don't surprise me principal

packages/contracts-sdk/src/lib/contracts-sdk.ts
MaximusHaximus: I'm just confused in that I don't see this call to `signTransaction` anywhere in the code path anymore; it seems like now we always run the code from the `!this.isPKP` branch instead?

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1696,17 +1681,22 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
       claimAndMint: async (
         derivedKeyId: BytesLike,
         signatures: IPubkeyRouter.SignatureStruct[],
-        txOpts?: any
+        txOpts: ethers.PayableOverrides & { from?: string } = {}
```
MaximusHaximus, line 1684: 💖  Types ftw

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1696,17 +1681,22 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
...
       ) => {
         try {
-          const cost = await this.pkpNftContract.read.mintCost();
-          const tx = await this.pkpNftContract.write.claimAndMint(
-            2,
-            derivedKeyId,
-            signatures,
-            txOpts ?? { value: cost }
+          if (!txOpts.value) {
```
MaximusHaximus, line 1687: Would prefer if we didn't mutate `txOpts` since it could have been passed in, and setting the `value` could result in an unexpected side-effect if those opts were re-used for another call by the caller.

MaximusHaximus (changes requested):
💖 It warms my heart to see you making the code better from a documentation and types perspective in these PRs m8 <3 Just one more couple of questions/concerns and I think this is ready to rock

packages/contracts-sdk/src/lib/contracts-sdk.ts
FedericoAmura: On this particular case: This is also fixing the issue where the transaction would fail when user passes something for `txOpts` but without `cost` in it. The function would calculate it but not pass it to the tx, that required it

Anyway, I get your point and added an intermediate variable for all that

packages/contracts-sdk/src/lib/contracts-sdk.ts
FedericoAmura: Good call. Fixed with an internal _overrides

packages/contracts-sdk/src/lib/contracts-sdk.ts
FedericoAmura: This is less nitty than it seems. I just copied the way @glitch003 did it in trusting that would work, but there is an edge case that blows this up: `ethers.BigNumber` throw `underflow` when dealing with fractional numbers

So, we cannot do `gasEstimation.mul(1.1)` but we also cannot do `gasEstimation.mul(100+0.1).div(100)` which was an option with `typeof percentageIncrease === number`

I changed it to be directly defined as `BigNumber` in params and the constant at the top. I think that will be safer

packages/contracts-sdk/src/lib/contracts-sdk.ts
FedericoAmura: `!isPKP` path was populating, signing and sending the transaction directly in one step (line 1643)
`isPKP` path was populating the transaction (line 1627) then signing (line 1635) and sending (line 1639) it with the embedded signer

With the changes for this tkt, `!isPKP` path had to be separated into different steps, becoming almost identical (using the contract interfaces instead of the embedded signer/provider)

But the contracts are initialized with the embedded `signer` and `provider` [here](https://github.com/LIT-Protocol/js-sdk/blob/7327abc9b4cefec1e4bfa2fbbc6efdcaf1c87b77/packages/contracts-sdk/src/lib/contracts-sdk.ts#L484) so both paths really became identical just with different references

The only thing that changes when `isPKP === true` is [the provider](https://github.com/LIT-Protocol/js-sdk/blob/7327abc9b4cefec1e4bfa2fbbc6efdcaf1c87b77/packages/contracts-sdk/src/lib/contracts-sdk.ts#L391) and is then passed to contracts anyway

Ansonhkg (commented):
Suggest changes like this:
https://github.com/LIT-Protocol/js-sdk/pull/583

Feel free to merge if agreed

MaximusHaximus (changes requested):
Just one tweak to types and a spot we're still mutating provided arguments that we should clean up -- otherwise I think this look GTG

packages/contracts-sdk/src/lib/contracts-sdk.ts
MaximusHaximus: :+1: rock on man. I like it.

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1696,17 +1685,24 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
...
+          if (!overrides.value) {
+            const cost = await this.pkpNftContract.read.mintCost();
+            overrides.value = cost;
+          }
+          const tx = await this._callWithAdjustedOverrides(
+            this.pkpNftContract.write,
+            'claimAndMint',
+            [2, derivedKeyId, signatures],
+            overrides
           );
```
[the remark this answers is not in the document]
FedericoAmura: Chopped

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -2436,4 +2448,36 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
...
+  ) => {
+    const gasLimit = await contract.estimateGas[method](...args, overrides);
+    return gasLimit.mul(gasLimitAdjustment);
+  };
+
+  private _callWithAdjustedOverrides = async (
+    contract: ethers.Contract,
+    method: string,
+    args: unknown[],
+    overrides: ethers.PayableOverrides & { from?: string } = {},
```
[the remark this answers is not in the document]
FedericoAmura: Replaced

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -2436,4 +2448,36 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
       // },
     },
   };
+
+  private _getAdjustedGasLimit = async (
+    contract: ethers.Contract,
+    method: string,
+    args: unknown[],
+    overrides: ethers.PayableOverrides & { from?: string } = {},
```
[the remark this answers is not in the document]
FedericoAmura: Replaced

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -2274,11 +2281,12 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
     },
     write: {
       mint: async ({
-        mintCost,
+        txOpts,
         timestamp,
       }: {
-        mintCost: {
-          value: any;
+        txOpts: {
```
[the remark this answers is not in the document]
FedericoAmura: Replaced

Ansonhkg (approved):
Let's go!

MaximusHaximus (approved):
:+1: Nice work man. 💖 🚢

glitch003:
hey guys - this PR caused the tests to start failing on master with an error related to gas being too high - https://github.com/LIT-Protocol/js-sdk/actions/runs/10374454818/job/28735639774



it seems like the tests passed before it was merged, I wonder what happened.

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -2436,4 +2443,56 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
...
+    method: K,
+    args: Parameters<T['functions'][K]>,
+    overrides: ethers.CallOverrides = {},
+    gasLimitAdjustment: ethers.BigNumber = GAS_LIMIT_ADJUSTMENT
+  ): Promise<ethers.BigNumber> => {
+    const gasLimit = await contract.estimateGas[method as string](
+      ...args,
+      overrides
+    );
+    return gasLimit.mul(gasLimitAdjustment);
```
glitch003, line 2461: this multiplies the gas by 10, not by 10%.  going to make a PR to fix.

glitch003:
ok i found it.  we're mutliplying gas by 10 instead of 10 percent.  i'll put in a PR.  the error was `intrinsic gas too high -- tx.gas_limit > env.block.gas_limit` which makes sense, since it costs like 3.5m gas to mint a PKP and block gas limit is 30m and 3.5m * 10 = 35m which is higher than the block limit

packages/contracts-sdk/src/lib/contracts-sdk.ts
glitch003: ohh i think it actually multiplies it by 110?  

```
const GAS_LIMIT_INCREASE_PERCENTAGE = 10;
const GAS_LIMIT_ADJUSTMENT = ethers.BigNumber.from(100).add(
  GAS_LIMIT_INCREASE_PERCENTAGE
);
```

so GAS_LIMIT_ADJUSTMENT is 100 + 10 = 110.

PR #571, opened 2024-07-31
AuthSig + getSessionSigs Description Updates

awisniew207:
Added:
`AuthSig` properties explanation
`getSessionSigs` return explanation
Additional `getSessionSigs` properties explanation

Comments always welcome!


packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -1971,36 +1971,22 @@ export class LitNodeClientNodeJs
   };
 
   /**
-   * Get session signatures for a set of resources
+   * This is one of the methods that can be used to get session signatures.
```
spacesailor24, line 1974: Maybe change this to:

`Get session signatures for a set of [Lit resources](https://v6-api-doc-lit-js-sdk.vercel.app/interfaces/types_src.ILitResource.html#resource)`

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -1971,36 +1971,22 @@ export class LitNodeClientNodeJs
...
 
   /**
-   * Get session signatures for a set of resources
+   * This is one of the methods that can be used to get session signatures. 
    *
-   * High level, how this works:
-   * 1. Generate or retrieve session key
-   * 2. Generate or retrieve the wallet signature of the session key
+   * How this function works on a high level:
+   * 1. Generate or retrieve session keys (a public and private key pair)
```
spacesailor24, line 1977: Link to https://v6-api-doc-lit-js-sdk.vercel.app/interfaces/types_src.SessionKeyPair.html

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -1971,36 +1971,22 @@ export class LitNodeClientNodeJs
...
+   * 2. Generate or retrieve the `AuthSig` that specifies the session capabilities and resources
```
spacesailor24, line 1978: https://v6-api-doc-lit-js-sdk.vercel.app/interfaces/types_src.AuthSig.html

packages/types/src/lib/interfaces.ts
```diff
@@ -33,11 +33,37 @@ export interface AccsOperatorParams {
 
 /** ---------- Auth Sig ---------- */
 
+/**
+ * An `AuthSig` represents a cryptographic proof of ownership for an Ethereum address, created by signing a standardized [ERC-5573 SIWE](https://eips.ethereum.org/EIPS/eip-5573) (Sign-In with Ethereum) message. This signature serves as a verifiable credential, allowing the Lit network to associate specific permissions, access rights, and operational parameters with the signing Ethereum address. By incorporating various capabilities, resources, and parameters into the SIWE message before signing, the resulting `AuthSig` effectively defines and communicates these authorizations and specifications for the address within the Lit network.
```
spacesailor24, line 37: 🔥

packages/types/src/lib/interfaces.ts
```diff
@@ -33,11 +33,37 @@ export interface AccsOperatorParams {
...
+ */
 export interface AuthSig {
+
+  /**
+   * The signature produced by signing the `signMessage` property with the `address` property.
```
spacesailor24, line 42:
```suggestion
   * The signature produced by signing the `signMessage` property with the corresponding private key for the `address` property. 
```

packages/types/src/lib/interfaces.ts
```diff
@@ -33,11 +33,37 @@ export interface AccsOperatorParams {
...
+   */
   sig: any;
+
+  /**
+   * The method used to derive the signature (e.g, `web3.eth.personal.sign`).
+   */
   derivedVia: string;
+
+  /**
+   * A SIWE (Sign-In with Ethereum) message. This can be prepared by using one of the `createSiweMessage` functions from the [`@auth-helpers`](https://v6-api-doc-lit-js-sdk.vercel.app/modules/auth_helpers_src.html) package:
```
spacesailor24, line 52:
```suggestion
   * An [ERC-5573](https://eips.ethereum.org/EIPS/eip-5573) SIWE (Sign-In with Ethereum) message. This can be prepared by using one of the `createSiweMessage` functions from the [`@auth-helpers`](https://v6-api-doc-lit-js-sdk.vercel.app/modules/auth_helpers_src.html) package:
```

packages/types/src/lib/interfaces.ts
```diff
@@ -33,11 +33,37 @@ export interface AccsOperatorParams {
...
+   * -  [`createSiweMessage`](https://v6-api-doc-lit-js-sdk.vercel.app/functions/auth_helpers_src.createSiweMessage.html)
+   * -  [`createSiweMessageWithRecaps](https://v6-api-doc-lit-js-sdk.vercel.app/functions/auth_helpers_src.createSiweMessageWithRecaps.html)
+   * -  [`createSiweMessageWithCapacityDelegation`](https://v6-api-doc-lit-js-sdk.vercel.app/functions/auth_helpers_src.createSiweMessageWithCapacityDelegation.html)
+   */
   signedMessage: string;
+
+  /**
+   * The Ethereum public key address that was used to sign `signedMessage` and create the `sig`.
```
spacesailor24, line 60:
```suggestion
   * The Ethereum address that was used to sign `signedMessage` and create the `sig`.
```

packages/types/src/lib/interfaces.ts
```diff
@@ -1055,34 +1081,35 @@ export interface GetSignSessionKeySharesProp {
   body: SessionRequestBody;
 }
 export interface CommonGetSessionSigsProps {
+  // If you want to pass the `authNeededCallback`, having a PKP public key is necessary. This can be used if you do not have a wallet and wish to authenticate with your PKP.
```
spacesailor24: > If you want to pass the `authNeededCallback`, having a PKP public key is necessary.

I don't think is true?

packages/types/src/lib/interfaces.ts
```diff
@@ -1055,34 +1081,35 @@ export interface GetSignSessionKeySharesProp {
...
   pkpPublicKey?: string;
 
-  // When this session signature will expire.  The user will have to reauthenticate after this time using whatever auth method you set up.  This means you will have to call this signSessionKey function again to get a new session signature.  This is a RFC3339 timestamp.  The default is 24 hours from now.
+  // When this session signature will expire. After this time is up, you will need to reauthenticate using the same authentication method, generating a new session signature. The default time until expiration is 24 hours. The formatting is an [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) timestamp.
```
spacesailor24:
```suggestion
  // When this session signature will expire. After this time is up, you will need to reauthenticate, generating a new session signature. The default time until expiration is 24 hours. The formatting is an [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) timestamp.
```

packages/types/src/lib/interfaces.ts
```diff
@@ -1055,34 +1081,35 @@ export interface GetSignSessionKeySharesProp {
...
   expiration?: any;
 
-  //   The chain to use for the session signature.  This is the chain that will be used to sign the session key.  If you're using EVM then this probably doesn't matter at all.
+  // The chain to use for the session signature and sign the session key. If you're using EVM, this parameter isn't very important.
   chain?: Chain;
```
spacesailor24: I've never seen this value be anything other than `ethereum` tbh

packages/types/src/lib/interfaces.ts
```diff
@@ -1055,34 +1081,35 @@ export interface GetSignSessionKeySharesProp {
...
 
   /**
    * An array of resource abilities that you want to request for this session. These will be signed with the session key.
-   *
-   * @example If you want to request the ability to decrypt an access control condition, then you would pass
-   * [{ resource: new LitAccessControlConditionResource('someResource), ability: LitAbility.AccessControlConditionDecryption }]
+   * If you want to request the ability to decrypt an access control condition, then you would pass something similar to the example:
```
spacesailor24, line 1102: > ...to decrypt an access control condition

Confusing to me, @Ansonhkg @MaximusHaximus are we actually decrypting the ACC when we call decrypt?

packages/types/src/lib/interfaces.ts
```diff
@@ -1055,34 +1081,35 @@ export interface GetSignSessionKeySharesProp {
...
-   * @example If you passed nothing, and you're requesting to perform a decryption operation for an access
+   * If you passed nothing, and you're requesting to perform a decryption operation for an access
    * control condition, then the session capability object will be a wildcard for the access control condition,
    * which grants this session signature the ability to decrypt this access control condition.
    */
   sessionCapabilityObject?: ISessionCapabilityObject;
 
   /**
-   * If you want to ask Metamask to try and switch the user's chain, you may pass true here.  This will only work if the user is using Metamask.  If the user is not using Metamask, then this will be ignored.
+   * If you want to ask MetaMask to try and switch the user's chain, you may pass true here.  This will only work if the user is using MetaMask.  If the user is not using MetaMask, then this will be ignored.
```
spacesailor24, line 1119:
```suggestion
   * If you want to ask MetaMask to try and switch the user's chain, you may pass true here.  This will only work if the user is using MetaMask, otherwise this will be ignored.
```

packages/types/src/lib/interfaces.ts
```diff
@@ -1127,6 +1154,17 @@ export type AuthCallback = (params: AuthCallbackParams) => Promise<AuthSig>;
 /**
  * A map of node addresses to the session signature payload
  * for that node specifically.
+ * 
+ * Each individual session signature for each node includes the following properties:
+ * -  `sig`: The signature produced by the ed25519 key pair signing the `signedMessage` payload.
```
spacesailor24, line 1166:
```suggestion
 * -  `sig`: The signature produced by the ECDSA key pair signing the `signedMessage` payload.
```

packages/types/src/lib/interfaces.ts
```diff
@@ -1127,6 +1154,17 @@ export type AuthCallback = (params: AuthCallbackParams) => Promise<AuthSig>;
...
+ * 
+ * -  `derivedVia`: Should be `litSessionSignViaNacl`, specifies that the session signature object was created via the `NaCl` library.
+ * 
+ * -  `signedMessage`: The payload signed by the session key pair. This is the signed `AuthSig` with the contents of the AuthSig's `signedMessage` property being derived from the [`authNeededCallback`](https://v6-api-doc-lit-js-sdk.vercel.app/interfaces/types_src.GetSessionSigsProps.html#authNeededCallback) property.
+ * 
+ * -  `address`: The session key pair public key.
```
spacesailor24, line 1172: @Ansonhkg @MaximusHaximus Is this actually a public key, or the derived address?

packages/types/src/lib/interfaces.ts
awisniew207: It's not, got that idea from [here](https://v6-api-doc-lit-js-sdk.vercel.app/interfaces/types_src.GetSessionSigsProps.html#authNeededCallback). Might need to rewrite it as well

packages/types/src/lib/interfaces.ts
awisniew207: I'll put a note about that

packages/types/src/lib/interfaces.ts
awisniew207: Changed this

awisniew207:
Not sure why the most recent e2e test failed. I only changed a couple comments, looks like the `testExecuteJsSignAndCombineEcdsa` test failed

packages/types/src/lib/interfaces.ts
```diff
@@ -33,11 +33,36 @@ export interface AccsOperatorParams {
 
 /** ---------- Auth Sig ---------- */
 
+/**
+ * An `AuthSig` represents a cryptographic proof of ownership for an Ethereum address, created by signing a standardized [ERC-5573 SIWE](https://eips.ethereum.org/EIPS/eip-5573) (Sign-In with Ethereum) message. This signature serves as a verifiable credential, allowing the Lit network to associate specific permissions, access rights, and operational parameters with the signing Ethereum address. By incorporating various capabilities, resources, and parameters into the SIWE message before signing, the resulting `AuthSig` effectively defines and communicates these authorizations and specifications for the address within the Lit network.
```
joshLong145, line 37: So technically `5573` is for `SIWE ReCap` see here: https://ethereum-magicians.org/t/eip-5573-siwe-recap/10627 

So while you are correct I think we should specify `SIWE ReCap` over just `SIWE`

packages/types/src/lib/interfaces.ts
```diff
@@ -1055,34 +1081,41 @@ export interface GetSignSessionKeySharesProp {
   body: SessionRequestBody;
 }
 export interface CommonGetSessionSigsProps {
+  /**
+   *
```
joshLong145, line 1085: Can we add a description here?

packages/types/src/lib/interfaces.ts
joshLong145: So you can add an `ability` which will allow the `session` to decrypt content associated with the specific hash of the ACC conditions. So the conditions are not being decrypted. but the ability is looked up in the `resouceAbilityRequests` to figure out if the session has the ability to decrypt the content.

packages/types/src/lib/interfaces.ts
```diff
@@ -1055,34 +1081,41 @@ export interface GetSignSessionKeySharesProp {
...
+   * [{ resource: new LitAccessControlConditionResource('someResource`), ability: LitAbility.AccessControlConditionDecryption }]
    */
   resourceAbilityRequests: LitResourceAbilityRequest[];
 
   /**
    * The session capability object that you want to request for this session.
    * If you pass nothing, then this will default to a wildcard for each type of resource you're accessing.
    *
-   * @example If you passed nothing, and you're requesting to perform a decryption operation for an access
+   * If you passed nothing, and you're requesting to perform a decryption operation for an access
```
joshLong145, line 1111: I think the wording here is a bit confusing as we are explicitly outlining what will happen for all of our network options in a way which repeats language. Could we clean this up so we aren't repeating the same information?

packages/types/src/lib/interfaces.ts
```diff
@@ -1094,13 +1127,13 @@ export interface CommonGetSessionSigsProps {
   /**
    * @deprecated - use capabilityAuthSigs instead
    *  Used for delegation of Capacity Credit. This signature will be checked for proof of capacity credit.
-   * on both manzano and habanero networks capacity credit proof is required.
-   * see more here: https://developer.litprotocol.com/v3/sdk/capacity-credits
+   * on both Datil-test and Datil networks capacity credit proof is required.
```
joshLong145, line 1130: Im not sure if we want to use specific network names here as we will have to keep updating this comment header

packages/types/src/lib/interfaces.ts
joshLong145: In cases where the `session key` signs the `SIWE` it  will be the public key in its original format. But if it is an EOA wallet which signs then it will be the Ethereum address. It is a bit confusing as it can be two different data formats.

packages/types/src/lib/interfaces.ts
awisniew207: Changed first comment to:
> For example, an ability is added to grant a session permission to decrypt content associated with a particular Access Control Conditions (ACC) hash. When trying to decrypt, this ability is checked in the `resourceAbilityRequests` to verify if the session has the required decryption capability. 

The second comment is about `sessionCapabilityObject`, which going through the docs seemed very similar to `resourceAbilityRequests`. We don't have any example in the `developer-guides-code` or the Docs of using this, do we still encourage users to do so? Based on the description, a wildcard of each type of resource attempting to be accessed doesn't sound right. @joshLong145

packages/types/src/lib/interfaces.ts
awisniew207: Added:
> Session signature properties shared across all functions that generate session signatures.

packages/types/src/lib/interfaces.ts
awisniew207: Changed to:
> Capacity credits are required on the paid Lit networks (mainnets and certain testnets), and are not required on the unpaid Lit networks (certain testnets).

Tried to make it as general as possible, I'm not sure how to specify them otherwise without mentioning the networks by name.

awisniew207: Updated the below comment to say:
>When the session key signs the SIWE ReCap message, this will be the session key pair public key. If an EOA wallet signs the message, then this will be the EOA Ethereum address.

packages/types/src/lib/interfaces.ts
joshLong145: Yes you are correct in you assumption of being closely related to the `resourceAbilityRequests` we derive the `session capabilities` from the `resource request` you can see where we do this internally here: https://github.com/LIT-Protocol/js-sdk/blob/e66fb40a0d6ff269f25c6a06a7ad1cc00150326b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts#L281

This is not exposed since we derive from the given `ability requests` so hence the lack of examples.

joshLong145 (approved):

Ansonhkg (approved):
LGTM!

PR #573, opened 2024-08-02
Feature/lit 3745 Update WASM package

joshLong145:
Fixes rust build step for `wasm` package after a change in version to `hd-keys-curves-wasm` which was published as a public crate to `cargo registry`

Adds a wrapper method for `ecdsaCombnieAndVerify` which was added to the rust api surface.


packages/wasm/rust/Cargo.toml
```diff
@@ -1,8 +1,11 @@
 [package]
-name = "wasm"
+name = "lit-utilities-wasm"
```
Ansonhkg, line 2: What's the difference between `wasm` and `lit-utilities-wasm`? and why did we change from one to another?

packages/wasm/rust/Cargo.toml
```diff
@@ -15,7 +18,7 @@ blsful = { version = "2.5.7", default-features = false, features = ["rust"] }
 base64_light = "0.1"
 getrandom = { version = "0.2", features = ["js"] }
 hex = "0.4"
-hd-keys-curves-wasm = { version = "0.1", default-features = false, features = ["k256", "p256"], git = "https://github.com/LIT-Protocol/hd-keys-curves-wasm.git" }
+hd-keys-curves-wasm = { version = "1.0.0", default-features = false, features = ["k256", "p256"], git = "https://github.com/LIT-Protocol/hd-keys-curves-wasm.git" }
```
Ansonhkg, line 21: Do we always need to keep up to date to that repo?

Ansonhkg (approved):

packages/wasm/rust/Cargo.toml
joshLong145: I change the name for publishing to `cargo` since it would be a bit confusing if the package name just showed up as `wasm`.

packages/wasm/rust/Cargo.toml
joshLong145: Yes this package is published by @mikelodder7 I bumped the version and removed the `git url` for publishing as a public crate.

Ansonhkg (approved):
lgtm - will fix lint in staging branch

packages/wasm/rust/Cargo.toml
```diff
@@ -15,7 +18,7 @@ blsful = { version = "2.5.7", default-features = false, features = ["rust"] }
 base64_light = "0.1"
 getrandom = { version = "0.2", features = ["js"] }
 hex = "0.4"
-hd-keys-curves-wasm = { version = "0.1", default-features = false, features = ["k256", "p256"], git = "https://github.com/LIT-Protocol/hd-keys-curves-wasm.git" }
+hd-keys-curves-wasm = { version = "1.0.0", default-features = false, features = ["k256", "p256"] }
```
mikelodder7, line 21:
```suggestion
hd-keys-curves-wasm = { version = "1.0.1", default-features = false, features = ["k256", "p256"] }
```

packages/wasm/rust/Cargo.toml
```diff
@@ -39,7 +42,7 @@ sev = { version = "2.0.2", default-features = false, features = [
 rand = "0.8"
 serde_bytes = "0.11.14"
 tsify = { version = "0.4.5", default-features = false, features = ["js"] }
-jubjub = { version = "0.10.4", git = "https://github.com/LIT-Protocol/jubjub.git", branch = "main"}
+jubjub-plus = { version = "0.10.4" }
```
mikelodder7, line 45:
```suggestion
```

mikelodder7: This shouldn't be necessary

joshLong145: I thought I pushed this up. checking my local files.

packages/wasm/rust/Cargo.toml
joshLong145: Have updated

PR #576, opened 2024-08-02
Improve error management

FedericoAmura:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/access-control-conditions/src/lib/canonicalFormatter.ts
+++ b/packages/access-control-conditions/src/lib/canonicalFormatter.ts
@@ -1,5 +1,7 @@
-import { ILitError, LIT_ERROR } from '@lit-protocol/constants';
-
+import {
+  ILitError,
+  InvalidAccessControlConditions,
+} from '@lit-protocol/constants';
 import {
   ABIParams,
   AccessControlConditions,
@@ -14,8 +16,6 @@ import {
   JsonSigningResourceId,
   UnifiedAccessControlConditions,
 } from '@lit-protocol/types';
-
-import { throwError } from '@lit-protocol/misc';
 
 /** ---------- Local Functions ---------- */
 /**
@@ -40,7 +40,7 @@ const getOperatorParam = (cond: ConditionItem): AccsOperatorParams => {
  * @param { Array<ABIParams> } params
  * @returns { Array<ABIParams> }
  */
-const canonicalAbiParamss = (params: Array<ABIParams>): Array<ABIParams> => {
+const canonicalAbiParamss = (params: ABIParams[]): ABIParams[] => {
   return params.map((param) => ({
     name: param.name,
     type: param.type,
@@ -88,21 +88,27 @@ export const canonicalUnifiedAccessControlConditionFormatter = (
         return canonicalCosmosConditionFormatter(cond as AccsCOSMOSParams);
 
       default:
-        throwError({
-          message: `You passed an invalid access control condition that is missing or has a wrong "conditionType": ${JSON.stringify(
-            cond
-          )}`,
-          errorKind: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.kind,
-          errorCode: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.name,
-        });
+        throw new InvalidAccessControlConditions(
+          {
+            info: {
+              cond,
+            },
+          },
+          'You passed an invalid access control condition that is missing or has a wrong "conditionType": %s',
+          JSON.stringify(cond)
+        );
     }
   }
 
-  throwError({
-    message: `You passed an invalid access control condition: ${cond}`,
-    errorKind: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.kind,
-    errorCode: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.name,
-  });
+  throw new InvalidAccessControlConditions(
+    {
+      info: {
+        cond,
+      },
+    },
+    'You passed an invalid access control condition: %s',
+    JSON.stringify(cond)
+  );
 };
 
 /**
@@ -181,11 +187,14 @@ export const canonicalSolRpcConditionFormatter = (
         !('offset' in _assumedV2Cond.pdaInterface) ||
         !('fields' in _assumedV2Cond.pdaInterface)
       ) {
-        throwError({
-          message: `Solana RPC Conditions have changed and there are some new fields you must include in your condition.  Check the docs here: https://developer.litprotocol.com/AccessControlConditions/solRpcConditions`,
-          errorKind: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.kind,
-          errorCode: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.name,
-        });
+        throw new InvalidAccessControlConditions(
+          {
+            info: {
+              cond,
+            },
+          },
+          'Solana RPC Conditions have changed and there are some new fields you must include in your condition.  Check the docs here: https://developer.litprotocol.com/AccessControlConditions/solRpcConditions'
+        );
       }
 
       // -- else
@@ -226,11 +235,15 @@ export const canonicalSolRpcConditionFormatter = (
   }
 
   // -- else
-  throwError({
-    message: `You passed an invalid access control condition: ${cond}`,
-    errorKind: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.kind,
-    errorCode: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.name,
-  });
+  throw new InvalidAccessControlConditions(
+    {
+      info: {
+        cond,
+      },
+    },
+    'You passed an invalid access control condition: %s',
+    JSON.stringify(cond)
+  );
 };
 
 /**
@@ -284,11 +297,15 @@ export const canonicalAccessControlConditionFormatter = (
     return _return;
   }
 
-  throwError({
-    message: `You passed an invalid access control condition: ${cond}`,
-    errorKind: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.kind,
-    errorCode: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.name,
-  });
+  throw new InvalidAccessControlConditions(
+    {
+      info: {
+        cond,
+      },
+    },
+    'You passed an invalid access control condition: %s',
+    JSON.stringify(cond)
+  );
 };
 
 /**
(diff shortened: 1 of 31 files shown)


packages/crypto/src/lib/crypto.ts
```diff
@@ -184,7 +185,7 @@ export const combineEcdsaShares = async (
   const publicKey = Buffer.from(anyValidShare.publicKey, 'hex');
   const messageHash = Buffer.from(anyValidShare.dataSigned!, 'hex');
 
-  ecdsaVerify(variant!, messageHash, publicKey, [r, s, v]);
+  await ecdsaVerify(variant!, messageHash, publicKey, [r, s, v]);
```
FedericoAmura, line 188: I'm pretty sure this had to be awaited

packages/encryption/src/lib/encryption.ts
```diff
@@ -737,7 +803,7 @@ export const verifyJwt = ({
 
   const message = uint8arrayFromString(unsignedJwt);
 
-  verifySignature(publicKey, message, signature);
+  await verifySignature(publicKey, message, signature);
```
FedericoAmura, line 806: I'm pretty sure this had to be awaited

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -830,7 +843,7 @@ export class LitNodeClientNodeJs
     };
 
     // -- execute
-    keys.forEach(async (key) => {
+    for await (const key of keys) {
```
FedericoAmura, line 846: This surely had to be awaited. Without awaiting `signatures` was returned before it was populated with the result at the end of each iteration

packages/constants/src/lib/errors.ts
```diff
@@ -171,3 +190,103 @@ export const LIT_ERROR = {
...
+    constructor(options: Error | Options, message: string, ...params: any[]) {
+      if (options instanceof Error) {
+        options = {
+          cause: options,
+        };
+      }
+
+      // If the cause is not an Error, wrap it in one
+      if (!(options.cause instanceof Error)) {
+        options.cause = new Error(options.cause);
```
joshLong145, line 230: Will wrapping here have implications on the stack trace? Since this class is non named and we are creating a new instance of `Error` in scope of a function call we might want to take the stack trace from the `options` if it is of type `Error`

packages/constants/src/lib/errors.ts
```diff
@@ -171,3 +190,103 @@ export const LIT_ERROR = {
...
+
+const errorClasses: Record<string, LitErrorConstructor> = {};
+for (const key in LIT_ERROR) {
+  if (key in LIT_ERROR) {
+    const errorDef = LIT_ERROR[key];
+    errorClasses[errorDef.name] = createErrorClass(errorDef);
+  }
+}
+
+export const {
```
joshLong145, line 258: Do we still need to export explicitly snice we are creating an aggregate mapping of `errorClasses` or are we doing named exports to be explicit?

packages/crypto/src/lib/crypto.ts
joshLong145: I do not think we needed an await on this at the current point in time. but it shouldn't effect much since it will unblock in the same tick if the returned value is not a `Promise`

joshLong145 (commented):
Nice stuff, left some comments

packages/encryption/src/lib/encryption.ts
joshLong145: Same as above

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
joshLong145: Wondering if we just want to make this a normal `for` loop and ditch the iterators.

packages/access-control-conditions/src/lib/canonicalFormatter.ts
```diff
@@ -88,21 +85,27 @@ export const canonicalUnifiedAccessControlConditionFormatter = (
...
-          errorCode: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.name,
-        });
+        throw new InvalidAccessControlConditions(
+          {
+            info: {
+              cond,
+            },
+          },
+          'You passed an invalid access control condition that is missing or has a wrong "conditionType": %s',
+          JSON.stringify(cond)
```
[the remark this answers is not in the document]
FedericoAmura: It is passed as an object in the `info` property of the error. Can be obtained with `VError.info(err)`

packages/constants/src/lib/errors.ts
FedericoAmura: I prefer being explicit as when having one file per error; but with so many errors, it's hard to keep track of them all and modify them.
This felt like a good compromise to retain errors individually but centralize their properties and differences based on `ErrorConfig` and `LIT_ERROR`

packages/constants/src/lib/errors.ts
FedericoAmura: This is added because `cause` has to be an `Error` or the `VError` constructor will throw itself, therefore we wrap to comply with it.
When that happens thrower threw something else and we do not have a stacktrace originating from the thrower and hide.

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
FedericoAmura: Do we know if somebody was using this? The only way I see this working is that wasm was already loaded and it behaved completely synchronously regardless of the `async` and `await` calls.
If not, then the response would be a `{}`, generating race condition on that and prone to a future update breaking it.

packages/constants/src/lib/errors.ts
joshLong145: I see now, since we are in scope of a constructor  the stack trace is going is going to be from the scope of the constructor call in a way preserve the stack context.

packages/lit-node-client-nodejs/src/lib/helpers/get-signatures.ts
```diff
@@ -145,22 +150,21 @@ export const getSignatures = async <T>(params: {
...
-    });
+    throw new NoValidShares(
+      {},
+      'total number of valid signatures does not match requested'
+    );
   }
 
   // -- combine
-  for (var i = 0; i < allKeys.length; i++) {
+  for (const key of allKeys) {
```
joshLong145, line 160: Is there a benefit of using an iterator over just accessing the elements of the array?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
joshLong145: The warm calls are currently synchronous operations. In `v7` this will no longer be true, but not because the wasm execution has become asynchronous. Rather we have made the api async as to future proof it in some ways.

I don't think we need the async wrappers, personally would rather this just turn into a `for` loop as the iteration that using `for of` and `forEach` seems like will cause higher GC times.

packages/constants/src/lib/errors.ts
joshLong145: This makes sense to me. Nice inclusion.

packages/access-control-conditions/src/lib/canonicalFormatter.ts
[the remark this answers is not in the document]
joshLong145: Ah right, do we need to pass the `condition` as a string then if we already passing as an object since we can call `stringify` internally?

packages/lit-node-client-nodejs/src/lib/helpers/get-signatures.ts
FedericoAmura: We have an [eslint rule](https://typescript-eslint.io/rules/prefer-for-of/) to prefer that. I don't think there is a considerable difference between the two

This change is just to lower the IDE warnings so someday, that can help us

packages/access-control-conditions/src/lib/canonicalFormatter.ts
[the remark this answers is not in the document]
FedericoAmura: I wanted to keep the changes size as little as possible so this does not become impossible to test/merge, that is why I didn't change the error messages

But thinking more about it, no one should rely on messages strings and stringifying that can only bring problems as we already have it in the `info`. It is part of v7 also so users should expect breaking changes

Removing it as we will hardly revisit error messages anytime soon

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
FedericoAmura: In this loop, [`combineEcdsaShares` is async](https://github.com/LIT-Protocol/js-sdk/blob/3eaa12a99e3f085b62b5d1c3aa1864cd40af4318/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts#L929) and internally it calls `ecdsaCombine` and `ecdsaVerify` both of which are defined `async` because of a call to `loadModules`

At least when loading the wasm, they are asynchronous and will release the event loop. We are at v7 here
Or I am missing what you mean?

packages/access-control-conditions/src/lib/canonicalFormatter.ts
```diff
@@ -88,21 +85,25 @@ export const canonicalUnifiedAccessControlConditionFormatter = (
...
-          errorKind: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.kind,
-          errorCode: LIT_ERROR.INVALID_ACCESS_CONTROL_CONDITIONS.name,
-        });
+        throw new InvalidAccessControlConditions(
+          {
+            info: {
+              cond,
+            },
+          },
+          'You passed an invalid access control condition that is missing or has a wrong "conditionType"'
```
joshLong145, line 94: nit: Can these be made into constants? Unless it isn't intended to be reused.

joshLong145 (commented):
Looks good! Last remaining question for me is do we keep the static string in the error args or move them to constant declarations for consistency.

Ansonhkg (approved):
LGTM - can you also provide a short description at the top of what a developer needs to do to throw a new type error, such as creating a new error type and specifying the file where it should be done?

joshLong145 (approved):

Ansonhkg:
> > LGTM - can you also provide a short description at the top of what a developer needs to do to throw a new type error, such as creating a new error type and specifying the file where it should be done?
> 
> Added it in the main readme so consumers and devs along have it at hand

Awesome!!

PR #580, opened 2024-08-07
Feature/lit 3744 js sdk from slack use getg active unkicked validator struct and counts 2

Ansonhkg:
# Description

tldr: this change will increase about 40% connection speed

Merge multiple connection to a RPC call

- feat(contracts-sdk): introduce `getConnectionInfo` public static function, that calls the `getActiveUnkickedValidatorStructsAndCounts` smart contract function to get `epochInfo`, `minNodeCount` and `activeUnkickedValidatorStructs`
- feat(lit-core): refactor `_getValidatorData` to use `getConnectionInfo` from the contracts-sdk, which will be used in the initial `_connect` function, to set the stakingContract (to be re-used in other code), minNodeCount, and bootstrapUrls.

# Test

- Running local nodes from this PR https://github.com/LIT-Protocol/lit-assets/pull/1372 and ran 

```
NETWORK=localchain yarn test:local --filter=testUseEoaSessionSigsToExecuteJsSigning --exclude=Parallel
```


packages/core/src/lib/lit-core.ts
```diff
@@ -220,41 +220,85 @@ export class LitCore {
...
+    // Validate epoch
+    if (epoch === undefined || epoch === null) {
+      throwError({
+        message: 'epoch is required',
+        errorKind: LIT_ERROR.INIT_ERROR.kind,
+        errorCode: LIT_ERROR.INIT_ERROR.name,
+      });
+    }
+
+    const epochInInt = parseInt(epoch.toString());
```
joshLong145, line 284: we might want to specify a radix in this case since we might eventually change the numeric base. So perhaps we explicitly specify `base 10`

packages/core/src/lib/lit-core.ts
```diff
@@ -220,41 +220,85 @@ export class LitCore {
...
-        this.config.rpcUrl,
-        this.config.nodeProtocol
-      ),
-    ]);
+  /**
+   * Retrieves the validator data including staking contract, epoch, minNodeCount, and bootstrapUrls.
+   * @returns An object containing the validator data.
+   * @throws Error if minNodeCount is not provided, is less than or equal to 0, or if bootstrapUrls are not available.
+   */
+  private async _getValidatorData(): Promise<{
```
joshLong145, line 228: I think we can also remove the polling for the `epoch` as we have concluded that we only need to resync when the epoch transitions which we have an event listener for.

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -862,83 +862,66 @@ export class LitContracts {
     return addresses;
   }
 
-  public static getMinNodeCount = async (
```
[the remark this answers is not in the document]
Ansonhkg: Oh yea - I had in my mind that it will be merged to V7, but actually we want it before that. Will add this back!

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -862,83 +862,66 @@ export class LitContracts {
...
-      // Process each validator
-      for (const validator of activeValidators) {
-        validators.push(validator);
-      }
-    } else {
-      LitContracts.logger.error(
-        '❌ Active validator set does not meet the threshold'
-      );
+    if (activeUnkickedValidatorStructs.length <= minNodeCount) {
+      throw new Error('❌ Active validator set does not meet the threshold');
```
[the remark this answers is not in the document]
Ansonhkg: Yea, I don't think the SDK should function at all if it doesn't meet the threshold, but we could display the data in the error message.

Also makes sense to put it into a VError!

Ansonhkg:
Ready to review

NOTE: Need this to be merged for the CI to run:
https://github.com/LIT-Protocol/lit-assets/pull/1394

packages/core/src/lib/lit-core.ts
```diff
@@ -223,41 +224,66 @@ export class LitCore {
...
+      await LitContracts.getConnectionInfo({
+        litNetwork: this.config.litNetwork,
+        networkContext: this.config.contractContext,
+        rpcUrl: this.config.rpcUrl,
+        nodeProtocol: this.config.nodeProtocol,
+      });
 
-    if (minNodeCount <= 0) {
+    // Validate minNodeCount
+    if (minNodeCount === undefined || minNodeCount === null) {
```
FedericoAmura, line 247:
```suggestion
if (!minNodeCount) {
```

So it also checks if the minimum node count is set to 0. Can that happen?

packages/core/src/lib/lit-core.ts
```diff
@@ -223,41 +224,66 @@ export class LitCore {
...
       throwError({
-        message: `Failed to get bootstrapUrls for network ${this.config.litNetwork}`,
+        message: 'stakingContract is required',
         errorKind: LIT_ERROR.INIT_ERROR.kind,
         errorCode: LIT_ERROR.INIT_ERROR.name,
       });
     }
 
+    // Validate epoch
+    if (epochInfo.number === undefined || epochInfo.number === null) {
```
FedericoAmura, line 270: Same as `minNodeCount` with the conditional and error message

packages/core/src/lib/lit-core.ts
```diff
@@ -223,41 +224,66 @@ export class LitCore {
...
+        litNetwork: this.config.litNetwork,
+        networkContext: this.config.contractContext,
+        rpcUrl: this.config.rpcUrl,
+        nodeProtocol: this.config.nodeProtocol,
+      });
 
-    if (minNodeCount <= 0) {
+    // Validate minNodeCount
+    if (minNodeCount === undefined || minNodeCount === null) {
+      throw new Error('minNodeCount is required');
```
FedericoAmura, line 248:
```suggestion
  throw new Error('Could not obtain minNodeCount');
```

"required" gives me the impression that the consumer has to provide it

packages/core/src/lib/lit-core.ts
```diff
@@ -775,20 +798,20 @@ export class LitCore {
...
+
+  //   this._networkSyncInterval = setInterval(async () => {
+  //     if (
+  //       !this.lastBlockHashRetrieved ||
+  //       Date.now() - this.lastBlockHashRetrieved >= BLOCKHASH_SYNC_INTERVAL
+  //     ) {
+  //       await this._syncBlockhash();
+  //     }
+  //   }, BLOCKHASH_SYNC_INTERVAL);
+  // }
```
FedericoAmura, line 814: Awesome we are finally removing this

FedericoAmura (commented):
Just some minor comments

packages/core/src/lib/lit-core.ts
Ansonhkg: Yea that might happen if the value is not set on the contract

packages/core/src/lib/lit-core.ts
Ansonhkg: Good point!

joshLong145 (approved):

PR #584, opened 2024-08-08
fix: race condition on getting latest blockhash

Ansonhkg:
# Description

Fetch again if latest blockhash is unavailable.

### DIFF
--- a/local-tests/setup/session-sigs/get-eoa-session-sigs.ts
+++ b/local-tests/setup/session-sigs/get-eoa-session-sigs.ts
@@ -147,7 +147,7 @@ export const getEoaSessionSigsWithCapacityDelegations = async (
         expiration: expiration,
         resources: resourceAbilityRequests,
         walletAddress: fromWallet.address,
-        nonce: devEnv.litNodeClient.latestBlockhash,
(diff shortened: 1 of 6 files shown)


FedericoAmura:
Found a missing replacement [here](https://github.com/LIT-Protocol/js-sdk/blob/feature/lit-3763-js-sdk-patch-eth-blockhash-is-undefined/packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts#L166) and 

Would suggest making this a `private` or `protected` property to avoid future incorrect usages of that prop. Or `deprecated`, whatever helps us, and our customers, avoid this later

joshLong145 (approved):
Nice fix :)

joshLong145:
> Found a missing replacement [here](https://github.com/LIT-Protocol/js-sdk/blob/feature/lit-3763-js-sdk-patch-eth-blockhash-is-undefined/packages/lit-auth-client/src/lib/providers/EthWalletProvider.ts#L166)
> 
> Would suggest making this a `private` or `protected` property to avoid future incorrect usages of that prop. Or `deprecated`, whatever helps us, and our customers, avoid this later

Good call, I think `protected` would serve us well here.

FedericoAmura (approved):

PR #590, opened 2024-08-13
Feature/lit 3767 fix could only connect to 8 of 6 required nodes from 9 all nodes

FedericoAmura:
# Description

This PR:
- updates the "connected to X of Y nodes, with a minimum of Z" as that was misleading when X>=Z
- adds retry logic to each node handshake request (with a flag to [de]activate it, and retries until handshake timeout is reached). Defaults to `true`
- adds a per node HANDSHAKE timeout, configurable. (apart from full handshake timeout). This one overrides the request timeout if present. Defaults to `7500`ms.
- adds per node REQUEST timeout (applies also to handshake if present and handshake one is not), configurable. Defaults to `35000`ms (because LitActions have a 30s timeout, plus some network lag)

With default configs, the handshake process will attempt to connect to each node, the ones that don't succeed before 7500ms will be cancelled and retried in loop until they all succeed or handshake timeout is reached
When retry is set to false, the first failing node will fail the whole process (can be changed for naga allowing X=>Z handshakes)

Screenshot from 2024-08-14 19-15-36

## Type of change

# How Has This Been Tested?

- [x] Manually tested several configs with a demo app

# Checklist:

### DIFF
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -639,13 +639,11 @@ export class LitCore {
     await Promise.race([
       new Promise((_resolve, reject) => {
         timeoutHandle = setTimeout(() => {
-          const msg = `Error: Could not connect to enough nodes after timeout of ${
+          const msg = `Error: Could not connect to all required nodes after timeout of ${
             this.config.connectTimeout
-          }ms.  Could only connect to ${Object.keys(serverKeys).length} of ${
-            this.config.minNodeCount
-          } required nodes, from ${
+          }ms. Could only connect to ${Object.keys(serverKeys).length} of ${
             this.config.bootstrapUrls.length
-          } possible nodes.  Please check your network connection and try again.  Note that you can control this timeout with the connectTimeout config option which takes milliseconds.`;
+          } required nodes. Please check your network connection and try again. Note that you can control this timeout with the connectTimeout config option which takes milliseconds.`;
 
           try {
             // TODO: Kludge, replace with standard error construction
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -986,7 +986,11 @@ export class LitCore {
       const abortController = new AbortController();
       req.signal = abortController.signal;
       abortTimeout = setTimeout(() => {
-        abortController.abort(new Error(`Request to node timed out based on config. Timeout: ${requestTimeout}`));
+        abortController.abort(
+          new Error(
+            `Request to node timed out based on config. Timeout: ${requestTimeout}`
+          )
+        );
       }, requestTimeout);
     }
 
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -977,17 +977,17 @@ export class LitCore {
     };
 
     let abortTimeout: ReturnType<typeof setTimeout>;
-    if (timeout || this.config.nodeRequestTimeout) {
+    const requestTimeout = timeout || this.config.nodeRequestTimeout;
+    if (requestTimeout) {
       // TODO: Use AbortSignal.timeout once we update to TS >=4.9.5. It considers request active time instead of elapsed time
       // Example:
-      // const abortSignal = AbortSignal.timeout(timeout || this.config.nodeRequestTimeout);
+      // const abortSignal = AbortSignal.timeout(requestTimeout);
       // req.signal = abortSignal;
       const abortController = new AbortController();
       req.signal = abortController.signal;
-      abortTimeout = setTimeout(
-        () => abortController.abort('Request to node timed out'),
-        timeout || this.config.nodeRequestTimeout
-      );
+      abortTimeout = setTimeout(() => {
+        abortController.abort(new Error(`Request to node timed out based on config. Timeout: ${requestTimeout}`));
+      }, requestTimeout);
     }
 
     const sendRequestPromise = sendRequest(url, req, requestId);
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -979,7 +979,9 @@ export class LitCore {
     let abortTimeout: ReturnType<typeof setTimeout>;
     if (timeout || this.config.nodeRequestTimeout) {
       // TODO: Use AbortSignal.timeout once we update to TS >=4.9.5. It considers request active time instead of elapsed time
-      // AbortSignal.timeout(timout || this.config.nodeRequestTimeout)
+      // Example:
+      // const abortSignal = AbortSignal.timeout(timeout || this.config.nodeRequestTimeout);
+      // req.signal = abortSignal;
       const abortController = new AbortController();
       req.signal = abortController.signal;
       abortTimeout = setTimeout(
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -99,7 +99,14 @@ export type LitNodeClientConfigWithDefaults = Required<
   >
 > &
   Partial<
-    Pick<LitNodeClientConfig, 'storageProvider' | 'contractContext' | 'rpcUrl' | 'retryNodeHandshake'>
+    Pick<
+      LitNodeClientConfig,
+      | 'storageProvider'
+      | 'contractContext'
+      | 'rpcUrl'
+      | 'nodeRequestTimeout'
+      | 'retryNodeHandshake'
+    >
   > & {
     bootstrapUrls: string[];
   } & {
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/lit-core.ts
```diff
@@ -837,15 +865,13 @@ export class LitCore {
...
-      challenge: params.challenge,
+      challenge,
     };
 
     return await this.sendCommandToNode({
       url: urlWithPath,
       data,
       requestId,
-    }).catch((err: NodeErrorV3) => {
-      return err;
```
FedericoAmura, line 848: Hiding this error caused `handshakeResult` to be `err`, then placing `undefined` for all values in `keys` [here](https://github.com/LIT-Protocol/js-sdk/pull/590/files#diff-1cc39c82b504bd1b4173261252ff8fab20ec90a38134c9aa0f7d2e57094e8a89L539-L551)

Issue was that when returning an error, like the timeout one, finished everything without throwing. And this could put us in the `latestBlockhash is undefined` state

packages/core/src/lib/lit-core.ts
```diff
@@ -945,7 +971,25 @@ export class LitCore {
...
+    if (this.config.nodeRequestTimeout) {
+      // TODO: Use AbortSignal.timeout once we update to TS >=4.9.5. It considers request active time instead of elapsed time
+      // AbortSignal.timeout(this.config.nodeRequestTimeout)
+      const abortController = new AbortController();
+      req.signal = abortController.signal;
+      abortTimeout = setTimeout(
+        () => abortController.abort('Request to node timed out'),
+        this.config.nodeRequestTimeout
+      );
+    }
```
FedericoAmura, line 984: This timeout also applies for lit action or any other request made to the nodes. I can change it to just the handshake ones, or make it configurable afterwards to give consumers more flexibility

packages/core/src/lib/lit-core.ts
```diff
@@ -123,7 +130,10 @@ export class LitCore {
   config: LitNodeClientConfigWithDefaults = {
     alertWhenUnauthorized: false,
     debug: true,
-    connectTimeout: 20000,
+    connectTimeout: 20_000,
+    nodeConnectionTimeout: 7_500,
+    nodeRequestTimeout: 35_000, // LitActions have a 30s timeout + some buffer for network latency
+    retryNodeHandshake: true,
```
FedericoAmura, line 136: `connectTimeout` is global handshake timeout
`nodeConnectionTimeout` is per node handshake timeout, it overrides `nodeRequestTimeout` at handshake
`nodeRequestTimeout` is per node timeout to apply to ALL requests made to a node
`retryNodeHandshake` is a flag customer can use to allow retrying handshake or failing fast when the first node fails. This will be more handy on naga when we can just discard failing nodes if we reach threshold instead of waiting more to get them

packages/core/src/lib/lit-core.ts
```diff
@@ -945,7 +976,25 @@ export class LitCore {
       body: JSON.stringify(data),
     };
 
-    return sendRequest(url, req, requestId);
+    let abortTimeout: ReturnType<typeof setTimeout>;
+    if (timeout || this.config.nodeRequestTimeout) {
+      // TODO: Use AbortSignal.timeout once we update to TS >=4.9.5. It considers request active time instead of elapsed time
+      // AbortSignal.timeout(timout || this.config.nodeRequestTimeout)
```
Ansonhkg, line 982:
```suggestion
      // Example:
      // const abortSignal = AbortSignal.timeout(timeout || this.config.nodeRequestTimeout);
      // req.signal = abortSignal;
```

packages/types/src/lib/interfaces.ts
```diff
@@ -193,7 +193,14 @@ export interface LitNodeClientConfig {
...
   minNodeCount?: number;
   debug?: boolean;
+  // Timeout for connecting/handshaking with all nodes
   connectTimeout?: number;
+  // Timeout for connection request made to each node. If not provided, the default value is 7500ms
+  nodeConnectionTimeout?: number;
+  // Timeout for each request made to a node
+  nodeRequestTimeout?: number;
+  // Whether to retry handshaking a particular node if it failed, will retry until connectTimeout is reached
+  retryNodeHandshake?: boolean;
```
Ansonhkg, line 203: use jsDoc

```suggestion
  /**
   * Timeout for connecting/handshaking with all nodes
   */
  connectTimeout?: number;
  /**
   * Timeout for connection request made to each node. If not provided, the default value is 7500ms
   */
  nodeConnectionTimeout?: number;
  /**
   * Timeout for each request made to a node
   */
  nodeRequestTimeout?: number;
  /**
   * Whether to retry handshaking a particular node if it failed, will retry until connectTimeout is reached
   */
  retryNodeHandshake?: boolean;
```

image

packages/types/src/lib/interfaces.ts
```diff
@@ -641,6 +648,8 @@ export interface SendNodeCommand {
   url: string;
   data: any;
   requestId: string;
+  // Timeout for this request. Overrides the config's nodeRequestTimeout
```
Ansonhkg, line 651: use jsDoc
```suggestion
  /**
   * Timeout for this request. Overrides the config's nodeRequestTimeout
   */
```

packages/core/src/lib/lit-core.ts
```diff
@@ -637,16 +647,42 @@ export class LitCore {
...
+        }
+      }
+
+      // If we get here, we've failed to handshake with the node. We should retry.
+      logErrorWithRequestId(
+        requestId,
+        `Error while attempting handshake with node ${url}. Retrying...`
+      );
+      return await handshakeWithNodeWithRetry(url);
+    };
```
Ansonhkg, line 675: I think we need to add a max retry here for the recursion to avoid stack overflow if the number of retries is large, and it looks like the canRetry flag remains unchanged if the handshake fails?

```suggestion
const handshakeWithNodeWithRetry = async (url: string): Promise<void> => {
  let attempts = 0;
  const maxRetries = 3; // <-- set a max retry limit

  while (attempts < maxRetries) {
    try {
     const res =  await handshakeWithNode(url);
      
      // Reset  after a successful handshake to prevent further retries.
      canRetry = false;
      
      return res;
      
    } catch (e) {
      attempts++;
      if (!canRetry || attempts >= maxRetries) {
        throw e;
      }
    }

    logErrorWithRequestId(
      requestId,
      `Error while attempting handshake with node ${url}. Retrying attempt ${attempts}...`
    );
  }
};

```

FedericoAmura: The `maxRetries` is implicitly defined by  `connectTimeout/nodeConnectionTimeout`, which is 3 in this case
Or you mean a global `maxRetries` for all nodes combined?

Stack won't grow on this case because of [tail call optimization](https://exploringjs.com/es6/ch_tail-calls.html)

`canRetry` is not set to `false` on fail because you can call `connect` again, maybe waiting some time or after you see the connection is up again (maybe you have a connection monitor)

Ansonhkg (approved):

packages/core/src/lib/lit-core.ts
joshLong145: Can we check status codes to prevent retries on '500' or '429'

FedericoAmura: Good idea 🤔 

However I'm having issues doing that without custom errors as I have to bubble up some info from the network request layer to the handshake logic

I think it is better to first ship v7 with that and then resume this. If not even doing this we will have to update error handling afterwards

joshLong145: I see, could we instead check the `status` and attempt to log the `message` if present in the response body? If this doesn't sound like a good solution we can just hold it for now.

FedericoAmura: That is the issue. I ended up refactoring the `sendRequest` function in `misc` to throw something more informative but at the end, I still had to do something like checking if there was a code or 'network error' in the message. Not good

After merging v7 I will be able to modify that to throw `NetworkError` with the http code or whatever we need to properly resolve the retry

joshLong145: Thanks for the added details. Makes sense to me. I think we can go ahead without the status checks and follow up with them if that sounds good.

FedericoAmura: I added the validation for these cases in #624 where I could use the new errors

FedericoAmura:
Closing this and replaced by #624 as that is using `staging/v7` as base. It was easier that than rebasing
I also improved error handling and retry logic there. Using this PR would leave some followups and nasty code that we would have to update shortly so there was no reason to stay on this one

PR #591, opened 2024-08-13
fix: missing SEV check for Datil

Ansonhkg:
Adding Datil to `NETWORKS_REQUIRING_SEV`

### DIFF
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -116,6 +116,7 @@ const NETWORKS_REQUIRING_SEV: string[] = [
   LitNetwork.Habanero,
   LitNetwork.Manzano,
   LitNetwork.DatilTest,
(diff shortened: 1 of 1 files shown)


MaximusHaximus (approved):
:+1: Shucks.

packages/core/src/lib/lit-core.ts
```diff
@@ -116,6 +116,7 @@ const NETWORKS_REQUIRING_SEV: string[] = [
   LitNetwork.Habanero,
   LitNetwork.Manzano,
   LitNetwork.DatilTest,
+  LitNetwork.Datil,
```
joshLong145, line 119: Is this reworking `datil-test`

Ansonhkg: It's working on `datil-test` if that's what you meant?

glitch003 (approved):

PR #597, opened 2024-08-14
divide by 100 after multiplying by 110 to get a 10% increase

glitch003:
The https://github.com/LIT-Protocol/js-sdk/pull/567 PR seems to have a bug in it that multiplies the gas limit by 110 instead of increasing it by 10%.  This is fixed by adding a division by 100 after the multiplication, which results in a 10% increase

Edit: this PR also adds a fix to the epoch delay, to make it work with the tests.  It works fine on live networks, but in the tests, when we try to use epoch 2, the delay hasn't passed yet, so we try to use epoch 1.  but epoch 1 is the first epoch and has no peers in it.  so the signing process fails in the tests.  the nodes go "i don't know about that peer".  so i put in a fix to only apply the delay if epochNumber >= 3.

### DIFF
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -893,7 +893,7 @@ export class LitCore {
       Math.floor(Date.now() / 1000) <
         this._epochCache.startTime +
           Math.floor(EPOCH_PROPAGATION_DELAY / 1000) &&
-      this._epochCache.currentNumber > 2
+      this._epochCache.currentNumber >= 3
     ) {
       return this._epochCache.currentNumber - 1;
     }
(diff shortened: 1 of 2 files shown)


joshLong145 (approved):
Good catch!

packages/core/src/lib/lit-core.ts
```diff
@@ -891,7 +891,9 @@ export class LitCore {
       this._epochCache.currentNumber &&
       this._epochCache.startTime &&
       Math.floor(Date.now() / 1000) <
-        this._epochCache.startTime + Math.floor(EPOCH_PROPAGATION_DELAY / 1000)
+        this._epochCache.startTime +
+          Math.floor(EPOCH_PROPAGATION_DELAY / 1000) &&
+      this._epochCache.currentNumber >= 3
```
joshLong145, line 896: This makes sense, approving as this is assuringly to prevent rolling the epoch back to either of the initial dkgs.

glitch003: exactly :-D

FedericoAmura:
Great catch. Thanks!!!

PR #603, opened 2024-08-20
Fix/manually fetch blockhash when its not available

Ansonhkg:
# Description

Add public RPCs in the SDK as a fallback only after the Indexers fails

Note: Also we should add fallback in the Indexers service


packages/core/src/lib/lit-core.ts
```diff
@@ -751,15 +766,65 @@ export class LitCore {
...
+          const latestBlock = await provider.getBlock('latest');
+          this.latestBlockhash = latestBlock.hash;
+          this.lastBlockHashRetrieved = Date.now();
+          log(
+            'Successfully retrieved blockhash manually: ',
+            this.latestBlockhash
+          );
+        } catch (ethersError) {
+          logError('Failed to manually retrieve blockhash using ethers');
+        }
```
FedericoAmura, line 827: Why is this duplicated here? If the provider fails it will likely fail again and originally it was fine not doing anything anyway. Maybe the `then` one is the extra one and it should throw when `!this.latestBlockhash`?


BTW, can't we remove the `_networkSyncInterval`? We are now always calling `getLatestBlockhash` instead of accessing `latestBlockhash` from outside. We could [make it `protected` or `private`](https://github.com/LIT-Protocol/js-sdk/pull/584#issuecomment-2278550582) in v7

packages/core/src/lib/lit-core.ts
```diff
@@ -734,6 +735,20 @@ export class LitCore {
     };
   }
 
+  private _getProviderWithFallback =
+    async (): Promise<ethers.providers.JsonRpcProvider | null> => {
+      for (const url of FALLBACK_RPC_URLS) {
+        try {
+          const provider = new ethers.providers.JsonRpcProvider(url);
```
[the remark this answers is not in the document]
Ansonhkg: agreed - but i think that's actually 120 seconds (2 mins). I'm cutting it to 60s.

packages/core/src/lib/lit-core.ts
```diff
@@ -776,15 +805,45 @@ export class LitCore {
...
+        log(
+          'Attempting to fetch blockhash manually using ethers with fallback RPC URLs...'
+        );
+        const provider = await this._getProviderWithFallback();
+
+        if (!provider) {
+          logError(
+            'All fallback RPC URLs failed. Unable to retrieve blockhash.'
+          );
+          return;
```
FedericoAmura, line 833: I think I found an edge case that can break. Follow me on this one:
- `fetch(this._blockHashUrl)` at line 802 fails -> we get into the `catch` clause
- the `catch` clause cannot get a provider or the one it gets cannot set the `latestBlockhash` (exiting at this line or failing at 837 `await provider.getBlock('latest')`)

If that happens (unstable providers/network might be a reason) then `_syncBlockhash` does not update `latestBlockhash`. The problem arises when considering that neither it nor its caller `getLatestBlockhash` discarded the old value of `latestBlockhash`.
So, if we already had a `latestBlockhash`, that might be outdated, then on `getLatestBlockhash` (line 450+) we might be not returning the real latest blockhash but an outdated one

It might not be such an issue as I think nodes still accept a wide range of blockhashes. Anyway, I think this would be easily fixed by setting `this.latestBlockhash = null` at the beginning of `_syncBlockhash`

Ansonhkg: I think Instead of setting this.latestBlockhash to null immediately, we can check if the existing blockhash is outdated. If the blockhash is outdated, then proceed with the logic to fetch a new one. This prevents clearing a valid blockhash if it's still current and avoids using an outdated blockhash when it fails to sync. (we can re-use the code commented-out at 867)

so adding this in the beginning:
```
  private async _syncBlockhash() {
    const currentTime = Date.now();
    const blockHashValidityDuration = BLOCKHASH_SYNC_INTERVAL;

    if (
      this.latestBlockhash &&
      this.lastBlockHashRetrieved &&
      currentTime - this.lastBlockHashRetrieved < blockHashValidityDuration
    ) {
      log('Blockhash is still valid. No need to sync.');
      return;
    }
```

FedericoAmura: I like it 🚀

FedericoAmura (approved):

joshLong145 (approved):

PR #608, opened 2024-08-22
feat: remove encryption package from lit-node-client

FedericoAmura:
# Description

Screenshot from 2024-08-22 18-47-17

This PR removes the inclusion of `@lit-protocol/encryption` inside `@lit-protocol/lit-node-client-nodejs`

Changes:
- Adds new functions (with its test) to encrypt and decrypt `Uint8Array` data types, which is the base type for all. This way when using encryption/decryption everything can be done from `@lit-protocol/encryption` in a standard way
- moves `verifyJWT` function from `@lit-protocol/encryption` to `@lit-protocol/crypto`
- moves the following exported classes, interfaces and functions from `@lit-protocol/encryption` to `@lit-protocol/misc`
  - `isTokenOperator`
  - `isValidBooleanExpression`
  - `safeParams`
  - `AccessControlConditionsValidatorProps`
  - `AuthMaterialValidatorProps`
  - `ParamsValidatorsType`
  - `safeParams`
- removes re-exports of `@lit-protocol/encryption` functions from `@lit-protocol/lit-node-client-nodejs`

Copied references in the old packages were not maintained because that would maintain the inclusion path and prevent tree shaking, and likely include encryption package (with JsZip 🤐 ) even when not using it

## Type of change

# Breaking changes
The functions moved and indicated above are no longer exported from `@lit-protocol/lit-node-client-nodejs` nor its children `@lit-protocol/lit-node-client` and have to be imported from the corresponding package

# How Has This Been Tested?

- [x] Updated tests that do encryption & decryption
- [x] Added test for the new `Uint8Array` encryption & decryption functions
- [x] CI still runs fine 😄 

# Checklist:


Ansonhkg (approved):
I feel a few kgs lighter just looking at this!

packages/crypto/src/lib/crypto.ts
```diff
@@ -452,3 +455,75 @@ export const checkSevSnpAttestation = async (
...
   return sevSnpVerify(report, data, signatures, challenge, vcekCert);
 };
+
+declare global {
+  // `var` is required for global hackery
+  // FIXME: `any` types for wasm are no bueno
+  // eslint-disable-next-line no-var, @typescript-eslint/no-explicit-any
+  var wasmExports: any;
+  // eslint-disable-next-line no-var, @typescript-eslint/no-explicit-any
+  var wasmECDSA: any;
```
joshLong145, line 465: I don't think we sell need these defined on the global as in `v7` we scope the `wasm module`

joshLong145 (commented):
Approving but left a comment on the `global` declare

packages/crypto/src/lib/crypto.ts
FedericoAmura: It is used [here](https://github.com/LIT-Protocol/js-sdk/blob/feature/lit-3579-v8-encryption-package-out-of-litnodeclient/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts#L990) 🫠

joshLong145: Hmm good catch! Making a task for me to go in and refactor this.

joshLong145 (approved):

PR #609, opened 2024-08-22
feat: add `GLOBAL_OVERWRITE_IPFS_CODE_BY_NETWORK`

Ansonhkg:
# Description

By default, we will always fetch the IPFS content for the datil mainnet network. Other network can specify to overwrite by passing it the `ipfsOptions` param to the `executeJs`, `getPkpSessionSigs` and `getLitActionSessionSigs` functions.

This is where we change the [default](https://github.com/LIT-Protocol/js-sdk/pull/609/files#diff-f0af9d412918612361cd43d65265f31c219f32b98baaafe4f3eb98c9aea221f9R54).

Also added `FALLBACK_IPFS_GATEWAYS` for fallback gateways, here's the speed test from the UK

image

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -2216,7 +2216,6 @@ export class LitNodeClientNodeJs
         // Check if IPFS options are provided and if the code should be fetched from IPFS and overwrite the current code.
         // This will fetch the code from the specified IPFS gateway using the provided ipfsId,
         // and update the params with the fetched code, removing the ipfsId afterward.
-
         const overwriteCode =
           params.ipfsOptions?.overwriteCode ||
           GLOBAL_OVERWRITE_IPFS_CODE_BY_NETWORK[this.config.litNetwork];
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -965,8 +965,10 @@ export class LitNodeClientNodeJs
    * @returns The base64-encoded fallback IPFS code.
    * @throws An error if the code retrieval fails.
    */
-
-  async _getFallbackIpfsCode(gatewayUrl: string | undefined, ipfsId: string) {
+  private async _getFallbackIpfsCode(
+    gatewayUrl: string | undefined,
+    ipfsId: string
+  ) {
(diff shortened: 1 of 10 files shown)


packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -955,13 +957,46 @@ export class LitNodeClientNodeJs
...
+  /**
+   * Retrieves the fallback IPFS code for a given IPFS ID.
+   *
+   * @param gatewayUrl - the gateway url.
+   * @param ipfsId - The IPFS ID.
+   * @returns The base64-encoded fallback IPFS code.
+   * @throws An error if the code retrieval fails.
+   */
+
+  async _getFallbackIpfsCode(gatewayUrl: string | undefined, ipfsId: string) {
```
FedericoAmura, line 969: Shouldn't this be `private`?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2190,6 +2224,26 @@ export class LitNodeClientNodeJs
...
         }
 
+        // Check if IPFS options are provided and if the code should be fetched from IPFS and overwrite the current code.
+        // This will fetch the code from the specified IPFS gateway using the provided ipfsId,
+        // and update the params with the fetched code, removing the ipfsId afterward.
+        const overwriteCode =
+          params.ipfsOptions?.overwriteCode ||
+          GLOBAL_OVERWRITE_IPFS_CODE_BY_NETWORK[this.config.litNetwork];
+
+        if (overwriteCode && props.litActionIpfsId) {
```
joshLong145, line 2234: Will the fallback work in the case where the `cid` is not one of our wrapped keys implementations?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: Yup! Good catch! :))

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
Ansonhkg: Yea, it will work on all CIDs now instead of just "whitelisting" the wrapped key lit actions, cus not only the wrapped key component is affected.

joshLong145 (approved):

FedericoAmura (approved):
lgtm

PR #611, opened 2024-08-23
Fix request params sent to nodes

joshLong145:
Fixes an issue where we send the `sessionSigs` map to all nodes when performing network operations where we should only be sending a single value as the `authSig` from this mapping.

This fix is a stop gap as we need a `ParamsValidator` which will perform the following operations
- Check that all properties required for a network operation are defined
    - If all required properties are not found throw an exception
-  Perform basic type assertion that required properties are of the correct type
- Check for properties which are not relevant for a given network operation and remove them


FedericoAmura (approved):
Should work as a hotfix. We can also create better types for `SendNodeCommand` that improve this apart from the `ParamsValidator`

Ansonhkg (approved):

PR #579, opened 2024-08-06
Feature/lit 3580 refactor js sdk change all enums to constants

FedericoAmura:
# Description

This PR changes all the enums used around the SDK to literal types and their derived types in order to improve type and values definitions and their imports on other packages

## Migration guide

Several enums have been moved or replaced by others in `@lit-protocol/constants` as constant values that behave similarly to enums.
They also export types representing their keys and values, with suffixes `_TYPE` and `_VALUE` respectively in order to be able to use them in type-safe manner.

For example, the `LitNetwork` enum has been replaced by `LIT_NETWORK` constant, with the corresponding `LIT_NETWORK_TYPES` and `LIT_NETWORK_VALUES` types.

The following table gives a detail of all the constants changed or removed.

| Old package                  | Value                            | Update                                                                                                                         |
|------------------------------|----------------------------------|--------------------------------------------------------------------------------------------------------------------------------|
| `@lit-protocol/auth-browser` | `WALLET_ERROR`                   | Replaced enum by a constant.                                                                                                   |
| `@lit-protocol/auth-helpers` | `LitAbility`                     | Replaced by constant `LIT_ABILITY` in constants package.                                                                       |
| `@lit-protocol/auth-helpers` | `LitNamespace`                   | Replaced by constant `LIT_NAMESPACE` in constants package.                                                                     |
| `@lit-protocol/auth-helpers` | `LitRecapAbility`                | Replaced by constant `LIT_RECAP_ABILITY` in constants package.                                                                 |
| `@lit-protocol/auth-helpers` | `LitResourcePrefix`              | Replaced by constant `LIT_RESOURCE_PREFIX` in constants package.                                                               |
| `@lit-protocol/constants`    | `AuthMethodScope`                | Replaced by constant `AUTH_METHOD_SCOPE` in constants package.                                                                 |
| `@lit-protocol/constants`    | `AuthMethodType`                 | Replaced by constant `AUTH_METHOD_TYPE` in constants package.                                                                  |
| `@lit-protocol/constants`    | `CENTRALISATION_BY_NETWORK`      | Does not include `localhost` anymore. Use `CENTRALISATION_BY_NETWORK.Custom` instead.                                          |
| `@lit-protocol/constants`    | `EITHER_TYPE`                    | Replaced by constant `EITHER_TYPE` in constants package.                                                                       |
[...]

PR #622, opened 2024-08-28
Feature/lit 3795 review audit dependencies

FedericoAmura:
# Description

This PR checks our dependencies and removes a bunch of them, updates others and adds depd to loudly notify deprecations to our consumers

# Changelog
- updated dependencies versions
- removed some dependencies not used
- cleaned repeated dependencies inside packages
- removed autogen_internal (not used anymore)
- updated Typescript (also on #624, was needed there). Also unified `moduleResolution` and `esModuleInterop`
- removed @cosmjs dependencies from contracts-sdk, those are heavy and were imported just for a cosmos address calculation (they are still in pkp-cosmos but that is a leaf package, so now they will be included only when really using cosmos)
- removed bitcoinjs-lib
- removed secp256k1
- added depd to deprecate all the things loudly
- moved `addresses` file and included functions from `contracts-sdk` to `misc` (`publicKeyConvert` and `derivedAddresses`)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


packages/contracts-sdk/src/lib/addresses.ts
```diff
@@ -22,6 +22,164 @@ export interface TokenInfo {
...
 }
 
+/**
+ * Converts a public key between compressed and uncompressed formats.
+ *
+ * @param publicKey - Public key as a Buffer (33 bytes compressed or 65 bytes uncompressed)
+ * @param compressed - Boolean flag indicating whether the output should be compressed
+ * @returns Converted public key as a Buffer
+ */
+export function publicKeyConvert(
```
joshLong145, line 32: Do we need to export this method? Seems like an internal util

packages/contracts-sdk/src/lib/addresses.ts
```diff
@@ -22,6 +22,164 @@ export interface TokenInfo {
...
+  publicKey: Buffer,
+  compressed: boolean = true
+): Buffer {
+  if (compressed) {
+    // Compress the public key (if it's not already compressed)
+    if (publicKey.length === 65 && publicKey[0] === 0x04) {
+      const x = publicKey.subarray(1, 33);
+      const y = publicKey.subarray(33, 65);
+      const prefix = y[y.length - 1] % 2 === 0 ? 0x02 : 0x03;
+      return Buffer.concat([Buffer.from([prefix]), x]);
```
joshLong145, line 42: nit: Ideally we move away from `Buffer` to `TextEncoder` as `Buffer` is a node utility. Can we refactor this if possible?

packages/contracts-sdk/src/lib/addresses.ts
FedericoAmura: Yes, it is also used in [pkp-base](https://github.com/LIT-Protocol/js-sdk/pull/622/files#diff-d5bbc3deae289e4022c8b5430b1b455cdb2e4c887da2ce9f39e61a4ef2d6ba4c)

joshLong145: Do you think it makes sense to move these to `misc` as we already import that package in both `contracts-sdk` and `pkp-base`

joshLong145: My main concern is introducing `contracts-sdk` as a dependency to `pkp-base` currently these are the main imports from that package


```
import { publicKeyConvert } from 'secp256k1';

import { LitNodeClient } from '@lit-protocol/lit-node-client';
import { logError } from '@lit-protocol/misc';
import {
  AuthenticationProps,
  JsonExecutionSdkParams,
  PKPBaseProp,
  AuthSig,
  PKPBaseDefaultParams,
  SigResponse,
  RPCUrls,
  AuthMethod,
  SessionSigsMap,
} from '@lit-protocol/types';
``` 

I am wondering if we can strip all deps from these new key converters that are not part of the runtime api. such that they can be migrated without bringing over dependencies. I am a bit worried about making `contracts-sdk` a required dep as it will be rather bloated with the ABI package requirement

FedericoAmura: The import path was already `pkp-base` -> `lit-node-client` -> `lit-node-client-nodejs` -> `core` -> `contracts-sdk` so this is not really creating the link between the two

You mean implementing `bech32`, `bs58`, etc. ourselves?

joshLong145: Sorry I should have mentioned that I was assuming the refactor of `core` to no longer depend on `contracts-sdk`
Ideally, as we want to break the coupling of `core` and `contracts-sdk`  since we really only use it to resolve the staking contract. If we bind to it here for these utilities we end up kind of back at a similar points.

FedericoAmura: Oh got it now, considering that makes sense. Moving...

packages/constants/src/lib/constants/constants.ts
```diff
@@ -621,7 +625,18 @@ export const LIT_CHAINS: LITChain<LITEVMChain> = {
 /**
  * @deprecated Will be removed in version 7.x. - This is using the OLD chornicle testnet. `LIT_CHAINS['chronicleTestnet']` instead, or use `LIT_CHAINS['yellowstone']` for the new Chronicle Yellowstone Testnet (Jul 2024). (Updated to use `yellowstone` chain instead 22 July 2024)
  */
-export const LIT_CHAIN_RPC_URL = LIT_CHAINS['chronicleTestnet'].rpcUrls[0];
+const LIT_CHAIN_RPC_URL_TARGET = {
+  url: LIT_CHAINS['chronicleTestnet'].rpcUrls[0],
+};
```
Ansonhkg, line 630: I think we can just remove this as we already have this comment?
```
@deprecated Will be removed in version 7.x.
```

packages/constants/src/lib/constants/constants.ts
```diff
@@ -974,31 +1003,78 @@ export const PKP_CLIENT_SUPPORTED_CHAINS = ['eth', 'cosmos'];
...
 
+const RELAY_URL_CAYENNE_TARGET = {
+  url: 'https://relayer-server-staging-cayenne.getlit.dev',
+};
 /**
  * @deprecated Will be removed in version 7.x. - Use RELAYER_URL_BY_NETWORK.Cayenne instead
  */
-export const RELAY_URL_CAYENNE =
-  'https://relayer-server-staging-cayenne.getlit.dev';
+export const RELAY_URL_CAYENNE = new Proxy(RELAY_URL_CAYENNE_TARGET, {
```
Ansonhkg, line 1012: I think we can just remove this as we already have this comment?
```
@deprecated Will be removed in version 7.x.
```

Ansonhkg (commented):
Should we remove the constants that have:
```
@deprecated Will be removed in version 7.x.
```

in the jsDoc?

Ansonhkg:
Need to resolve conflicts too

packages/constants/src/lib/constants/constants.ts
FedericoAmura: Removed

packages/constants/src/lib/constants/constants.ts
FedericoAmura: Removed

FedericoAmura:
> Should we remove the constants that have:
> 
> ```
> @deprecated Will be removed in version 7.x.
> ```
> 
> in the jsDoc?

Removed them. And left the ones that are really renaming to ease the transition into enums #579

Ansonhkg (approved):
lgtm

PR #624, opened 2024-09-02
feat: implemented node handshake retry logic over v7 branch

FedericoAmura:
# Description

This PR:
- updates the "connected to X of Y nodes, with a minimum of Z" as that was misleading when X>=Z
- adds retry logic to each node handshake request on certain scenarios (with a flag to [de]activate it, and retries until handshake timeout is reached). Defaults to `true`
- adds a per node HANDSHAKE timeout, configurable. (apart from full handshake timeout). This one overrides the request timeout if present. Defaults to `7500`ms.
- adds per node REQUEST timeout (applies also to handshake if present and handshake one is not), configurable. Defaults to `35000`ms (because LitActions have a 30s timeout, plus some network lag)
- reports correctly why the handshake failed using new VErrors

With default configs, the handshake process will attempt to connect to each node, the ones that don't succeed before 7500ms will be cancelled and retried in loop until they all succeed or handshake timeout is reached
When retry is set to false, the first failing node will fail the whole process (can be changed for naga allowing X=>Z handshakes)

Screenshot from 2024-09-02 21-30-39

Bonus of this PR:
Updates TS to 5.4 in order to use `AbortSignal.timeout` and not have to wait until #622 and then need a followup

## Type of change

# How Has This Been Tested?

- [x] Manually tested several configs with a demo script and playing around with configs

# Checklist:

### DIFF
--- a/packages/core/src/lib/lit-core.ts
+++ b/packages/core/src/lib/lit-core.ts
@@ -1,3 +1,5 @@
+// @ts-expect-error No types available for this package
+import { VError } from '@openagenda/verror';
 import { ethers } from 'ethers';
 
 import {
@@ -18,25 +20,26 @@ import {
   CENTRALISATION_BY_NETWORK,
   HTTP,
   HTTPS,
+  InitError,
+  InvalidArgumentException,
+  InvalidEthBlockhash,
+  InvalidNodeAttestation,
+  InvalidParamType,
   LIT_CURVE,
   LIT_CURVE_VALUES,
   LIT_ENDPOINT,
   LIT_ERROR_CODE,
   LIT_NETWORK,
   LIT_NETWORKS,
+  LitNodeClientBadConfigError,
+  LitNodeClientNotReadyError,
+  NetworkError,
+  NodeError,
   RPC_URL_BY_NETWORK,
   STAKING_STATES,
   STAKING_STATES_VALUES,
-  version,
-  InitError,
-  InvalidParamType,
-  NodeError,
   UnknownError,
-  InvalidArgumentException,
-  LitNodeClientBadConfigError,
-  InvalidEthBlockhash,
-  LitNodeClientNotReadyError,
-  InvalidNodeAttestation,
+  version,
 } from '@lit-protocol/constants';
 import { LitContracts } from '@lit-protocol/contracts-sdk';
 import { checkSevSnpAttestation, computeHDPubKey } from '@lit-protocol/crypto';
@@ -64,7 +67,6 @@ import {
   NodeClientErrorV0,
   NodeClientErrorV1,
   NodeCommandServerKeysResponse,
-  NodeErrorV3,
   RejectedNodePromises,
   SendNodeCommand,
   SessionSigsMap,
@@ -103,7 +105,15 @@ export type LitNodeClientConfigWithDefaults = Required<
   >
 > &
   Partial<
-    Pick<LitNodeClientConfig, 'storageProvider' | 'contractContext' | 'rpcUrl'>
+    Pick<
+      LitNodeClientConfig,
+      | 'storageProvider'
+      | 'contractContext'
+      | 'rpcUrl'
+      | 'nodeConnectTimeout'
+      | 'nodeRequestTimeout'
+      | 'retryNodeHandshake'
+    >
   > & {
     bootstrapUrls: string[];
   } & {
@@ -126,7 +136,10 @@ export class LitCore {
   config: LitNodeClientConfigWithDefaults = {
     alertWhenUnauthorized: false,
     debug: true,
-    connectTimeout: 20000,
+    connectTimeout: 20_000, // Default for full handshake
+    nodeConnectTimeout: 7_500, // Default for handshake per node
+    nodeRequestTimeout: 35_000, // LitActions have a 30s timeout + some buffer for network latency
+    retryNodeHandshake: true,
     checkNodeAttestation: false,
     litNetwork: 'cayenne', // Default to cayenne network. will be replaced by custom config.
     minNodeCount: 2, // Default value, should be replaced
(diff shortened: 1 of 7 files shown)


packages/core/src/lib/lit-core.ts
```diff
@@ -645,31 +658,78 @@ export class LitCore {
     const serverKeys: Record<string, JsonHandshakeResponse> = {};
 
     let timeoutHandle: ReturnType<typeof setTimeout>;
+    let canRetry = this.config.retryNodeHandshake ?? true;
```
joshLong145, line 661: Do we want this defaulted to `true`?

packages/core/src/lib/lit-core.ts
```diff
@@ -645,31 +658,78 @@ export class LitCore {
...
+    };
+
+    const handshakeWithNodeWithRetry = async (url: string): Promise<void> => {
+      try {
+        return await handshakeWithNode(url);
+      } catch (e) {
+        const errorInfo = VError.info(e);
+        // Only retry requests if there was a 5xx error or a timeout error (not client side errors)
+        const isRetryableRequest =
+          errorInfo.statusCode >= 500 ||
```
joshLong145, line 678: we might also want to avoid retrying `404` `429` etc.

packages/core/src/lib/lit-core.ts
```diff
@@ -943,6 +1003,11 @@ export class LitCore {
       body: JSON.stringify(data),
     };
 
+    const requestTimeout = timeout || this.config.nodeRequestTimeout;
+    if (requestTimeout) {
+      req.signal = AbortSignal.timeout(requestTimeout);
```
joshLong145, line 1008: Nice :)

packages/encryption/src/lib/encryption.ts
```diff
@@ -128,14 +128,10 @@ export const encryptToJson = async (
...
-  params: T
+export async function decryptFromJson(
+  params: DecryptFromJsonProps
 ): Promise<
-  T extends { parsedJsonData: { dataType: 'file' } }
-    ? ReturnType<typeof decryptToFile>
-    : T extends { parsedJsonData: { dataType: 'string' } }
-    ? ReturnType<typeof decryptToString>
-    : never
+  ReturnType<typeof decryptToFile> | ReturnType<typeof decryptToString>
```
joshLong145, line 134: Nice change. much simpler to follow.

packages/core/src/lib/lit-core.ts
FedericoAmura: It is avoiding those. It is flagging the request as retryable only for 5xx errros

packages/core/src/lib/lit-core.ts
FedericoAmura: If we go with timeouting each node at 7.5 seconds and the full handshake at 20, I would say we do (trying almost 3 times per node)
The other option would be to retain the old default of just the handshake at 20s, and nodes individually do not timeout

I'm open for discussion on this, we could even make them mandatory to force the dev into thinking what to do. My numbers were thinking in one of our customers

packages/core/src/lib/lit-core.ts
joshLong145: Ah I thought this was inverted. Thanks for the clarification.

packages/core/src/lib/lit-core.ts
joshLong145: Interesting. I think we should keep the timeout global for all node handshakes at `20s` and allow users to explicitly turn on `retryNodeHandshake` main reason is many users actually use our sdk in CI which can cause a cascade of retry ops to the nodes which might cause per issues due to request buffer back pressure. 


But if we do not feel this is a valid concern then we can leave the default. This is just something ive noticed when we had the older retry logic.

FedericoAmura: Good point, I think however we should be safe enough of DDOSing ourselves if we set a high enough timeout. Kind of targetting the top 5% anomalies of usage
I asked @anshss in slack what numbers he got in the testing he made last days

joshLong145: This makes sense to me.

packages/core/src/lib/lit-core.ts
```diff
@@ -943,6 +1003,11 @@ export class LitCore {
...
+    }
+
     return sendRequest(url, req, requestId);
```
Ansonhkg, line 1011: should this be `await`?

packages/core/src/lib/lit-core.ts
FedericoAmura: For what I've seen with Ansh I feel like 7.5s is still a safe threshold to avoid retying too much
I'm also not a big fan of releasing turned off things that will likely go unnoticed

But on the other side, I think this will change on the near future with naga once we settle just with the threshold and/or convert lit core to an event emitter. So it is a change that will be changed in a not-so-distant future. Thus forcing our consumers handle breaking changes two times in a close timeframe

I now don't have a preference on the default value of this. What do you think?

packages/core/src/lib/lit-core.ts
FedericoAmura: The function is `async` so it is not needed

Ansonhkg: ah you're right

packages/core/src/lib/lit-core.ts
Ansonhkg: Linking it here for internal:
https://litprotocol.slack.com/archives/C07DNQVC0FR/p1726051959979759?thread_ts=1725035360.246599&cid=C07DNQVC0FR

DashKash54:
Based on our discussion today do we still need additional retries on the SDK to solve fundamentally node problems? @MaximusHaximus @Ansonhkg @FedericoAmura @joshLong145

PR #630, opened 2024-09-10
feat: include wrapped keys lit actions code in the bundle

FedericoAmura:
# Description

This PR adds a new `@lit-protocol/wrapped-keys-bc` package derived from the original `@lit-protocol/wrapped-keys-bc` but with the lit actions code embedded and preferred over sending the IPFS CID. This gives customer the option to choose between using IPFS with the original package or using the code directly with the trade off of installation size

The new package is derived and built from the original, in a cloning & adapting style. So both packages will always be in-sync (as long as the included lit actions and their CIDs are)

## Type of change

# How Has This Been Tested?

- [x] Integration tests pass (manually run, not included in CI to avoid code duplication. CI runs the original/CID version of wrapped keys)
- [x] Manually using [this demo](https://github.com/anshss/wrapped-keys-test/tree/main) and the published [v0.0.2 beta](https://www.npmjs.com/package/@lit-protocol/wrapped-keys-bc/v/0.0.2)

# Checklist:


packages/wrapped-keys/esbuild.config.js
```diff
@@ -1,7 +1,45 @@
...
+      if (result.errors.length > 0) {
+        console.error('Build failed with errors:', result.errors);
+        return;
+      }
+
+      result.outputFiles.forEach((outputFile) => {
+        let content = outputFile.text;
+
+        // Use JSON.stringify to safely encode the content
+        const wrappedContent = `module.exports = ${JSON.stringify(content)};`;
```
FedericoAmura, line 28: Importing a JS file as a string is totally not worth it when you can just wrap it with quotes when building it. Specially on both node.js and browser envs

packages/wrapped-keys/src/lib/generated/litActions/common/exportPrivateKey.js
```diff
@@ -0,0 +1,2 @@
+// DO NOT EDIT THIS FILE. IT IS GENERATED ON BUILD. RUN `yarn bundle` IN THE ROOT DIRECTORY TO UPDATE THIS FILE.
+module.exports = "";
```
FedericoAmura, line 2: These are dummy files. The lit actions build step will populate all these with the actual lit action code that is sent to the nodes, instead of the IPFS CID

In the bundle without the lit actions code this will not satisfy the `if (!litActionCode) {` condition and therefore the IPFS CID will be used

packages/wrapped-keys/src/lib/generated/litActions/common/exportPrivateKey.js
```diff
@@ -0,0 +1,2 @@
+// DO NOT EDIT THIS FILE. IT IS GENERATED ON BUILD. RUN `yarn bundle` IN THE ROOT DIRECTORY TO UPDATE THIS FILE.
+module.exports = '';
```
FedericoAmura, line 2: These are dummy files. The lit actions build step will populate all these with the actual lit action code that is sent to the nodes, instead of the IPFS CID

In the bundle without the lit actions code this will not satisfy the if (!litActionCode) { condition and therefore the IPFS CID will be used

packages/wrapped-keys/project.json
```diff
@@ -11,7 +11,14 @@
...
         "main": "packages/wrapped-keys/src/index.ts",
         "tsConfig": "packages/wrapped-keys/tsconfig.lib.json",
-        "assets": ["packages/wrapped-keys/*.md"],
+        "assets": [
+          "packages/wrapped-keys/*.md",
+          {
+            "glob": "**/*",
+            "input": "packages/wrapped-keys/src/lib/generated/litActions",
+            "output": "src/lib/generated/litActions"
+          }
```
DashKash54, line 20: Why is this explicitly required to be defined as all these files are already `.js`?

packages/wrapped-keys/esbuild.config.js
DashKash54, line 1: We plan to support both the new bundled Lit Action as well as IPFS Cid fetching for apps that are sensitive to large SDK bundle sizes. Given this, shouldn't the bundler output both the existing files in `/dist` and the new ones in `generated/litActions`?

DashKash54 (commented):
Looks good! Left a couple comments
"you can just wrap it with quotes when building it"- nice idea @FedericoAmura

packages/wrapped-keys/project.json
DashKash54: Is the idea that we run the bundler script separately for the `@lit-protocol/wrapped-keys-bc` release? And for `@lit-protocol/wrapped-keys` these files don't exist so they won't be included in the final build hence it's required to explicitly define it here?

packages/wrapped-keys/esbuild.config.js
joshLong145: Do you think importing is worth it from the perspective of tree shaking?
Perhaps we use named exports so when the package is consumed unused action can be tree shook by the users bundler.

packages/wrapped-keys/src/lib/generated/litActions/common/exportPrivateKey.js
joshLong145: Perhaps we generate these files through some manifest which will tell a script what files to generate as a pre build step. So we do not need explicit files in the folder structure created by hand. This will also allow us to git ignore these skeleton files.

packages/wrapped-keys/src/lib/generated/litActions/solana/signTransactionWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,2 @@
+// DO NOT EDIT THIS FILE. IT IS GENERATED ON BUILD. RUN `yarn bundle` IN THE ROOT DIRECTORY TO UPDATE THIS FILE.
+module.exports = '';
```
joshLong145, line 2: Same as above

packages/wrapped-keys/src/lib/generated/litActions/solana/signMessageWithSolanaEncryptedKey.js
```diff
@@ -0,0 +1,2 @@
+// DO NOT EDIT THIS FILE. IT IS GENERATED ON BUILD. RUN `yarn bundle` IN THE ROOT DIRECTORY TO UPDATE THIS FILE.
+module.exports = '';
```
joshLong145, line 2: Same as above

packages/wrapped-keys/src/lib/generated/litActions/solana/generateEncryptedSolanaPrivateKey.js
```diff
@@ -0,0 +1,2 @@
+// DO NOT EDIT THIS FILE. IT IS GENERATED ON BUILD. RUN `yarn bundle` IN THE ROOT DIRECTORY TO UPDATE THIS FILE.
+module.exports = '';
```
joshLong145, line 2: Same as above

packages/wrapped-keys/src/lib/generated/litActions/ethereum/signTransactionWithEthereumEncryptedKey.js
```diff
@@ -0,0 +1,2 @@
+// DO NOT EDIT THIS FILE. IT IS GENERATED ON BUILD. RUN `yarn bundle` IN THE ROOT DIRECTORY TO UPDATE THIS FILE.
+module.exports = '';
```
joshLong145, line 2: Same as above

packages/wrapped-keys/src/lib/generated/litActions/ethereum/signMessageWithEthereumEncryptedKey.js
```diff
@@ -0,0 +1,2 @@
+// DO NOT EDIT THIS FILE. IT IS GENERATED ON BUILD. RUN `yarn bundle` IN THE ROOT DIRECTORY TO UPDATE THIS FILE.
+module.exports = '';
```
joshLong145, line 2: Same as above

packages/wrapped-keys/src/lib/generated/litActions/ethereum/generateEncryptedEthereumPrivateKey.js
```diff
@@ -0,0 +1,2 @@
+// DO NOT EDIT THIS FILE. IT IS GENERATED ON BUILD. RUN `yarn bundle` IN THE ROOT DIRECTORY TO UPDATE THIS FILE.
+module.exports = '';
```
joshLong145, line 2: Same as above

packages/wrapped-keys/src/lib/lit-actions-client/utils.ts
```diff
@@ -48,3 +51,12 @@ export function postLitActionValidation(
 export function getLitActionCid(network: Network, actionType: LitActionType) {
   return LIT_ACTION_CID_REPOSITORY[actionType][network];
 }
+
+export function getLitActionCode(
+  network: Network,
+  actionType: LitActionType
+): string | undefined {
+  const litActionCode = LIT_ACTION_BUNDLED_CODE_PATH[actionType][network];
```
joshLong145, line 59: should we check if `litActionCode` is defined before returning?

packages/wrapped-keys/src/lib/generated/litActions/common/exportPrivateKey.js
FedericoAmura: This is what I want to achieve with the double bundle process. Some script of something like that to clone the wrapped-keys nx project and generate the expanded one and to minimize the overhead of maintaining another thing

packages/wrapped-keys/esbuild.config.js
FedericoAmura: Yes, this is why we need two different packages. One with the files with empty strings and one with the stringified versions of lit actions

Esbuild replaces files, and both wrapped-keys bundles will be identical except that one will have files with code where the other does not. Therefore, the flow is build&publish-no-LA-version -> esbuild-replacing-files -> build&publish-LA-version

packages/wrapped-keys/src/lib/generated/litActions/common/exportPrivateKey.js
FedericoAmura: The files have to exists before running anything, at dev time. If they are not there, the shared code will fail when trying to import those files and DX will be seriously degraded

packages/wrapped-keys/src/lib/lit-actions-client/utils.ts
FedericoAmura: And throw if it is not? It is kind of expected that you can get an falsy value (empty string). That is even normal operation when the bundle does not include them and you have to use the CID
This way we can define just a subset of bundled LAs because the querying is the same than for CID. But it also makes sense to throw now

Let me know if this is what you were thinking or I misunderstood

packages/wrapped-keys/src/lib/generated/litActions/common/exportPrivateKey.js
joshLong145: I see, but can't a script generate these file stubs beforehand? So the DX is not affected? Maybe I am missing something.

packages/wrapped-keys/src/lib/lit-actions-client/utils.ts
joshLong145: Yes this is my thought process. I think throwing is the best option since returning `undefined` will probably cause undefined behaviors from the caller perspective.

packages/wrapped-keys/src/lib/generated/litActions/common/exportPrivateKey.js
FedericoAmura: Why do you want a script for that? And when do you want to run it?

packages/wrapped-keys/src/lib/lit-actions-client/utils.ts
FedericoAmura: Updated. Also added a check for having one or the other. lit-node-client does this, but we can already validate it here

packages/wrapped-keys/esbuild.config.js
FedericoAmura: Changed no named exports

packages/wrapped-keys/esbuild.config.js
```diff
@@ -22,15 +68,19 @@ const esbuild = require('esbuild');
...
+    plugins: [wrapIIFEInStringPlugin],
+    write: false,
   });
   await esbuild.build({
     entryPoints: ['./src/lib/litActions/common/src/exportPrivateKey.js'],
     bundle: true,
     minify: true,
     sourcemap: false,
-    outdir: './src/lib/litActions/common/dist',
+    outdir: './src/lib/generated/litActions/common',
```
Ansonhkg, line 81: good renaming!

packages/wrapped-keys/src/lib/lit-actions-client/utils.ts
```diff
@@ -48,3 +51,50 @@ export function postLitActionValidation(
...
+}
+
+/**
+ * Fetch the Lit action code or its IPFS CID for a given network and action type.
+ *
+ * @param {Network} network The network to get the code or CID for.
+ * @param {LitActionType} actionType The type of action to get the code or CID for.
+ * @returns {{ litActionCode?: string, litActionIpfsCid?: string }} The Lit action code or its IPFS CID.
+ */
+export function getLitActionCodeOrCid(
```
Ansonhkg, line 76: Just want to note for readers that we have introduced a new param in the `executeJs` function where we could overwrite the LitAction `code` property.

https://github.com/LIT-Protocol/js-sdk/blob/816aa6147a3863efa18ba90e0bbf23f0974adae9/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts#L1047-L1065

But in this case we are "pulling content" for a static code, so this param is irrelevant.

Ansonhkg (approved):
lgtm!

MaximusHaximus (approved):
Just one small change to fix up type naming,  otherwise this LGTM. Mentioned a few FYI's but nothing really blocking.

packages/wrapped-keys/esbuild.config.js
```diff
@@ -1,7 +1,45 @@
...
+  }
+  ensureDirectoryExistence(dirname);
+  fs.mkdirSync(dirname);
+};
+
+const wrapIIFEInStringPlugin = {
+  name: 'wrap-iife-in-string',
+  setup(build) {
+    build.onEnd((result) => {
+      if (result.errors.length > 0) {
```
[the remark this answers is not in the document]
FedericoAmura: Great safety check. Added

packages/wrapped-keys/src/lib/lit-actions-client/constants.ts
```diff
@@ -19,4 +26,23 @@ const LIT_ACTION_CID_REPOSITORY: LitCidRepository = Object.freeze({
   }),
 });
 
-export { LIT_ACTION_CID_REPOSITORY };
+const LIT_ACTION_BUNDLED_CODE_PATH: LitCidRepository = Object.freeze({
```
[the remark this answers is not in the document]
FedericoAmura: Good! Done

joshLong145 (approved):

PR #633, opened 2024-09-12
fix: add support for different os types for the sed shell script command

Ansonhkg:
Add support for different OS type when executing `sed` command in a shell script. Previously it was only working on Linux (GNU sed), but we need it to work on macOs as well (BSD sed)


MaximusHaximus (approved):
:+1:

FedericoAmura (approved):

PR #628, opened 2024-09-06
Feature/lit 3701 js sdk refactor contract sdk ethers client bootstrapping

joshLong145:
# Description

Refactors `LitContracts` to  allow for optional standup of contract clients  based on the configuration of the `ContractContext`

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #649, opened 2024-09-24
feat: add type improvement

Ansonhkg:
# Description

This PR is a suggestion for https://github.com/LIT-Protocol/js-sdk/pull/628

Add new type for the contract names instead of string[]

image


joshLong145:
Makes sense, I wonder if we should make a constant list of ContractNames and use `key of` But can try this out in my branch after merging this.

joshLong145 (approved):

PR #651, opened 2024-09-24
LIT-3896 - Better error logs & request id tracking and submission for wrapped-keys service client

MaximusHaximus:
# Description

- Updated logging logic in wrapped-keys so that we now log HTTP error codes for failed requests, and `error.cause` when it is defined.
- `error.cause` should give us insights into the _reason_ why a `fetch failed` error is thrown; previously all we saw in logs was the string `fetch failed`, because fetch puts the really useful info in a `cause` property on the error
- Added requestId generation and logging to the wrapped-keys service client.  Error logs now include our _internal_ request id, and we send that request ID to the API backend in the `x-correlation-id` header, which provides integration with cloudwatch and `pino-lambda` (for service logs) -- It was just an added benefit that it was also the simplest way (identical logic for GET and POST routes)

Example output from `testFailImportWrappedKeysWithSamePrivateKey` test:
image

## Type of change

# How Has This Been Tested?

- [ ] Ran local-tests against `datil-test` instances of the lambda functions; verified functionality is working as expected

# Checklist:

### DIFF
--- a/packages/wrapped-keys/src/lib/service-client/client.ts
+++ b/packages/wrapped-keys/src/lib/service-client/client.ts
@@ -19,20 +19,18 @@ export async function listPrivateKeyMetadata(
 ): Promise<StoredKeyMetadata[]> {
   const { litNetwork, sessionSig } = params;
 
+  const requestId = generateRequestId();
   const { baseUrl, initParams } = getBaseRequestParams({
     litNetwork,
     sessionSig,
     method: 'GET',
+    requestId,
   });
 
   const pkpAddress = getPkpAddressFromSessionSig(sessionSig);
 
-  const requestId = generateRequestId();
-  const url = new URL(`${baseUrl}/${pkpAddress}`);
-  url.searchParams.set('requestId', requestId);
-
   return makeRequest<StoredKeyMetadata[]>({
-    url: url.toString(),
+    url: `${baseUrl}/${pkpAddress}`,
     init: initParams,
     requestId,
   });
@@ -49,19 +47,17 @@ export async function fetchPrivateKey(
 ): Promise<StoredKeyData> {
   const { litNetwork, sessionSig, id } = params;
 
+  const requestId = generateRequestId();
   const { baseUrl, initParams } = getBaseRequestParams({
     litNetwork,
     sessionSig,
     method: 'GET',
+    requestId,
   });
   const pkpAddress = getPkpAddressFromSessionSig(sessionSig);
 
-  const requestId = generateRequestId();
-
-  const url = new URL(`${baseUrl}/${pkpAddress}/${id}`);
-  url.searchParams.set('requestId', requestId);
   return makeRequest<StoredKeyData>({
-    url: url.toString(),
+    url: `${baseUrl}/${pkpAddress}/${id}`,
     init: initParams,
     requestId,
   });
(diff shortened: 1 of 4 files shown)


packages/wrapped-keys/src/lib/service-client/utils.ts
```diff
@@ -104,29 +105,40 @@ async function getResponseJson<T>(response: Response): Promise<T | string> {
   }
 }
 
+export function generateRequestId(): string {
+  return Math.random().toString(16).slice(2);
+}
```
Ansonhkg, line 110: Is the idea to keep the utility function within the `wrapped-keys` package instead of reusing the one in `lit-core`, cur should `wrapped-keys` and `lit-core` import it from the `misc` package?

packages/wrapped-keys/src/lib/service-client/client.ts
```diff
@@ -19,17 +19,20 @@ export async function listPrivateKeyMetadata(
 ): Promise<StoredKeyMetadata[]> {
   const { litNetwork, sessionSig } = params;
 
+  const requestId = generateRequestId();
```
DashKash54, line 22: In the future can we unify the `requestId` being sent to the node with the one sent to the AWS infra? @Ansonhkg

packages/wrapped-keys/src/lib/service-client/utils.ts
```diff
@@ -104,29 +105,40 @@ async function getResponseJson<T>(response: Response): Promise<T | string> {
...
+  try {
+    const response = await fetch(url, { ...init });
 
-  if (!response.ok) {
-    const errorMessage = await getResponseErrorMessage(response);
-    throw new Error(`Failed to make request for wrapped key: ${errorMessage}`);
-  }
+    if (!response.ok) {
+      const errorMessage = await getResponseErrorMessage(response);
+      throw new Error(`HTTP(${response.status}): ${errorMessage}`);
```
DashKash54, line 126: It's good to keep a common searchable text for all failed Wrapped Key operations?
```suggestion
      throw new Error(`HTTP(${response.status}): Failed to make request for wrapped key: ${errorMessage}`);
```

packages/wrapped-keys/src/lib/service-client/utils.ts
```diff
@@ -104,29 +105,40 @@ async function getResponseJson<T>(response: Response): Promise<T | string> {
...
+    }
 
-  return result;
+    return result;
+  } catch (e: unknown) {
+    throw new Error(
+      `Request(${requestId}) for wrapped key failed. Error: ${
+        (e as Error).message
+        // @ts-expect-error Unknown, but `cause` is on `TypeError: fetch failed` errors
+      }${e.cause ? ' - ' + e.cause : ''}`
```
DashKash54, line 141: This should be helpful for identifying the Artillery failures

DashKash54 (approved):
Looks good, left a couple comments

packages/wrapped-keys/src/lib/service-client/utils.ts
MaximusHaximus: I noticed that the method was made `protected` [recently](https://github.com/LIT-Protocol/js-sdk/pull/640/files) and it was simple enough that the duplication didn't feel like an issue worthy enough of making it part of our public API again...

packages/wrapped-keys/src/lib/service-client/client.ts
MaximusHaximus: Unify how? We can't re-use the exact same ID because there could be more than 1 node request id per call to wrapped keys, or vice-versa.  If you mean the method of creating it, see my [comment here](https://github.com/LIT-Protocol/js-sdk/pull/651#discussion_r1774252142)

packages/wrapped-keys/src/lib/service-client/utils.ts
MaximusHaximus: This error is an 'inner' error that is caught below in the block L136-L142.  We will insert this string _into_ the message on the error we throw on L137, which prefixes this string with consistent wording: `Request(${requestId}) for wrapped key failed. Error:`

packages/wrapped-keys/src/lib/service-client/utils.ts
MaximusHaximus: Yup :+1: The HTTP error code plus this ought to give us a clear error reason ❇️

Ansonhkg (approved):
⛴️..

MaximusHaximus:
Thanks, guys <3 :)

PR #652, opened 2024-09-25
fix: wrong encoding - should be `base58` instead of `base64`

Ansonhkg:
# Description

Fix wrong encoding - it should be `base58` instead of `base64`.

As per @spacesailor24 comment for the following Lit Action:

```js
const transaction = Transaction.from(
  Buffer.from(unsignedTransaction.serializedTransaction, 'base64')
);

transaction.sign(solanaKeyPair);
const signature = transaction.signature.toString('base64');

if (broadcast) {
  const solanaConnection = new Connection(
    clusterApiUrl(unsignedTransaction.chain),
    'confirmed'
  );
  const transactionSignature = await solanaConnection.sendRawTransaction(
    transaction.serialize()
  );

  Lit.Actions.setResponse({ response: transactionSignature });
} else {
  Lit.Actions.setResponse({ response: signature });
}
```

When broadcast is true, transactionSignature will be base58 encoded, but when false, signature is base64 encoded

### DIFF
--- a/packages/wrapped-keys-lit-actions/src/lib/solana/signTransactionWithSolanaEncryptedKey.js
+++ b/packages/wrapped-keys-lit-actions/src/lib/solana/signTransactionWithSolanaEncryptedKey.js
@@ -6,6 +6,7 @@ const {
 } = require('@solana/web3.js');
 
 const { removeSaltFromDecryptedKey } = require('../utils');
+const { bs58 } = require('bs58');
 
 /**
  *
@@ -80,7 +81,7 @@ const { removeSaltFromDecryptedKey } = require('../utils');
     );
 
     transaction.sign(solanaKeyPair);
-    const signature = transaction.signature.toString('base64');
+    const signature = bs58.encode(transaction.signature);
 
     if (broadcast) {
       const solanaConnection = new Connection(
--- a/packages/wrapped-keys-lit-actions/src/lib/solana/signTransactionWithSolanaEncryptedKey.js
+++ b/packages/wrapped-keys-lit-actions/src/lib/solana/signTransactionWithSolanaEncryptedKey.js
@@ -80,7 +80,7 @@ const { removeSaltFromDecryptedKey } = require('../utils');
     );
 
     transaction.sign(solanaKeyPair);
-    const signature = bs58.encode(transaction.signature);
+    const signature = ethers.utils.base58.encode(transaction.signature);
 
     if (broadcast) {
       const solanaConnection = new Connection(
--- a/packages/wrapped-keys-lit-actions/src/lib/solana/signTransactionWithSolanaEncryptedKey.js
+++ b/packages/wrapped-keys-lit-actions/src/lib/solana/signTransactionWithSolanaEncryptedKey.js
@@ -6,7 +6,6 @@ const {
 } = require('@solana/web3.js');
 
 const { removeSaltFromDecryptedKey } = require('../utils');
-const { bs58 } = require('bs58');
 
 /**
  *


MaximusHaximus (approved):
:+1: good catch!

FedericoAmura:
This made me realize a few things considering these LA maintenance

1. Probably we need a way to update those LAs on publish? And a test to validate that the LA in `wrapped-keys-lit-actions` is the same as the CID in `wrapped-keys``
2. Updating any LA changes its CID, so this change also needs to update that [here](https://github.com/LIT-Protocol/js-sdk/blob/master/packages/wrapped-keys/src/lib/lit-actions-client/constants.ts#L6) with the updated one
3. We also have [this repo](https://github.com/LIT-Protocol/wrapped-key-fallback-service) serving the LAs **unbundled** code in an express app. If that is used I think we could update it to serve the bundled version from the SDK, or unbundled from this repo to avoid duplicating work there

Ansonhkg:
> This made me realize a few things considering these LA maintenance
> 
> 1. Probably we need a way to update those LAs on publish? And a test to validate that the LA in `wrapped-keys-lit-actions` is the same as the CID in `wrapped-keys``
> 2. Updating any LA changes its CID, so this change also needs to update that [here](https://github.com/LIT-Protocol/js-sdk/blob/master/packages/wrapped-keys/src/lib/lit-actions-client/constants.ts#L6) with the updated one
> 3. We also have [this repo](https://github.com/LIT-Protocol/wrapped-key-fallback-service) serving the LAs **unbundled** code in an express app. If that is used I think we could update it to serve the bundled version from the SDK, or unbundled from this repo to avoid duplicating work there
> 
> cc. @MaximusHaximus

re:  ... express app. If that is used ...

No, it’s not used - It was the initial idea for bundling the LA code, but we disregarded it because we don’t want to maintain another microservice.

MaximusHaximus:
> This made me realize a few things considering these LA maintenance
> 
> 1. Probably we need a way to update those LAs on publish? And a test to validate that the LA in `wrapped-keys-lit-actions` is the same as the CID in `wrapped-keys``
> 2. Updating any LA changes its CID, so this change also needs to update that [here](https://github.com/LIT-Protocol/js-sdk/blob/master/packages/wrapped-keys/src/lib/lit-actions-client/constants.ts#L6) with the updated one
> 3. We also have [this repo](https://github.com/LIT-Protocol/wrapped-key-fallback-service) serving the LAs **unbundled** code in an express app. If that is used I think we could update it to serve the bundled version from the SDK, or unbundled from this repo to avoid duplicating work there
> 
> cc. @MaximusHaximus

Yes - we currently handle publishing these as an entirely manual process, but now that this is its own package it would be awesome to get the publishing to IPFS codified into the project scripts for this package.

What do you think about adding publishing script to this package, and during build+publish to NPM, we actually compute the CID of these actions after bundling them, and _generate_ the LitActionCidRepository in _this_ package, then import it into `wrapped-keys` 🤔....? With tree shaking, people using the CID repository wouldn't end up with the action source code in their app.

We could even replicate the `setLitActionCode()` method -- we could import the litActionCidRepository from this package as a default, but allow `setLitActionCids()` to be called by consumers of the primary package to use custom actions without being forced to bundle their custom actions source into their app. 🤔

spacesailor24 (changes requested):
I just found out that ethers includes [base58 en/decoding](https://docs.ethers.org/v5/api/utils/encoding/#Bse58), so we don't have to import `bs58`

Ansonhkg:
> I just found out that ethers includes [base58 en/decoding](https://docs.ethers.org/v5/api/utils/encoding/#Bse58), so we don't have to import `bs58`

Good catch! Thanks!!

PR #654, opened 2024-09-25
feat: remove support for cayenne, habanero and manzano

FedericoAmura:
# Description

This PR removes `cayenne`, `habanero` and `manzano` support from the SDK. They are replaced with `datil` family of networks

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Integration testing

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #656, opened 2024-09-25
feat: remove `ecdsa` feature in lit-node-client-nodejs

Ansonhkg:
# Description

In addition for https://github.com/LIT-Protocol/js-sdk/pull/654, we also need to remove ecdsa from lit-node-client-nodejs


FedericoAmura (approved):
Thanks for completing!

PR #661, opened 2024-09-30
Feature/lit 3911 add request id aggregation to logger when logging is disabled

joshLong145:
Allows for `request Ids` to be tracked if `debug` is not enabled on `LitNodeClient` by allowing logging calls to pass through the `Logger` instance.

Adds time based sorting to the array of `requests Ids` returned by `getRequestIds` on `LitCore`

### DIFF
--- a/packages/logger/src/lib/logger.ts
+++ b/packages/logger/src/lib/logger.ts
@@ -3,10 +3,10 @@ import { hashMessage } from 'ethers/lib/utils';
 import { version } from '@lit-protocol/constants';
 
 export enum LogLevel {
-  INFO = 0,
-  DEBUG = 1,
-  WARN = 2,
-  ERROR = 3,
+  ERROR = 0,
+  INFO = 1,
+  DEBUG = 2,
+  WARN = 3,
   FATAL = 4,
   TIMING_START = 5,
   TIMING_END = 6,
@@ -207,6 +207,7 @@ export class Logger {
   private _config: Record<string, any> | undefined;
   private _isParent: boolean;
   private _children: Map<string, Logger>;
+  private _timestamp: number;
 
   public static createLogger(
     category: string,
@@ -232,6 +233,7 @@ export class Logger {
     this._config = config;
     this._children = new Map();
     this._isParent = isParent;
+    this._timestamp = Date.now();
   }
 
   get id(): string {
@@ -240,6 +242,10 @@ export class Logger {
 
   get category(): string {
     return this._category;
+  }
+
+  get timestamp(): number {
+    return this._timestamp;
   }
 
   get Logs(): Log[] {
@@ -312,14 +318,15 @@ export class Logger {
     const arrayLog = log.toArray();
     if (this._config?.['condenseLogs'] && !this._checkHash(log)) {
       (this._level >= level || level === LogLevel.ERROR) &&
-        this._consoleHandler(...arrayLog);
+        this._consoleHandler && this._consoleHandler(...arrayLog);
       (this._level >= level || level === LogLevel.ERROR) &&
         this._handler &&
         this._handler(log);
+
       (this._level >= level || level === LogLevel.ERROR) && this._addLog(log);
     } else if (!this._config?.['condenseLogs']) {
       (this._level >= level || level === LogLevel.ERROR) &&
-        this._consoleHandler(...arrayLog);
+        this._consoleHandler && this._consoleHandler(...arrayLog);
       (this._level >= level || level === LogLevel.ERROR) &&
         this._handler &&
         this._handler(log);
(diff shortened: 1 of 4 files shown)


packages/logger/src/lib/logger.ts
```diff
@@ -3,14 +3,14 @@ import { hashMessage } from 'ethers/lib/utils';
...
 
 export enum LogLevel {
-  INFO = 0,
-  DEBUG = 1,
-  WARN = 2,
-  ERROR = 3,
+  OFF = -1,
+  ERROR = 0,
+  INFO = 1,
+  DEBUG = 2,
```
joshLong145, line 9: Note: These orderings are not standard. going to come back and re work the level enumeration weights in a fast follow PR with other changes we discussed.

Ansonhkg (approved):

packages/misc/src/lib/misc.ts
```diff
@@ -279,21 +276,16 @@ export const log = (...args: any): void => {
     return;
   }
 
-  if (globalThis?.litConfig?.debug !== true) {
```
MaximusHaximus, line 282: 🤔 Why does removing this check and the one on L310-L312 below not result in logging just being blanket enabled even when debug is set to false?

packages/misc/src/lib/misc.ts
```diff
@@ -226,8 +226,6 @@ declare global {
   var litConfig: any;
   var wasmExport: any;
   var wasmECDSA: any;
-  var logger: any;
-  var logManager: any;
```
MaximusHaximus, line 230: isn't logManager still a global with these changes?

packages/misc/src/lib/misc.ts
```diff
@@ -353,7 +339,7 @@ export const logErrorWithRequestId = (id: string, ...args: any) => {
     globalThis.logManager.get(globalThis.logger.category, id).error(...args);
 };
 
-export const logError = (...args: any) => {
```
MaximusHaximus, line 356: Why is it OK to leave the check on L370 below here, but we had to remove it in `logErrorWithRequestId` and `log`?

MaximusHaximus (changes requested):
hmmm -- The `logXXXX` methods from `misc` seem to be imported into a few more places that aren't passing a 'logger' as the first argument to them -- at first glance, I see contracts-sdk, lit-node-client-nodejs, crypto, encryption and EthWalletProvider, also `logError` in `pkp-base`. Do we need to update all of those places too?  Do we really need to add `logger` as the first arg to these functions to get the request IDs to print or is that an opportunistic change? 🤔 

Added a couple of questions in-line, but also, my apologies if I'm missing something obvious, but I'm having some difficulty understanding how this doesn't result in logging being fully enabled even when `debug` is disabled entirely -- what keeps us from logging all logs now that we don't exit early in misc methods?

packages/misc/src/lib/misc.ts
joshLong145: If you look in `core` where the logger is [created](https://github.com/LIT-Protocol/js-sdk/blob/feature/lit-3911-add-request-id-aggregation-to-logger-when-logging-is-diabled/packages/core/src/lib/lit-core.ts#L208) we now will set the logger to `OFF` which will prevent logs from outputting as it won't have a handler, nor will logs be stored. But it will allow a new logger child to be created which will aggregate request ids.

packages/logger/src/lib/logger.ts
```diff
@@ -427,14 +435,20 @@ export class LogManager {
...
     for (const category of this._loggers.entries()) {
       for (const child of category[1].Children) {
-        keys.push(child[0]);
+        keys.push([child[0], child[1].timestamp]);
       }
     }
 
-    return keys;
+    return keys
+      .sort((a: [string, number], b: [string, number]) => {
```
Ansonhkg, line 446: the goal here is to sort the array keys based on the timestamp (child[1].timestamp]) right? If so, I think we need to do the following instead, because the sorting callback should return a positive number, a negative number, and a 0 if they are equal.

```js
  return keys
    .sort((a: [string, number], b: [string, number]) => {
        return a[1] - b[1];
    }) // Corrected sorting
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

> A comparator conforming to the constraints above will always be able to return all of 1, 0, and -1, or consistently return 0. For example, if a comparator only returns 1 and 0, or only returns 0 and -1, it will not be able to sort reliably because anti-symmetry is broken. A comparator that always returns 0 will cause the array to not be changed at all, but is reliable nonetheless.

joshLong145: Ah! good catch, I had this refactored to the same but I did not push it. Since any value < 0 will result in: `[a,b]` or the original as the sort result this is rather elegant.

packages/logger/src/lib/logger.ts
FedericoAmura: Maybe use the levels from another lib like [ethers](https://docs.ethers.org/v5/api/utils/logger/#Logger-levels)?

FedericoAmura (approved):

Ansonhkg (approved):
LGTM!

PR #663, opened 2024-10-01
Feature/lit 3918 no globalthis variables anymore

FedericoAmura:
# Description

This PR removes the global variables used in several to share configs between modules or helper functions
Also removes the cdn installation instructions as those depend on a now gone global variable
The intent for this is to allow consumers of the SDK to have multiple versions of it without conflicts

# Changelog
- `globalThis.litWCProvider` from `auth-browser` package is gone
- added `setMiscLitConfig` function to pass litCore config up to `misc` package instead of relying on `globalThis.litConfig`

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


packages/misc/src/lib/misc.ts
```diff
@@ -1,3 +1,4 @@
+import { LitNodeClientConfig } from '@lit-protocol/types';
```
joshLong145, line 1: Is there a reason we did not also scope the `loggerManager` in this pr? I can handle this with my follow up but want to confirm.

FedericoAmura: Just wanted to avoid creating a big amount of conflicts with your changes. I can save it also in a variable in `misc` and import it into core returned from `bootstrapLogManager`. Or as you prefer, let me know 😃

joshLong145: I think this makes sense, Ill merge this into my branch where I am working on a couple follow ups

packages/misc/src/lib/misc.ts
```diff
@@ -171,7 +178,7 @@ export const logWithRequestId = (id: string, ...args: any) => {
   }
 
   // check if config is loaded yet
-  if (!globalThis?.litConfig) {
+  if (!litConfig) {
```
joshLong145, line 181: Making a note to reevaluate this module of scoping of the `LitConfig` instance for the logger refactor task: https://linear.app/litprotocol/issue/LIT-3921/migrate-to-interfacing-with-logger-through-module-over-global-object

joshLong145 (approved):

packages/auth-browser/src/lib/chains/eth.ts
```diff
@@ -386,8 +391,7 @@ export const connectWeb3 = async ({
     };
 
     if (isBrowser()) {
-      // @ts-ignore
-      globalThis.litWCProvider = wcProvider;
+      litWCProvider = wcProvider;
```
Ansonhkg, line 394: double checked that this variable is not used in other modules as it's now locally scoped instead of global

Ansonhkg (approved):
LGTM

PR #664, opened 2024-10-01
fix/v7: Remove preinstall script to avoid local build when installing as a dependency

Ansonhkg:
## 1. Remove `preinstall`

This PR removes the `preinstall` script from `package.json` in `wasm` package to prevent the unnecessary execution of the Rust build process (`yarn rust:build`) during package installation. Since the package is already built and published to npm, running the build step during installation is redundant and causes issues for users installing the package as a dependency.

Screenshot 2024-10-01 at 19 07 20

## 2. Fix postbuild script
Additionally, this PR also fixed & enhanced the `copyWasmBinary.mjs` postbuild script, where it removes the module import from the generated code since we encoded the entire binary. If no replacements were made, it will throw an error so that we know we probably need to fix this script.

image


joshLong145 (approved):
Nice :)

PR #658, opened 2024-09-26
chore: update package version to `6.7.0`

Ansonhkg:
# Changelog

All notable changes to this project will be documented in this file.

### 🚀 Features
- https://github.com/LIT-Protocol/js-sdk/pull/628
- https://github.com/LIT-Protocol/js-sdk/pull/651
- https://github.com/LIT-Protocol/js-sdk/pull/646

### 🐞 Bugs
- https://github.com/LIT-Protocol/js-sdk/pull/625
- https://github.com/LIT-Protocol/js-sdk/pull/649
- https://github.com/LIT-Protocol/js-sdk/pull/657

PR #639, opened 2024-09-16
LIT-3877 - Bundle wrapped keys actions - wrapped-keys-lit-actions

MaximusHaximus:
# Description

This PR drops the `@lit-protocol/wrapped-keys-bc` package that was hacky and confusing due to duplication and replaces it with a new package `@lit-protocol/wrapped-keys-lit-actions` that exports a lit action code repository used that can extend `@lit-protocol/wrapped-keys` like the following pictures

Extension can be with all the lit actions
Screenshot from 2024-09-18 18-36-31
Or per lit action (useful for bundling just the ones you want). This case will also guard against bundling things not supported such as unused actions or networks
Screenshot from 2024-09-18 17-01-02
Screenshot from 2024-09-18 17-01-19

When calling a WK LA extending with the code repository, the code string will be sent to the nodes
Screenshot from 2024-09-18 17-38-39
And when using WK without extension, the IPFS CID is still used
Screenshot from 2024-09-18 17-39-00

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Manual modification of local tests
- [x] Local demo app

# Checklist:

- [x] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #667, opened 2024-10-03
fix: `connection` is undefined when connecting to LitContracts

Ansonhkg:
# Description

Fix for this: (Unable to connect using the LitContracts client)
image

Due to this was added in the PR:
https://github.com/LIT-Protocol/js-sdk/pull/639/files#diff-2dcc2c94e7788b1bbaef5e6cf9158025a2580d238f807bd3cb923bbe355eb47bR401-R406

And we need this patch for the following affected versions:
- https://github.com/LIT-Protocol/js-sdk/pull/648
- https://github.com/LIT-Protocol/js-sdk/pull/658
- https://github.com/LIT-Protocol/js-sdk/pull/666

Anything <= 6.5.3 are working

### DIFF
--- a/packages/contracts-sdk/src/lib/contracts-sdk.ts
+++ b/packages/contracts-sdk/src/lib/contracts-sdk.ts
@@ -415,12 +415,12 @@ export class LitContracts {
     }
 
     this.log('Your Signer:', this.signer);
-    this.log('Your Provider:', this.provider.connection);
+    this.log('Your Provider:', this.provider?.connection!);
 
     if (!this.provider) {
       this.log('No provider found. Will try to use the one from the signer.');
       this.provider = this.signer.provider;
-      this.log('Your Provider(from signer):', this.provider.connection);
+      this.log('Your Provider(from signer):', this.provider?.connection!);
     }
 
     const addresses: any = await LitContracts.getContractAddresses(


FedericoAmura (approved):

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -415,12 +415,12 @@ export class LitContracts {
...
 
     this.log('Your Signer:', this.signer);
-    this.log('Your Provider:', this.provider.connection);
+    this.log('Your Provider:', this.provider?.connection!);
 
     if (!this.provider) {
       this.log('No provider found. Will try to use the one from the signer.');
       this.provider = this.signer.provider;
-      this.log('Your Provider(from signer):', this.provider.connection);
+      this.log('Your Provider(from signer):', this.provider?.connection!);
```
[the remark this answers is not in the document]
Ansonhkg: No - it will try to use the provider first, and if it's not there then it will use singer.provider

packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -415,12 +415,12 @@ export class LitContracts {
     }
 
     this.log('Your Signer:', this.signer);
-    this.log('Your Provider:', this.provider.connection);
+    this.log('Your Provider:', this.provider?.connection!);
```
joshLong145, line 418: should we move this log under the check for `this.provider` being defined.

Ansonhkg: You mean wrap it inside a if/else to check if 'connection' exists then print it out? If so, I don't think so, cus I'd like to see it connection is undefined in different environments

joshLong145: I see, if we want to log out when `this.provider` is undefined then we can keep  it.

joshLong145 (approved):

PR #674, opened 2024-10-04
feat: add session sigs formatter/reader

Ansonhkg:
# Description

Log formatted session sigs out to check expiration dates for the outer & inner auth sigs.

image


joshLong145:
❤️ This is nice

joshLong145 (approved):

PR #662, opened 2024-10-01
feat: replace bs58 dep with ethers

FedericoAmura:
# Description

This PR removes bs58 from our dependencies and uses ethers to base58 encoding/decoding where it is needed

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] "Bundle simplification"

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #675, opened 2024-10-08
LIT-3920 - wrapped-keys - Batch generate private keys & optionally sign messages in a single operation

MaximusHaximus:
# Description

This PR adds a new API function to `wrapped-keys` and corresponding LITAction to `wrapped-keys-lit-actions` facilitates the execution of multiple operations -- previously requiring executing *4* independent actions -- in a single API call to a single LIT action execution.

#### Noteworthy changes in this PR:
- Refactored existing LIT actions to allow composition of existing generation and message signing functionality into the new LIT action, while using the same code in the existing LIT actions that individually generate keys or sign messages with keys for a specific network (evm or solana)
- Added a new `LIT_ACTION_CID_REPOSITORY_COMMON` constant, and a new `litActionCodeRepositoryCommon` constant to `wrapped-keys`; 'common' LIT actions are not network-specific, so these repositories do not have multiple entries keyed by `Network` (evm or solana) under each LIT action entry; instead there is a single entry per LIT action which provides functionality for multiple networks.

##### `api.batchGeneratePrivateKeys()`
- Takes an array of `actions`, where each action can define either network (evm or solana), and an optional message to sign and return the signature of
##### `config.setLitActionsCodeCommon()`
- Allows injection of a LIT actions code repository containing 'common' LIT actions that are not defined per-network
##### `batchGenerateEncryptedKeys()`
- Facilitates generating `n` new keys _and_ optionally signing a message for each key, and returning the results of all generated keys and signed messages in a _single execution of one LIT action_

## Type of change

# How Has This Been Tested?

- [X] Ran all wrapped-keys localtests and added a new one to test the new batch generate/sign LIT action
- [X] Wrote and executed custom tests to verify that existing LIT actions continue to function identically to how they did before the internal refactors done in this PR

# Checklist:


local-tests/setup/tinny-environment.ts
```diff
@@ -460,6 +460,9 @@ export class TinnyEnvironment {
       await tx.wait();
     } catch (e) {
       throw new Error(`Failed to send funds to ${walletAddress}: ${e}`);
+    } finally {
+      // @ts-expect-error We don't have a user, but this works
+      this.releasePrivateKeyFromUser({ privateKey });
```
MaximusHaximus, line 465: Fixes the infinite hang during test run when you only have a couple of private keys in scope in your environment var

local-tests/test.ts
```diff
@@ -104,6 +104,12 @@ import { testFailImportWrappedKeysWithExpiredSessionSig } from './tests/wrapped-
 import { testExportWrappedKey } from './tests/wrapped-keys/testExportWrappedKey';
 import { testSignMessageWithSolanaEncryptedKey } from './tests/wrapped-keys/testSignMessageWithSolanaEncryptedKey';
 import { testSignTransactionWithSolanaEncryptedKey } from './tests/wrapped-keys/testSignTransactionWithSolanaEncryptedKey';
+import { testBatchGeneratePrivateKeys } from './tests/wrapped-keys/testBatchGeneratePrivateKeys';
+
+import { setLitActionsCodeToLocal } from './tests/wrapped-keys/util';
+
+// Use the current LIT action code to test against
+setLitActionsCodeToLocal();
```
MaximusHaximus, line 112: This makes our tinny tests use the locally built code for all LIT actions in wrapped-keys instead of running them against the CIDs in `wrapped-keys` LIT_ACTION_CID_REPOSITORY

local-tests/tests/wrapped-keys/testFailEthereumSignTransactionWrappedKeyWithInvalidParam.ts
```diff
@@ -70,11 +70,7 @@ export const testFailEthereumSignTransactionWrappedKeyWithInvalidParam = async (
         id,
       });
     } catch (e: any) {
-      if (
-        e.message.includes(
-          'Error executing the Signing Lit Action: Error: When signing transaction- invalid hexlify value'
-        )
-      ) {
+      if (e.message.includes('invalid hexlify value')) {
```
MaximusHaximus, line 73: Made this less specific since the exact string isn't important and this is now being properly prefixed consistently due to refactors in error handling in the LA in this PR

local-tests/tests/wrapped-keys/testFailEthereumSignTransactionWrappedKeyWithMissingParam.ts
```diff
@@ -66,11 +66,7 @@ export const testFailEthereumSignTransactionWrappedKeyWithMissingParam = async (
         id,
       });
     } catch (e: any) {
-      if (
-        e.message.includes(
-          'Error executing the Signing Lit Action: Error: Missing required field: toAddress'
-        )
-      ) {
+      if (e.message.includes('Missing required field: toAddress')) {
```
MaximusHaximus, line 69: Made this less specific since the exact string isn't important and this is now being properly prefixed consistently due to refactors in error handling in the LA in this PR

local-tests/tests/wrapped-keys/testSignTransactionWithSolanaEncryptedKey.ts
```diff
@@ -125,7 +126,7 @@ export const testSignTransactionWithSolanaEncryptedKey = async (
     // const confirmation = await solanaConnection.confirmTransaction(signedTx);
     // console.log(confirmation); // { context: { slot: 321490379 }, value: { err: null } }
 
-    const signatureBuffer = Buffer.from(signedTx, 'base64');
+    const signatureBuffer = Buffer.from(ethers.utils.base58.decode(signedTx));
```
MaximusHaximus, line 129: This test was failing if using the LIT action code merged since #652, but passed because we were using the CID to run the tests.  Now that we run the tests against the current, in-repo LIT actions, it needed to be updated to match the changes made in #652.

local-tests/tests/wrapped-keys/util.ts
```diff
@@ -1,7 +1,50 @@
...
+  LitActionCodeRepository,
+  LitActionCodeRepositoryCommon,
+  EthereumLitTransaction,
+} from '@lit-protocol/wrapped-keys';
+
+const emptyLitActionRepositoryCommon: LitActionCodeRepositoryCommon = {
+  batchGenerateEncryptedKeys: '',
+};
+
+const emptyLitActionRepository: LitActionCodeRepository = {
```
MaximusHaximus, line 20: Sadly we can't run all of the tests twice (once w/ cids and once w/ actions) because this is a global mutation and it would require runInBand=true to avoid inconsistent behaviour.  When we migrate to Jest, I'll make that happen (since Jest sandboxes suites)

packages/wrapped-keys-lit-actions/src/index.ts
```diff
@@ -1,12 +1,16 @@
+import * as batchGenerateEncryptedKeys from './generated/common/batchGenerateEncryptedKeys';
 import * as exportPrivateKey from './generated/common/exportPrivateKey';
 import * as generateEncryptedEthereumPrivateKey from './generated/ethereum/generateEncryptedEthereumPrivateKey';
-import * as signMessageWithEthereumEncryptedKey from './generated/ethereum/signMessageWithEthereumEncryptedKey';
-import * as signTransactionWithEthereumEncryptedKey from './generated/ethereum/signTransactionWithEthereumEncryptedKey';
+import * as signMessageWithEthereumEncryptedKey from './generated/ethereum/signMessageWithEncryptedEthereumKey';
```
MaximusHaximus, line 4: Renamed the files for least-to-most-specific naming convention. Exports remain named the same until we ship a breaking change, then we'll make those consistent with the filenames.

packages/wrapped-keys-lit-actions/src/lib/common/exportPrivateKey.js
```diff
@@ -13,31 +16,21 @@ const { removeSaltFromDecryptedKey } = require('../utils');
...
+      // Silently exit on nodes which didn't run the `decryptToSingleNode` code
+      return;
+    }
 
-  try {
     const privateKey = removeSaltFromDecryptedKey(decryptedPrivateKey);
     Lit.Actions.setResponse({ response: privateKey });
   } catch (err) {
-    Lit.Actions.setResponse({ response: err.message });
+    Lit.Actions.setResponse({ response: `Error: ${err.message}` });
```
MaximusHaximus, line 34: New structure ensures all errors thrown from the LA result in `Error:` prefix in the response.

packages/wrapped-keys-lit-actions/src/lib/ethereum/internal/signTransaction.js
```diff
@@ -0,0 +1,141 @@
...
+
+    if (!unsignedTransaction.value) {
+      throw new Error('Missing required field: value');
+    }
+
+    if (!unsignedTransaction.chainId) {
+      throw new Error('Missing required field: chainId');
+    }
+
+    return {
```
MaximusHaximus, line 21: Moved hexlify call to this so that we get errors about invalid parsing before running the rest of the LA code like fetching nonces etc.

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const actionResults = await batchGenerateKeysWithLitAction({
+    ...params,
+    litActionIpfsCid: litActionCode ? undefined : litActionIpfsCid,
+    litActionCode: litActionCode ? litActionCode : undefined,
+    accessControlConditions: [allowPkpAddressToDecrypt],
+    pkpSessionSigs,
+  });
+
+  return Promise.all(
```
MaximusHaximus, line 44: If any of these calls fails, the entire execution will exit even if the LA was successful and we only need to re-try storing the results using `storePrivateKey()`. A retry on network errors would be good to add at some point since the alternative is that the consumer re-runs the entire process from the top again, even though we might have valid signatures and ciphertexts in scope.

packages/wrapped-keys/src/lib/api/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { NETWORK_EVM, NETWORK_SOLANA } from '../constants';
+import { KeyType, Network } from '../types';
+
+export function getKeyTypeFromNetwork(network: Network): KeyType {
```
MaximusHaximus, line 4: Extracted this to be re-used in both generate codepaths

packages/wrapped-keys/src/lib/lit-actions-client/types.ts
```diff
@@ -36,3 +42,8 @@ export type LitActionCodeRepository = Readonly<
 export type LitActionCodeRepositoryInput = Partial<
   Record<LitActionType, LitActionCodeRepositoryEntryInput>
 >;
+
+export type LitActionCodeRepositoryCommon = Record<LitActionTypeCommon, string>;
```
MaximusHaximus, line 46: New common repository just has string per key, no { evm, solana } objects since common action(s) work on multiple networks.

packages/wrapped-keys/src/lib/types.ts
```diff
@@ -134,6 +134,32 @@ export interface ExportPrivateKeyResult {
...
+export type GeneratePrivateKeyAction = ApiParamsSupportedNetworks & {
+  generateKeyParams: { memo: string };
+  signMessageParams?: { messageToSign: string | Uint8Array };
+};
+
+/** @typedef BatchGeneratePrivateKeysParams
+ * @extends BaseApiParams
+ */
+export type BatchGeneratePrivateKeysParams = BaseApiParams & {
+  actions: GeneratePrivateKeyAction[];
```
MaximusHaximus, line 149: Can generate `n` of any combination of solana or evm keys.

local-tests/test.ts
DashKash54: Since the idea is that users use the IPFS id shouldn't the tests reflect that?

local-tests/tests/wrapped-keys/testSignTransactionWithSolanaEncryptedKey.ts
DashKash54: Shouldn't we use IPFS Cid in the tests too? https://github.com/LIT-Protocol/js-sdk/pull/675#discussion_r1792648904

local-tests/tests/wrapped-keys/util.ts
DashKash54: Run them sequentially?

packages/wrapped-keys-lit-actions/src/lib/common/exportPrivateKey.js
DashKash54: Since these are Lit Action errors should we prefix the error msg with "Lit Action Error" or the `err.message` displays that already?

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
DashKash54: Cool we plan to update this later with a retry and NOT now?

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
DashKash54, line 1: A user can chose to only generate multiple keys without being necessitated to sign a message with them?

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
```diff
@@ -0,0 +1,159 @@
...
+} = require('../solana/internal/generatePrivateKey');
+const { signMessageSolanaKey } = require('../solana/internal/signMessage');
+
+/* global accessControlConditions, actions, Lit*/
+
+async function processEthereumAction(action, ndx) {
+  const { network, generateKeyParams } = action;
+  const messageToSign = action.signMessageParams?.messageToSign;
+
+  const ethereumKey = generateEthereumPrivateKey();
```
DashKash54, line 17: Private key should be generated inside only a single node as this will create a different privateKey on each node! @MaximusHaximus

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
```diff
@@ -0,0 +1,159 @@
...
+
+  const [generatedPrivateKey, messageSignature] = await Promise.all([
+    Lit.Actions.runOnce(
+      { waitForResponse: true, name: `encryptEthereumPrivateKey_${ndx}` },
+      async () =>
+        JSON.stringify(
+          await encryptPrivateKey({
+            accessControlConditions,
+            publicKey: ethereumKey.publicKey,
```
DashKash54, line 26: Why is the pubkey required in the encrypt function since we only encrypt the private key? If we wanna return it in the same object we can just add it separately in the stringify?

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
```diff
@@ -0,0 +1,159 @@
...
+            privateKey: ethereumKey.privateKey,
+          })
+        )
+    ),
+    messageToSign
+      ? signMessageEthereumKey({
```
DashKash54, line 32: Sign message is not running inside the same node the key is generated?

packages/wrapped-keys-lit-actions/src/lib/common/internal/getDecryptedKey.js
```diff
@@ -0,0 +1,20 @@
+/* global Lit */
+
+export async function getDecryptedKey({
+  accessControlConditions,
+  ciphertext,
+  dataToEncryptHash,
+}) {
+  try {
+    // May be undefined, since we're using `decryptToSingleNode`
```
DashKash54, line 9: This should always return the decrypted data to the selected node, what is meant by this comment?

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
```diff
@@ -0,0 +1,159 @@
...
+          messageToSign: messageToSign,
+          privateKey: ethereumKey.privateKey,
```
DashKash54, line 34: `ethereumKey` will be different for each node so the `messageSignature` won't match the returned `generatedPrivateKey`

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
```diff
@@ -0,0 +1,159 @@
...
+      : {}),
+  };
+}
+
+async function processSolanaAction(action, ndx) {
+  const { network, generateKeyParams } = action;
+
+  const messageToSign = action.signMessageParams?.messageToSign;
+
+  const solanaKey = generateSolanaPrivateKey();
```
DashKash54, line 56: Same as Ethereum, different privateKey on each node

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (
+      results[0].generatedPrivateKey.memo !== 'Test evm key' ||
+      results[1].generatedPrivateKey.memo !== 'Test solana key'
+    ) {
+      throw new Error(
+        'Results not in order sent; expected evm as first result, solana as second'
+      );
+    }
+
+    if (
```
DashKash54, line 59: Lets' add assertions to check whether the messages are signed by the `generatedPrivateKey`(s) @MaximusHaximus

packages/wrapped-keys-lit-actions/src/lib/ethereum/internal/signMessage.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+export async function signMessageEthereumKey({ privateKey, messageToSign }) {
+  const { signature, walletAddress } = await signMessage({
+    privateKey,
+    messageToSign,
+  });
+
+  const recoveredAddress = verifyMessageSignature({ messageToSign, signature });
```
DashKash54, line 30: **LATER:** Ideally this shouldn't be required and this adds an additional overhead as well, not sure by how much though. So it'll be great if we can add timings to see this

DashKash54 (changes requested):
@MaximusHaximus there appears to be a problem in the batch method as each node generates its own random private key and signs with it. But only one node returns the encrypted key to the user hence the encrypted key won't match the signedMessage

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
MaximusHaximus: :+1: yes, just making a note about the failure mode

packages/wrapped-keys-lit-actions/src/lib/common/exportPrivateKey.js
MaximusHaximus: In order to get the error, it has to be parsed from a node response -- so the SDK throws a specific type of error indicating that it was an error running a request against a LIT node and the stack trace from the SDK indicates it comes from handling a node response

local-tests/tests/wrapped-keys/util.ts
MaximusHaximus: We'd have to configure the entire suite to always run _all_ tests to `runInBand` the way `local-tests` is currently designed; since we're already in the midst of migrating to Jest and it will be able to run them in parallel it didn't seem like the effort was justified to enhance the existing local-tests code to allow subsets of tests to run in band

local-tests/test.ts
MaximusHaximus: I guess it depends on the users; some people will be using the code directly, others the CIDs... what I want to do is run the entire suite against the cids and then again against the current code ideally, but see other comment re: doing so without jest.

The reason this is here right now is because we have a chicken and egg problem - in-progress code in PRs isn't in IPFS, and pinning repeatedly to IPFS as we iterate on a lit action would result in a ton of churn and manual updates right now.

We could make this code always use IPFS CIDs, but it would require that all actions we want to test would exist in IPFS before we can test against them -- and we want to run tests while developing the actions where they will be changing often.  

In this PR, I wanted to be sure tests pass with the _current code_ rather than previous CIDs -- once we have confirmed all of the LIT actions code here is GTG and published all of the lit actions to IPFS, I can update the CIDs in `wrapped-keys` and remove this code so the tests run against CIDs again

local-tests/tests/wrapped-keys/testSignTransactionWithSolanaEncryptedKey.ts
MaximusHaximus: Once all the actions have been reviewed and we've published/pinned them to IPFS, we can remove the injection of the LIT action code here.  Right now it's using injected code so that we can be sure this commit is sane and functional even though it's not in IPFS yet (see other comment)

local-tests/setup/tinny-environment.ts
```diff
@@ -447,8 +446,9 @@ export class TinnyEnvironment {
    * @throws If there is an error sending the funds.
    */
   getFunds = async (walletAddress: string, amount = '0.001') => {
+    const privateKey = await this.getAvailablePrivateKey();
```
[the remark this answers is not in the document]
MaximusHaximus: I needed the privateKey in scope in the finally{} block, so if I move it inside the try{}, it won't be in scope in the finally{} due to block scoping rules.  also, if this call throws an error, I don't want the finally() to run that tries to release the private key -- since we won't know what private key to release if this call throws an error

local-tests/tests/wrapped-keys/util.ts
```diff
@@ -1,7 +1,50 @@
...
+  litActionRepository,
+} from '@lit-protocol/wrapped-keys-lit-actions';
+
+import type {
+  LitActionCodeRepository,
+  LitActionCodeRepositoryCommon,
+  EthereumLitTransaction,
+} from '@lit-protocol/wrapped-keys';
+
+const emptyLitActionRepositoryCommon: LitActionCodeRepositoryCommon = {
```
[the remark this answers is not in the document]
MaximusHaximus: The way the fallback logic inside of the `wrapped-keys` package works, if the code for a particular LIT action is an empty string (falsey), it uses CIDs -- otherwise it uses the string as the LIT action code. It's 'empty' because it contains no code at all, for any key -- basically it resets the entire code repository so that any calls to use the LIT actions will use the defined CIDs instead of trying to use code.  For type safety it needs to be an object with the correct keys, where all appropriate values remain strings.

local-tests/tests/wrapped-keys/util.ts
```diff
@@ -1,7 +1,50 @@
...
+  },
+  signMessage: {
+    evm: '',
+    solana: '',
+  },
+  generateEncryptedKey: {
+    evm: '',
+    solana: '',
+  },
+  exportPrivateKey: {
```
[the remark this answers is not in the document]
MaximusHaximus: You're right -- I wanted to move this into the common repository, but didn't want to ship a breaking change here which would mean shipping js-sdk v7 since we don't have independent package versioning yet :(

packages/wrapped-keys-lit-actions/src/index.ts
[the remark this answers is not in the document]
MaximusHaximus: Am keen to do this, but am trying to ship this without any breaking changes, and the repository shape is an external API -- see https://github.com/LIT-Protocol/js-sdk/pull/675#discussion_r1792687675

packages/wrapped-keys-lit-actions/src/lib/common/internal/getDecryptedKey.js
MaximusHaximus: This was the existing logic, which i retained here.  My understanding is that because this code runs on all of the nodes, but the `decryptToSingleNode` call only actually returns a value on  _the one node that is chosen by backend consensus_, all of the the nodes that are executing this action that _aren't_ chosen to be 'the single node' that runs the decrypt will return undefined for this call.  As a result, we can't continue executing any further logic on that node since we don't have the privateKey in scope -- so  we exit immediately, and that node returns an empty result.  The SDK looks at the returned results from all nodes, and uses the only _non-empty result_ from the responses from all of the nodes.

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
```diff
@@ -0,0 +1,73 @@
...
+            keyType: getKeyTypeFromNetwork(network),
+            pkpAddress,
+          },
+          litNetwork: litNodeClient.config.litNetwork,
+        });
+
+        return {
+          ...(signature ? { signedMessage: { signature } } : {}),
+          generatedPrivateKey: {
+            memo: generatedPrivateKey.memo,
```
DashKash54, line 64: Isn't this provided in the params passed by the user?

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
```diff
@@ -0,0 +1,73 @@
...
+            id: id,
```
DashKash54, line 65: nit
```suggestion
            id,
```

DashKash54 (changes requested):
Few more comments

packages/wrapped-keys-lit-actions/src/lib/solana/generateEncryptedSolanaPrivateKey.js
```diff
@@ -1,40 +1,32 @@
...
-
-  const resp = await Lit.Actions.runOnce(
-    { waitForResponse: true, name: 'encryptedPrivateKey' },
-    async () => {
-      const solanaKeypair = Keypair.generate();
-      const privateKey =
-        LIT_PREFIX + Buffer.from(solanaKeypair.secretKey).toString('hex');
-      let utf8Encode = new TextEncoder();
-      const to_encrypt = utf8Encode.encode(privateKey);
+  const { privateKey, publicKey } = generateSolanaPrivateKey();
```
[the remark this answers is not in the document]
MaximusHaximus: :+1: good catch. This was working -- perhaps coincidentally due to internal incidental ordering of responses.
- I modified the LA to use `runOnce()` on the `processActions` call
- I added test assertions that verify the signatures -- but, see above -- they were _passing with the broken implementation anyway :(_

packages/wrapped-keys-lit-actions/src/lib/ethereum/generateEncryptedEthereumPrivateKey.js
```diff
@@ -1,37 +1,33 @@
...
-  const LIT_PREFIX = 'lit_';
-
-  const resp = await Lit.Actions.runOnce(
-    { waitForResponse: true, name: 'encryptedPrivateKey' },
-    async () => {
-      const wallet = ethers.Wallet.createRandom();
-      const privateKey = LIT_PREFIX + wallet.privateKey.toString();
-      let utf8Encode = new TextEncoder();
-      const to_encrypt = utf8Encode.encode(privateKey);
+  const { privateKey, publicKey } = generateEthereumPrivateKey();
```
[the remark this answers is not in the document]
MaximusHaximus: Should be GTG now - see https://github.com/LIT-Protocol/js-sdk/pull/675#discussion_r1792721090

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
MaximusHaximus: Yup -- but it's also the only identifying information that the consumer can get back from the LA that they can use to verify what they are getting back is what they sent, because they don't get the message back, and they don't know what the generated key would be -- so I added it as a convenience so that they could see in the result which entry was for which (easier debugging)

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
MaximusHaximus: :+1: changed

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
```diff
@@ -0,0 +1,73 @@
...
+  params: BatchGeneratePrivateKeysParams
+): Promise<BatchGeneratePrivateKeysResult> {
+  const { pkpSessionSigs, litNodeClient } = params;
+
+  const sessionSig = getFirstSessionSig(pkpSessionSigs);
+  const pkpAddress = getPkpAddressFromSessionSig(sessionSig);
+
+  const allowPkpAddressToDecrypt = getPkpAccessControlCondition(pkpAddress);
+
+  const { litActionCode, litActionIpfsCid } = getLitActionCodeOrCidCommon(
```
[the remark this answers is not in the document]
MaximusHaximus: Actually, this is named accurately -- one of these two values will _always_ be falsy -- if the code exists, it will be returned and `litActionIpfsCid` will be undefined, and if the code doesn't exist then `litActionIpfsCid` will be set but the code will be falsy.

packages/wrapped-keys-lit-actions/src/lib/ethereum/internal/signMessage.js
MaximusHaximus: I'm happy to remove it if it's superfluous -- I only copied it in because it was in the existing implementation :+1:  Is there any case where this would fail? Usually verification of the signature is the responsibility of the receiver of the signature rather than the creator of it and it does appear that this should always work unless ethers is broken internally for some reason (at which point the consumer's verify should fail anyway?)

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
MaximusHaximus: I added these assertions, but per https://github.com/LIT-Protocol/js-sdk/pull/675#discussion_r1792721090 - they passed even with the existing implementation -- perhaps coincidentally.

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
MaximusHaximus: Yes, they can omit the `messageToSign` (it's optional) and just get an array of generated keys back with no message signatures :+1: :)

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
MaximusHaximus: Because getting the data out of the runOnce requires that we `stringify()` the result (or it comes back as [Object Object]), injecting publicKey back into the result of this function required that I `JSON.parse()` the result, then spread the parsed resulting object back into a new object and inject publicKey into it, only to then JSON.stringify() it back again on the very next line -- very inefficient and tricky to reason about.

This way I get to avoid adding another JSON.parse() to the calling code in the existing network-specific actions :+1: :)

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
MaximusHaximus: :+1: https://github.com/LIT-Protocol/js-sdk/pull/675#discussion_r1792721090

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
MaximusHaximus: :+1: https://github.com/LIT-Protocol/js-sdk/pull/675#discussion_r1792721090

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
MaximusHaximus: https://github.com/LIT-Protocol/js-sdk/pull/675#discussion_r1792721090 :+1:

MaximusHaximus:
@DashKash54 Per our conversation, I've pushed commits that remove all `Lit.Actions.runOnce()` usage in the new generate function, and moved the generation of new private keys _inside_ of the runOnce blocks for existing keys.

Thinking about this change, though -- we could also normalize this behaviour and remove the runOnce usage for the _existing_ LIT actions, and configure them to use the same new 'only use response from a single node' logic.  This would improve performance for anyone using the per-network generate methods also.

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
```diff
@@ -0,0 +1,75 @@
...
+import { TinnyEnvironment } from 'local-tests/setup/tinny-environment';
+import { api } from '@lit-protocol/wrapped-keys';
+import { getPkpSessionSigs } from 'local-tests/setup/session-sigs/get-pkp-session-sigs';
+
+const { batchGeneratePrivateKeys } = api;
+
+/**
+ * Test Commands:
+ * ✅ NETWORK=cayenne yarn test:local --filter=testSignMessageWithSolanaEncryptedKey
+ * ✅ NETWORK=manzano yarn test:local --filter=testSignMessageWithSolanaEncryptedKey
```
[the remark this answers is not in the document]
FedericoAmura: +1 because v7 does not support those networks anymore. I would prefer we document just those so it is consistent after merge and v7 release

packages/wrapped-keys-lit-actions/src/lib/common/internal/getDecryptedKey.js
```diff
@@ -0,0 +1,20 @@
+/* global Lit */
+
+export async function getDecryptedKey({
```
FedericoAmura, line 3: nit: can this be renamed to `getDecryptKeyToSingleNode`. Makes it easier to understand from a caller perspective that it can not be chosen and get `undefined`

packages/wrapped-keys-lit-actions/src/lib/solana/internal/signMessage.js
```diff
@@ -0,0 +1,53 @@
+import { Keypair } from '@solana/web3.js';
+import bs58 from 'bs58';
```
FedericoAmura, line 2: We can use ethers which is already in the nodes deno env instead of this

FedericoAmura, line 2: We can use ethers which is already in the nodes deno env

FedericoAmura: What's more, we are dropping `bs58` from the SDK in v7 #662

packages/wrapped-keys/src/lib/lit-actions-client/batch-generate-keys.ts
```diff
@@ -0,0 +1,55 @@
...
+    sessionSigs: pkpSessionSigs,
+    ipfsId: litActionIpfsCid,
+    code: litActionCode,
+    jsParams: {
+      actions,
+      accessControlConditions,
+    },
+    ipfsOptions: {
+      overwriteCode:
+        GLOBAL_OVERWRITE_IPFS_CODE_BY_NETWORK[litNodeClient.config.litNetwork],
```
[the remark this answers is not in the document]
Ansonhkg: it's an optional overwrite inside executeJs, this is a global overwrite by network set in the constants file. it was introduced when we were waiting for the datil-prod upgrade. Not sure if we still want this now, but it's an option for users to fetch from client-side instead of node-side

local-tests/test.ts
MaximusHaximus: :+1: I agree @FedericoAmura <3 Ideally, all tests would run against the data in the same commit by _default_, and if we want to test against prior versions we'd explicitly do _that_.  The main thing is just getting the CID repo in the same commit to always match the current source code in that commit... I think we can do that with a husky hook script that runs on-commit

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
```diff
@@ -0,0 +1,155 @@
...
+    const pkpSessionSigsSigning = await getPkpSessionSigs(
+      devEnv,
+      alice,
+      null,
+      new Date(Date.now() + 1000 * 60 * 10).toISOString()
+    ); // 10 mins expiry
+
+    const solanaMessageToSign = 'This is a test solana message';
+    const evmMessageToSign = 'This is a test evm message';
+    const results = await batchGeneratePrivateKeys({
```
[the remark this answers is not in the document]
MaximusHaximus: Caught it while you were reviewing ❇️  :)

packages/wrapped-keys/src/lib/lit-actions-client/constants.ts
```diff
@@ -19,4 +19,8 @@ const LIT_ACTION_CID_REPOSITORY: LitCidRepository = Object.freeze({
   }),
 });
 
-export { LIT_ACTION_CID_REPOSITORY };
+const LIT_ACTION_CID_REPOSITORY_COMMON: LitCidRepositoryCommon = Object.freeze({
+  batchGenerateEncryptedKeys: '',
```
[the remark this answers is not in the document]
MaximusHaximus: Yup -- as soon as we've got :+1: on all the LIT action changes in this PR, I will run the new `sync-actions-to-ipfs` script, update this entry with the CIDs we've synced, and add a commit here that sets them

packages/wrapped-keys-lit-actions/src/lib/common/internal/getDecryptedKey.js
MaximusHaximus: Good idea :+1: <3

packages/wrapped-keys/src/lib/lit-actions-client/constants.ts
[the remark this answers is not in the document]
MaximusHaximus: I just pushed a commit that updates these cids to be IPFS CIDs for the LA code in this PR <3 :)

MaximusHaximus:
@DashKash54 I merged in the new executeJs change from #677 and [updated `batchGenerateKeysWithLitAction()`](https://github.com/LIT-Protocol/js-sdk/pull/675/commits/c3102c67cd02b1d4b2fc671a205a22244cbc64c4) to set it.

This behaviour leverages the existing executeJs function behaviour to get the absolute fastest response from a node and return immediately as soon as 1 successful response is received.  We have the added benefit that if _any one request to any node_ succeeds, the call will succeed.  Basically the best of both worlds when it comes to both speed of response _and_ resilience to errors.

packages/types/src/lib/interfaces.ts
```diff
@@ -522,14 +523,29 @@ export interface JsonExecutionSdkParams
...
+
+  /**
+   * number of responses required to consider the execution successful
+   */
+  numResponsesRequired?: number;
+
+  /**
+   * idea: the number of nodes to pay for running executions
+   */
+  // numNodesToRunOn?: number;
```
FedericoAmura, line 548: Great idea. Even more important on Naga

FedericoAmura (approved):

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
```diff
@@ -0,0 +1,147 @@
...
+          privateKey: ethereumKey.privateKey,
+        })
+      : Promise.resolve(),
+  ]);
+
+  return {
+    network,
+    generatedPrivateKey: {
+      ...generatedPrivateKey,
+      memo: generateKeyParams.memo,
```
DashKash54, line 37: This is good to return since we're generating multiple keys

DashKash54 (commented):
Left a few more comments

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
```diff
@@ -0,0 +1,154 @@
...
+  const { decryptedPrivateKey } = await exportPrivateKey({
+    litNodeClient,
+    network: 'evm',
+    id: evmResult.generatedPrivateKey.id,
+    pkpSessionSigs,
+  });
+
+  const recoveredAddress = verifyMessageSignature();
+
+  const wallet = new ethers.Wallet(decryptedPrivateKey);
```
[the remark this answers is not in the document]
MaximusHaximus: good call -- I actually took this assertion pattern from the original `testEthereumBroadcastTransactionGeneratedKey` test that you wrote, to keep things consistent with our existing tests. Should we go ahead and modify them both to avoid exporting the privateKey for either one

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
```diff
@@ -0,0 +1,154 @@
...
+
+const { batchGeneratePrivateKeys, exportPrivateKey } = api;
+
+async function verifySolanaSignature(
+  solanaResult: BatchGeneratePrivateKeysActionResult,
+  solanaMessageToSign
+) {
+  const {
+    signedMessage: { signature },
+    generatedPrivateKey: { generatedPublicKey },
```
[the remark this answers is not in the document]
MaximusHaximus: Sure :+1: changed this

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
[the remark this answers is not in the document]
MaximusHaximus: :+1: updated this.

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    console.log('solana verify sig');
+    await verifySolanaSignature(results[1], solanaMessageToSign);
+
+    console.log('evm verify sig');
+    await verifyEvmSignature(
+      devEnv.litNodeClient,
+      results[0],
+      evmMessageToSign,
+      pkpSessionSigsSigning
```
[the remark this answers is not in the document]
MaximusHaximus: All variables being passed into these functions are used -- if we stop using them, I'll remove them from the signature :)

verifyMessage requires { evmMessageToSign, evmResult.signMessage.signature }
export private key requires { litNodeClient, pkpSessionSigs }

local-tests/tests/wrapped-keys/util.ts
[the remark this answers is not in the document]
MaximusHaximus: Keeping the values on the object always as strings keeps the type consistent so handling values in the object requires less complexity in typescript -- because values on the object are all of type `string`, we always know what type a value is without needing to do assertions against it, and the code to check if it's a falsy string to know if we should try to use any particular value as code or not is the same as if the value was 'null'  :)

packages/wrapped-keys/src/lib/lit-actions-client/batch-generate-keys.ts
```diff
@@ -0,0 +1,56 @@
...
+    accessControlConditions,
+    litNodeClient,
+    actions,
+    pkpSessionSigs,
+    litActionIpfsCid,
+    litActionCode,
+  } = args;
+
+  const result = await litNodeClient.executeJs({
+    numResponsesRequired: 1,
```
DashKash54, line 40: This sends to only one node or sends to all but waits on only one? @MaximusHaximus

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
MaximusHaximus: :+1:
In theory the results coming back in the same _order_ that the array came in is 'enough', but being able to see the memo in the element you've 'zipped' together with an array you already had in your client context is pretty useful for debugging

packages/wrapped-keys/src/lib/lit-actions-client/constants.ts
DashKash54, line 1: This is the updated ones with consolidation and single node request?

DashKash54 (approved):
Looks good, approving but left few more comments & questions @MaximusHaximus

DashKash54 (approved):
Just a small suggestion for big PRs like this it'll be nice to keep the name same and raise another PR for updating the naming, makes easier to review @MaximusHaximus

packages/types/src/lib/interfaces.ts
```diff
@@ -687,7 +703,7 @@ export interface SigShare {
   bigr?: string; // backward compatibility
   bigR?: string;
   publicKey: string;
-  dataSigned?: string;
+  dataSigned?: string | 'fail';
```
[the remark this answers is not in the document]
MaximusHaximus: Not sure about exactly why this case might happen -- but this was added in https://github.com/LIT-Protocol/js-sdk/pull/677/files

local-tests/test.ts
DashKash54: > husky hook script that runs on-commit

Good idea

packages/wrapped-keys-lit-actions/src/lib/common/batchGenerateEncryptedKeys.js
DashKash54: No longer valid, we can change this later to move `pubKey` outside the `encryptPrivateKey`

packages/wrapped-keys-lit-actions/src/lib/solana/signTransactionWithEncryptedSolanaKey.js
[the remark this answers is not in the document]
MaximusHaximus: No big deal -- may as well fix it while we're here and thinking about it, and it's is a trivial change, I went ahead and made it :+1 :)

DashKash54:
> we could also normalize this behaviour and remove the runOnce usage for the existing LIT actions
We can update the exiting Lit Actions later as it's not only the Lit Action but also the client that needs to be updated with run on single node flag eg: `generatePrivateKey()`

packages/wrapped-keys-lit-actions/src/lib/solana/internal/signMessage.js
DashKash54: As long as it's the same functionality and returns the Solana message in base58 it's fine @MaximusHaximus

DashKash54:
>This behaviour leverages the existing executeJs function behaviour to get the absolute fastest response from any node and return immediately, as soon as any one successful response is received. 

It has a down side that all the node loads increase, we can reduce the network load by 8-10x if we run only on once
We haven't experience Lit Action throwing random errors like this so we should be confident to make this change @MaximusHaximus @Ansonhkg

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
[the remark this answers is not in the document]
DashKash54: I think we could make this change as it's a minor one? @MaximusHaximus

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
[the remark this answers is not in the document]
DashKash54: This is what I meant by this comment: https://github.com/LIT-Protocol/js-sdk/pull/675#discussion_r1794310145 @MaximusHaximus

packages/wrapped-keys/src/lib/lit-actions-client/batch-generate-keys.ts
MaximusHaximus: Sends to all, but returns the _first successful response_.

This way...
1. We aren't unlucky and waiting on a node that we chose randomly that is slow
2. If _any_ nodes return a successful response, the request succeeds, instead of the request failing if the one random node we chose throws an error.

packages/wrapped-keys/src/lib/lit-actions-client/constants.ts
MaximusHaximus: Yes, but I will update them again once we have the final :+1: on the current LA code

packages/wrapped-keys-lit-actions/src/lib/ethereum/signTransactionWithEncryptedEthereumKey.js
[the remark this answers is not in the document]
MaximusHaximus: No big deal -- may as well fix it while we're here and thinking about it, and it's is a trivial change, I went ahead and made it 👍 :)

packages/wrapped-keys-lit-actions/src/lib/solana/signTransactionWithEncryptedSolanaKey.js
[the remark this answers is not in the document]
DashKash54: We cna do a similar change to Etheruem as well, validate first decrypt later @MaximusHaximus

local-tests/test.ts
DashKash54, line 1: Why comment it out?

MaximusHaximus:
> > we could also normalize this behaviour and remove the runOnce usage for the existing LIT actions
> 
> We can update the existing Lit Actions **later** as it's not only the Lit Action but also the client that needs to be updated with run on single node flag eg: `generatePrivateKey()`

OK, I'll leave them as is for now -- just FYI, the client change only involves adding a single parameter into each API function's file where they call `executeJs` -- not a big deal.

DashKash54 (commented):
Please address the remaining comments specifically the ones in which I've tagged you @MaximusHaximus

packages/types/src/lib/interfaces.ts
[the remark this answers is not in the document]
Ansonhkg: `fail` is a string returned from the node if `Lit.Action.signEcdsa` failed to sign AND if the sigType is `K256`

DashKash54:
> OK, I'll leave them as is for now -- just FYI, the client change only involves adding a single parameter into each API function's file where they call executeJs -- not a big deal.

If you want feel free to do it now @MaximusHaximus

local-tests/test.ts
MaximusHaximus: Was just fixing that -- was a mistake :)

MaximusHaximus: All set

packages/wrapped-keys/src/lib/lit-actions-client/constants.ts
DashKash54: Yeah those look good

packages/wrapped-keys/src/lib/lit-actions-client/batch-generate-keys.ts
DashKash54: I think we should make this change as in our experience the nodes returns the value if they're hit, so we're reasonably confident about it and it reduce the load on the network as well @MaximusHaximus

packages/wrapped-keys-lit-actions/src/lib/solana/signTransactionWithEncryptedSolanaKey.js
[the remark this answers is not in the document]
MaximusHaximus: Already done <3 :0

packages/wrapped-keys/src/lib/lit-actions-client/batch-generate-keys.ts
MaximusHaximus: All set!

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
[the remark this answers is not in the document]
MaximusHaximus: :+1: Modified this test file -- but I was mistaken about it coming from `testEthereumBroadcastTransactionGeneratedKey` and the place it did come from was slightly different, so i left that one alone in the interest of time since this is kind of a :nit: and the existing test has been GTG

local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts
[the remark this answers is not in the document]
MaximusHaximus: :+1: gotcha. Removed them after changing the method of verifying

MaximusHaximus:
@DashKash54 All set

- Replaced `numResponsesRequired` with `runOnSingleNode` which sends only the request to a single, random node.
- I went ahead and updated the per-network generate LA's to use the new `runOnSingleNode` too
- I updated the CID repository with the latest CIDs

DashKash54 (approved):
LGTM!

PR #676, opened 2024-10-09
feat: only wait for threshold nodes when signing session keys

FedericoAmura:
# Description

This PR requires only the threshold nodes to sign sessions as it uses BLS, not ECDSA. Also fixes a check that was discarding node responses with an index value of `0` because it is considered `falsy` in JS

## Type of change

# How Has This Been Tested?

- [x] Integration tests validate this part of code (failed before the index fix)
- [x] Locally with a script that uses PKPs and LAs with session keys

# Checklist:

### DIFF
--- a/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
+++ b/packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
@@ -1869,7 +1869,7 @@ export class LitNodeClientNodeJs
       res = await this.handleNodePromises(
         nodePromises,
         requestId,
-        this.connectedNodes.size
+        this.config.minNodeCount
       );
       log('signSessionKey node promises:', res);
     } catch (e) {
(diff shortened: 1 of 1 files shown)


joshLong145 (approved):

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -1934,7 +1934,11 @@ export class LitNodeClientNodeJs
         for (const field of requiredFields) {
           const key: keyof BlsResponseData = field as keyof BlsResponseData;
 
-          if (!data[key] || data[key] === '') {
+          if (
+            data[key] === undefined ||
+            data[key] === null ||
```
Ansonhkg, line 1939: nit: i think `data[key] == null` will be true for both `null` and `undefined`

Ansonhkg (approved):
lgtm!

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
FedericoAmura: Screenshot from 2024-10-09 19-48-23
Sadly for us now, no. Javascript magic 🪄

PR #686, opened 2024-10-11
LIT-3956 - wrapped-keys - Allow batch of keys to be stored atomically (storeEncryptedKeyBatch)

MaximusHaximus:
# Description
### New client API method `storeEncryptedKeyBatch()`, and associated param and result types
- Added new method to allow consumers of the API to persist a batch of keys to the backend storage service in a single, atomic, all-or-nothing request
- Updated existing API method `batchGeneratePrivateKeys` to use the new atomic method service-client method (`storePrivateKeyBatch()`) to write keys instead of calling multiple, independent `storePrivateKey` API calls in parallel.

## Type of change

# How Has This Been Tested?

- [x] Ran localtests against `datil-test` (thus the TestNetworks) instance of our wrapped-keys REST API endpoints
- [x] Added new test for writing a batch of keys to the service with an _invalid_ entry in the array and verified that it throws the appropriate error from the backend

# Checklist:


packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
```diff
@@ -41,34 +41,40 @@ export async function batchGeneratePrivateKeys(
...
 
-  const results = await Promise.all(
-    actionResults.map(
-      async (result): Promise<BatchGeneratePrivateKeysActionResult> => {
-        const { generateEncryptedPrivateKey, network } = result;
+  const keyParamsBatch = actionResults.map((keyData) => {
+    const { generateEncryptedPrivateKey, network } = keyData;
+    return {
+      ...generateEncryptedPrivateKey,
+      keyType: getKeyTypeFromNetwork(network),
```
DashKash54, line 48: This is for Solana/EVM?

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
```diff
@@ -41,34 +41,40 @@ export async function batchGeneratePrivateKeys(
...
+    litNetwork: litNodeClient.config.litNetwork,
+  });
 
-        const signature = result.signMessage?.signature;
+  const results = actionResults.map(
+    (actionResult, ndx): BatchGeneratePrivateKeysActionResult => {
+      const {
+        generateEncryptedPrivateKey: { memo, publicKey },
+      } = actionResult;
+      const id = ids[ndx]; // Result of writes is in same order as provided
```
DashKash54, line 64: What do the ids mean and can you remind me what is it used for? In the same order means EVM first then Solana, how is it determined or does it even matter a long as the process is atomic?

packages/wrapped-keys/src/lib/api/store-encrypted-key-batch.ts
```diff
@@ -0,0 +1,54 @@
...
+        dataToEncryptHash,
+        ciphertext,
+        keyType,
+      })
+    );
+
+  return storePrivateKeyBatch({
+    pkpAddress,
+    storedKeyMetadataBatch,
+    sessionSig: getFirstSessionSig(pkpSessionSigs),
```
DashKash54, line 51: Already extracted the sessionSig above
```suggestion
    sessionSig,
```

packages/wrapped-keys/src/lib/service-client/types.ts
```diff
@@ -30,6 +30,14 @@ export interface StoreKeyParams extends BaseApiParams {
   >;
 }
 
+export interface StoreKeyBatchParams extends BaseApiParams {
+  pkpAddress: string;
```
DashKash54, line 34: Why is `pkpAddress` required as it isn't require in `StoreKeyParams`?

packages/wrapped-keys/src/lib/api/store-encrypted-key-batch.ts
```diff
@@ -0,0 +1,54 @@
...
+        publicKey,
+        memo,
+        dataToEncryptHash,
+        ciphertext,
+        keyType,
+      })
+    );
+
+  return storePrivateKeyBatch({
+    pkpAddress,
```
DashKash54, line 49: Why is this required isn't the `pkpAddress` extracted from the PKP sessionSig?

packages/wrapped-keys/src/lib/types.ts
```diff
@@ -101,6 +101,31 @@ export interface StoreEncryptedKeyResult {
...
+
+/** Result of storing a batch of private keys in the wrapped keys backend service
+ * Includes an array of unique identifiers, which are necessary to get the encrypted ciphertext and dataToEncryptHash in the future
+ *
+ * @typedef StoreEncryptedKeyBatchResult
+ * @property { string } pkpAddress The LIT PKP Address that the key was linked to; this is derived from the provided pkpSessionSigs
+ * @property { string[] } ids Array of the unique identifiers (UUID V4) of the encrypted private keys in the same order provided
+ */
+export interface StoreEncryptedKeyBatchResult {
+  ids: string[];
```
DashKash54, line 125: Are the ids used by the user in any way?

DashKash54 (commented):
Looks good left a few comments

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
MaximusHaximus: Yep! it takes in the network string that is user-provided (evm or solana), and returns the type of key for that network (ed25519 or K256)

packages/wrapped-keys/src/lib/api/batch-generate-private-keys.ts
MaximusHaximus: The id is a UUIDv4 that we generate when storing into DynamoDB -- it has no direct meaning, it's just used as a unique identifier for that persisted key from then on.  When a consumer uses the `list` method on the wrapped keys client, it returns an array of all keys persisted for the session sig's PKP, including the ID.  When someone needs the encrypted data (dataToEncryptHash, cipherText) for a particular key, they the the ID of the key they want to the `getEncryptedKey()` method :+1:

packages/wrapped-keys/src/lib/api/store-encrypted-key-batch.ts
MaximusHaximus: Good question! I checked into this, and it was a bit of a rabbit hole -- I realised that this method _doesn't_ need it -- but `list` and `get` do -- I _think_ it just ended up in all of the `service-client` methods dependencies parasitically even though we don't always need it.

This question unearthed an issue I hadn't noticed before, however -- we were importing the `getPkpAddressFromSessionSig()` method into `listPrivateKeyMetadata()` and `fetchPrivateKey()` _but we shouldn't have been_  the original design intention for `wrapped-keys` was that these operations would always be handled in the `api` layer and the result of them would be passed into the `service-client`, so we never ran those methods multiple times unnecessarily while servicing a `api` client call.

Thankfully, everything in the `service-client` is private so I fixed this problem and removed the `pkpAddress` parameter from the `service-client` parameter types that it was not needed for.

Fixing this encapsulation issue also exposed that we had imported the `service-client` and the internal util method into our `testImportPrivateKey` local-test -- so I fixed that too ;)

Good catch!

packages/wrapped-keys/src/lib/service-client/types.ts
MaximusHaximus: Good catch -- see https://github.com/LIT-Protocol/js-sdk/pull/686#discussion_r1797487662 - I resolved this

packages/wrapped-keys/src/lib/types.ts
MaximusHaximus: The _end-user_ generally would never see them -- they would see the `memo` field and probably public key -- the `id` is only used by developers when querying specific keys encrypted data :+1:

DashKash54 (approved):
LGTM!

PR #691, opened 2024-10-15
ci: testing snapshot container

joshLong145:
testing new container with snapshot in ci

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     env:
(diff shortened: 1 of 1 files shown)


Ansonhkg (changes requested):
Is the volume a blocker to getting this PR merged?

Ansonhkg (commented):
Is the volume a blocker to getting this PR merged?

joshLong145:
> Is the volume a blocker to getting this PR merged?

The container used in this PR was published as a test. But does provide some improvements to CI times as you no longer need to install modules for lit assets. The test container was published to a tag so it should persist if you want to merge this pr as is, if not can be closed.

PR #692, opened 2024-10-15
Rename `storyProtocolTestnet` to `storyIlaid`. Add `campTestnet` and `hushedNorthstar`

spacesailor24:
Requires: https://github.com/LIT-Protocol/accs-schemas/pull/20


Ansonhkg (approved):
Sehr Gut!

PR #694, opened 2024-10-16
feat: add human readable format for attentuation in sessionSigs

Ansonhkg:
# Description

```
import { formatSessionSigs } from '@lit-protocol/misc'
```

image


FedericoAmura (approved):
Nice

PR #698, opened 2024-10-16
feat: build lit actions and update cids when building sdk

FedericoAmura:
# Description

This PR:
1. When building the SDK => builds Lit Actions in `wrapped-keys-lit-actions` and updates the associated CID in `LIT_ACTION_CID_REPOSITORY` and `LIT_ACTION_CID_REPOSITORY_COMMON`
2. Before publishing the SDK => publishes lit actions code to IPFS

Both things are attached to our existing processes to enforce consistency of the built, referenced and published code

## Type of change

# How Has This Been Tested?

- [x] Tested triggering our current `sync-actions-to-ipfs` script automatically on a publish dry run
- [x] Tested building and updating references locally

Screenshot from 2024-10-17 17-18-29

# Checklist:


Ansonhkg:
nit: I think we need to document where do we get the `LIT_IPFS_JWT` token and it's from [Pinata](https://app.pinata.cloud/)

packages/wrapped-keys-lit-actions/project.json
```diff
@@ -4,15 +4,15 @@
...
+    "prebuild": {
       "executor": "nx:run-commands",
       "options": {
         "commands": [
           "yarn --cwd ./packages/wrapped-keys-lit-actions generate-lit-actions"
         ],
         "cwd": "."
-      }
+      },
+      "dependsOn": []
```
FedericoAmura, line 15: The explicit dependencies definition avoids a circular dependency as both WK and WKLA depend on each other to build but on different targets

The explicit flow is as follows
`WKLA:build -> WK:build -> WK:prebuild -> WKLA:prebuild -> nothing`

packages/wrapped-keys/update-ipfs-cids.js
```diff
@@ -0,0 +1,82 @@
...
+  fs.writeFileSync(constantsPath, constantsContent, 'utf-8');
+}
+
+updateConstants()
+  .then(() => {
+    console.log('Constants file updated successfully!');
+  })
+  .catch((err) => {
+    console.error('Error updating constants:', err);
+  });
```
MaximusHaximus, line 82: I think if this fails, we actually would like the entire process to error out so we don't accidentally publish new actions _source code_, but with old CIDs -- if this fails, but the source code build succeeded, someone might be running totally different LIT action code when manually setting the code repository than when using CIDs, _for the same package version_.  IMO let's remove the catch() and just let the error bubble up

tools/scripts/pub.mjs
```diff
@@ -101,8 +100,23 @@ await question('Are you sure you want to publish to? (y/n)', {
     // await 1 second
     await new Promise((resolve) => setTimeout(resolve, 1000));
 
+    // first publish the wrapped keys lit actions by running yarn publish-lit-actions in ./packages/wrapped-keys-lit-actions
+    await spawnCommand(
```
MaximusHaximus, line 104: _Any_ time we publish the wrapped-keys-lit-actions package, even if done manually using a direct NPM command, we want this to happen -- can we make this happen as part of a pre-publish NPM script instead of bespoke logic and shell execution here?

e.g. -- if we add this to `prepublish` it should always run when publishing, as part of that process...

https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-scripts

packages/wrapped-keys/update-ipfs-cids.js
```diff
@@ -0,0 +1,82 @@
...
+      solana: await Hash.of(generateEncryptedSolanaPrivateKey),
+    },
+    exportPrivateKey: {
+      evm: await Hash.of(exportPrivateKey),
+      solana: await Hash.of(exportPrivateKey),
+    },
+    batchGenerateEncryptedKeys: await Hash.of(batchGenerateEncryptedKey),
+  };
+
+  Object.entries(newCIDs).forEach(([key, value]) => {
```
MaximusHaximus, line 58: Dealing with injecting types and Object.freeze() here makes this code quite awkward and brittle.

Rather than doing string replacement operations, what do you think about us extracting the content of the `LIT_ACTION_CID_REPOSITORY` to a dedicated `.json` file that we can _literally replace the entire contents of_, from this script?

We would then import that JSON file into the `constants.ts` file in the lit-actions-client, _recursively call Object.freeze() on the object_, and export it with the LitCidRepository type annotation?  Incidentally, in order to export the object `as const` (which I'd like for us to do in the future) would require extracting the `Object.freeze()` calls out of the object definition also, because you can't mark an object that is wrapped in `Object.freeze()` with the `as const` pragma.

Then this entire constant object could be a _literal_ JSON object that is imported, typed, frozen, and exported for consumers by code inside the wrapped-keys constants file.

packages/wrapped-keys-lit-actions/project.json
MaximusHaximus: <3 nice work dude.

MaximusHaximus (changes requested):
Rather than add this logic into our pub.mjs, I'd really like to keep the logic encapsulated in NPM pre-publish hook in the package itself.  Added a couple other comments in-line.  Nice work getting this moving -- I ran into the dependency issue myself a few weeks ago but didn't figure out the explicit dep fix at the time. <3

packages/wrapped-keys/update-ipfs-cids.js
FedericoAmura: Good point. Removed

tools/scripts/pub.mjs
FedericoAmura: Good point. Changed this to this two things
- on LAWK: a prepublishOnly script to build and publish LAs on IPFS when publishing
- bonus, on LA: a prepare script to update its CIDs on publish but also on install. Installation should be trivial but useful to allow us to spot any issue

packages/wrapped-keys/update-ipfs-cids.js
FedericoAmura: Tried everything in a single operation on a TS file as `const`, I prefer JS/TS files over json. But it turned out with the same level of complexity so json it is. Updated

PR #702, opened 2024-10-22
LIT-3995 - (wrapped-keys) - Add signTransaction functionality to supported actions for `batchGeneratePrivateKeys`

MaximusHaximus:
# Description
- Allows consumers to provide a transaction to be signed by each key they request be generated, in the same LIT action execution as the key is generated in.  Basically matches the sign message capability, but for transactions.
- I'm not sure how useful this is with our Solana integration; the existing sign transaction test logic fails with 'unknown signer' -- I **think** that's because the serialized TX expects the `fromPubkey` and `feePayer` in the serialized TX to match the pubKey of the generated key that is doing the signer -- but obviously that can't happen since we generate the key in the LIT action runtime. 🤔 

## Type of change

# How Has This Been Tested?

- [X] Ran all existing wrapped-keys local-tests
- [X] Updated `local-tests/tests/wrapped-keys/testBatchGeneratePrivateKeys.ts` to sign an ethereum tx and verify the signature returned is hex (same logic as our existing signTransactionEthereum test)

# Checklist:


packages/wrapped-keys-lit-actions/src/lib/internal/ethereum/signTransaction.ts
DashKash54, line 1: @MaximusHaximus we don't wanna allow fetching nonce, gas, etc or anything from the RPC. It should all be provided by the user either as a serialized tx or as params and we could construct the transaction inside the LA but we should not fetch anything from the RPC we can remove this fetching altogether

DashKash54 (changes requested):
The reason it works for Ethereum tx is because internally in the `signTransactionEthereumKey()` we add the from address: https://github.com/LIT-Protocol/js-sdk/blob/1e074af718772bb88b0df87da986393d30db05c8/packages/wrapped-keys-lit-actions/src/lib/internal/ethereum/signTransaction.ts#L169
We always need to provide the `from` address (and `feePayer` as well in Solana tx). Since we're generating the wallet and signing right after that the user can just provide the params that we can use to craft a tx object inside the LA itself.
**Note:** The user should always provide the gas and anything else that currently requires an RPC call as we should no longer support making RPC calls from within the LA as it causes arbitrary delay due to the blockchain. Therefore, we shouldn't even support tx broadcast or at least not in the `batchGenerate()`. Since we're just creating the wallet the nonce will be 0. @MaximusHaximus lemme know if you have any questions

MaximusHaximus:
This was cancelled / un-necessary; closing for now.

PR #710, opened 2024-11-04
fix(lit-core): LIT-4016 - Enhance error handling during epoch changes

MaximusHaximus:
# Description

**Updated error handling and logic flow for handling epoch change events, and signalling errors during epoch change processing to consumers/listeners**

- LitCore is now an event emitter
  - Used [EventEmitter3](https://github.com/primus/eventemitter3) which is extremely widely used, ESM compatible, and written in ES3 for maximum compatibility (!)
- Events for `disconnected`, `connected` and `error` are emitted
- We no longer call `_stopListeningForNewEpoch()` during `_connect()`; concurrent calls to `connect()` are chained automatically _already_.
  - This fixes a corner case where failures to connect during event handling would cause LitCore to stop listening for future epoch changes.
- Error handling / execution flow updated in `_handleStakingContractStateChange()`
  - Errors encountered anywhere in the process will emit error/disconnected events appropriately
  - `this.ready` is now set to `false` when epoch change events are not processed correctly
- We now throw an explicit `NotReady` error from `getSessionSigs()` if we get to the point we're going to map across `this.connectedNodes`, but the `LitCore` instance is not ready.

## Type of change

# How Has This Been Tested?
- [X] Ran all local-tests

# Checklist:


packages/core/src/lib/lit-core.ts
```diff
@@ -492,11 +507,6 @@ export class LitCore {
   }
 
   private async _connect() {
-    // Ensure an ill-timed epoch change event doesn't trigger concurrent config changes while we're already doing that
```
MaximusHaximus, line 495: Removed this call to stop listening -- if the code inside our listener calls `connect()`, it will just chain on any already-pending `connect()` logic (see `connect()` logic and promise chain on `this._connectingPromise.

DashKash54:
Currently the loading of modules in the `connect()` is coupled with the node handshake? it's possible that the handshake fails for any reason so the `this.ready=false` but we'll still try to reload the WASM modules even though they've been loaded in the `gloabalThis`? Maybe multiple loads isn't that bad or shall we revisit this?

DashKash54:
Are we able to successfully & repeatedly reproduce this error @MaximusHaximus ? How has the fix been tested?

packages/core/src/lib/lit-core.ts
```diff
@@ -330,39 +328,55 @@ export class LitCore {
...
-          // this.ready = false; Should we assume core is invalid if we encountered errors refreshing from an epoch change?
-          const { message = '' } = err as
-            | Error
-            | NodeClientErrorV0
-            | NodeClientErrorV1;
-          logError(
-            'Error while attempting to reconnect to nodes after epoch transition:',
-            message
+        } else {
+          // In case of centralised networks, we don't run `connect()` flow, so we will manually update epochInfo here
```
DashKash54, line 350: Why is a centralized network treated differently? The only difference should be in the attestation but the handshake should be invariant to the network?

packages/core/src/lib/lit-core.ts
DashKash54: But will it chain the promises since we specifically check that if a pending connection is open then just return it?

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
```diff
@@ -2203,6 +2203,18 @@ export class LitNodeClientNodeJs
 
     const signatures: SessionSigsMap = {};
 
+    if (!this.ready) {
```
DashKash54, line 2206: Okay `getSessionSig()` was the only function missing `this.ready` check. Actually we check this in the `signSessionKey()`?

DashKash54 (commented):
Left a few comments

packages/core/src/lib/lit-core.ts
MaximusHaximus: We always re-set `_connectionPromise` to null when a `connect()` call finishes -- so if we are connecting, multiple calls to `connect()` will all return the same promise.  This means that if we're still processing `connect()` when we receive another epoch change event, it will be effectively a no-op.

It occurred to me that we could implement cancellation across this entire callstack, and then cancel the entire existing call to connect() (along with any pending fetch() calls that it is running), and then run it again from the top. I'd like to implement a much more robust network handling layer based on the `v7` branch and land this as-is for now, since it's an improvement over what we've got.

MaximusHaximus:
> Can you please confirm that in the `_connect()` even if we're unable to handshake to a single node we never set `this.ready=true`? @MaximusHaximus @Ansonhkg

Yes, that's right -- if any handshakes fail, the entire connect() chain rejects immediately

packages/core/src/lib/lit-core.ts
MaximusHaximus: My understanding is that for centralized networks, the node list doesn't change, so re-handshaking with all the nodes on epoch change would be redundant; if that's not true, we should definitely fix it...

MaximusHaximus:
> Currently the loading of modules in the `connect()` is coupled with the node handshake? it's possible that the handshake fails for any reason so the `this.ready=false` but we'll still try to reload the WASM modules even though they've been loaded in the `gloabalThis`? Maybe multiple loads isn't that bad or shall we revisit this?

Our crypto module code checks to see if globalThis.<module name> has been set, and if it has, it does nothing at all. I agree it's a bit confusing -- the good news is that in v7+, we no longer keep global state around and there is no `loadModules()` method -- so this is temporary logic, thankfully :)

packages/lit-node-client-nodejs/src/lib/lit-node-client-nodejs.ts
MaximusHaximus: We do check it in `signSessionKey()`, but that method isn't called from `getSessionSigs()` -- it is only called from the `authNeededCallback()` defined in `getPkpSessionSigs()` and in some auth providers -- so I added it here for completeness :+1:

MaximusHaximus:
> Are we able to successfully & repeatedly reproduce this error @MaximusHaximus ? How has the fix been tested?

Unfortunately this is a very corner-case issue that is caused by failures in entirely internal code.  It also requires that we trigger epoch changes to actually verify that the fix is working :(

With very creative Jest mocks, and long-running Shiva tests, I can write a test case that will reproduce it consistently and verify this fix is always present, but our current local-tests don't really facilitate this degree of testing, and we don't have live epoch-change tests yet :(.  I was, however, able to test the fix by adding manual `throw Error()` calls into specific spots in the SDK callstack and verifying the SDK followed the right code paths :+1:

packages/core/src/lib/lit-core.ts
```diff
@@ -308,20 +310,16 @@ export class LitCore {
...
+    try {
+      log(`New state detected: "${state}"`);
 
-    if (state === StakingStates.Active) {
-      // We always want to track the most recent epoch number on _all_ networks
-
-      this._epochState = await this._fetchCurrentEpochState(
-        validatorData.epochInfo
-      );
+      if (state === StakingStates.Active) {
```
FedericoAmura, line 316: Shouldn't we throw/disconnect when the new state is `Paused`?

FedericoAmura (commented):
LGTM but it is pointing to master with v7 code. This should be updated (and use v7 errors) or point the PR to a v6 branch

packages/core/src/lib/lit-core.ts
```diff
@@ -330,39 +328,55 @@ export class LitCore {
...
+      this.emit(
+        'error',
+        new Error(
+          'Error while attempting to reconnect to nodes after epoch transition:' +
+            message
+        )
+      );
+
+      // Signal to any listeners that we're 'disconnected' from LIT network
+      this.emit('disconnected', true);
```
FedericoAmura, line 379: I would prefer listening to only one event. We have `connected` below. Here we can emit:
```suggestion
      this.emit('connected', false);
```

As a listener I would expect to listen to connection status as this is likely the status I'm interested in (to know if I can use Lit), instead of just one of those events

MaximusHaximus: Emitting `disconnect` and `connect` events is a pretty stand/common pattern, since the action taken is typically very different for each case; it's very common to define `onDisconnect` and `onConnect` functions in the app logic layer and just bind them directly as event handlers. If we emit a single `connection` event with the state of the connection, then every event handler in the app layer is probably going to start with an if() else {} block that needs to check the payload emitted to the event, and identify the type, and call the appropriate connect or disconnect handler.

Also, most consumers will actually not really care about _connected_ events because the `connect()` method itself returns a promise that they can simply wait on -- but listening to `disconnect` events is actually quite important for all consumers.

I personally think emitting separate events is a nice devex -- but I've got no argument against adding another _connection_ type event that is emitted every time the connection state changes, either connecting or disconnecting, if you think that would be a good addition to the API :+1:

One thing to consider is that we can also strongly type `disconnectHandler` and `connectHandler` functions -- and the payload provided to the listener would likely make sense to actually be a different type.  Actually, I think that my making this always emitt a boolean is probably a poor API, now that you mention it -- what do you think about modifying the `disconnect` event to emit the `Error` that was encountered during the disconnect, e.g. `this.emit('disconnected', { reason: error })`?

packages/core/src/lib/lit-core.ts
MaximusHaximus: I agree - let's change this for v8

packages/core/src/lib/lit-core.ts
MaximusHaximus: Good call, if so -- if `Paused` state on the contract means the existing node set can no longer serve requests, we should do exactly that.  I was under the impression that `Paused` meant epochs might not be changing, but the _current epoch of nodes still functioned_.  If I'm wrong, we should definitely change it as you suggested

packages/core/src/lib/lit-core.ts
FedericoAmura: Great points. The thing that tickled me was that we were emitting events with a boolean that is always the same, seems like it is neither a `connected` status API nor a `onConnect` + `onDisconnect` events

But your last suggestion I think nails the thing. And we can make something similar for `connected` when we see the need for it

PR #736, opened 2024-12-11
Fix CI Build: what happens if we get rid of nrwl packages and replace with nx equiv…

glitch003:
…ilant?  the build issue seems to be related to conflicting nx versions

Ref: https://github.com/nrwl/nx/issues/26517

The error in our CI: 
```
[4/4] Building fresh packages...
error /home/runner/work/js-sdk/js-sdk/node_modules/@nrwl/eslint-plugin-nx/node_modules/@nrwl/tao/node_modules/nx, /home/runner/work/js-sdk/js-sdk/node_modules/@nrwl/eslint-plugin-nx/node_modules/@nx/workspace/node_modules/nx: Command failed.
Exit code: 135
Command: node ./bin/post-install
Arguments: 
Directory: /home/runner/work/js-sdk/js-sdk/node_modules/@nrwl/eslint-plugin-nx/node_modules/@nrwl/tao/node_modules/nx
Output:
Bus error (core dumped)
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

PR #737, opened 2024-12-12
Get latest commit hash for shiva in CI

glitch003:
Right now we have a hardcoded lit actions commit hash.  This will try to pull down the latest and use that, instead.  Github artifacts expire 30 days after publishing so this is prob why CI broke.  

This is an extension of https://github.com/LIT-Protocol/js-sdk/pull/736 where I debugged the underlying issue (well, multiple issues) and found that the commit hash was the first broken thing (because the artifact gets deleted after 30 days), and then we needed to change the `rust/lit-node-build-commit-hash` ubuntu version to 22, which finally fixed it.

This uses our own fork of an action that will pull the latest hash from the lit-assets repo, which is here: https://github.com/LIT-Protocol/last-successful-build-action

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     env:
-      DATIL_COMMIT_HASH: 2f38a85e7b72c1f352ec51b05768db7971697b2c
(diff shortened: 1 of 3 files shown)


FedericoAmura (approved):
Great findings!!!

Ansonhkg (approved):

PR #738, opened 2024-12-13
fix: undefined tokenId

Ansonhkg:
# WHAT

When calling the `contractClient.pkpNftContractUtils.read.getTokensInfoByAddress` function, `tokenId` is returning `undefined`. This little fix added a check to ensure addrs.tokenId is assigned a value if it’s missing. If the `derivedAddresses` function doesn’t return a `tokenId`, the `tokenId` from the context (tokenId) is used as a fallback.


packages/contracts-sdk/src/lib/contracts-sdk.ts
```diff
@@ -1903,10 +1903,14 @@ https://developer.litprotocol.com/v3/sdk/wallets/auth-methods/#auth-method-scope
         for (let i = 0; i < tokenIds.length; i++) {
           const tokenId = tokenIds[i];
           const pubKey = await this.pkpNftContract.read.getPubkey(tokenId);
-          const addrs = await derivedAddresses({
+          const addrs: TokenInfo = await derivedAddresses({
```
FedericoAmura, line 1906: Probably should make the `tokenId` optional in `TokenInfo` to keep Typescript consistency

FedericoAmura (approved):

PR #739, opened 2024-12-13
Feature: added versioned transaction solana signature

kouraf:
#### Summary of Changes
I have introduced a new feature to the LIT SDK and Actions to support versioned transactions for Solana. Specifically:
- Added a parameter called `versionedTransaction` to enable handling of versioned transaction signatures for wrapped keys.
- Updated the LIT Action to utilize this param.
- if `versionedTransaction` is `false` it uses the legacy transaction

#### Testing
- I have tested the action thoroughly, and it works great with the new parameter.
- For the SDK, while the code changes align with the intended functionality, I haven't been able to verify the implementation entirely. Additional testing might be necessary.

#### Notes
- I'm unsure if there's a specific coding style or conventions used in the codebase. If there are, I’d be happy to refactor the code to match the style!
- I'm available to assist further with these changes or address any feedback to ensure the implementation aligns with your expectations.

Looking forward to hearing your thoughts!


FedericoAmura:
This is awesome. Thanks @kouraf !

I've made some small changes: formatting and base update, but also made `versionedTransaction` optional so we don't require consumers to specify it (defaulting to falsy and previous behaviour after all) and we can release this as a minor change.
As a followup we can require it set to `true` with a versioned transaction, or let the action itself decide based on the type of transaction it receives

FedericoAmura:
> Hello! Hello! @FedericoAmura i just pushed an exemple you can see and test [here](https://github.com/kouraf/test-wrapped-key-versioned-tx) feel free to let me know your thoughts 😄
> 
> I'm not entirely sure about allowing the action itself to decide based on the type of transaction it receives. A few transactions are compatible with both "legacy" and "versioned" formats, and attempting to auto-detect them based on their base64 encoding could lead to errors. However, an interesting feature to consider adding later could be multisigning.

Thanks for the tests! Looking perfect to me

What I was thinking is that they can pass the wrong `versionedTransaction`. We can enforce some of that with types but not entirely when using JS instead of TS

Nice addition. Thanks a lot for your contribution!

FedericoAmura (approved):

PR #743, opened 2024-12-20
Update arb sepolia RPC. Add Vana. Bump accs-schemas package version

spacesailor24:


Ansonhkg (changes requested):
@spacesailor24 can we update the yarn lock file (run yarn) and run prettier (run yarn prettier) then commit the relevant files please?

PR #766, opened 2025-01-15
fix(ci): update Rust version to 1.83.0 for wasm compatibility

Ansonhkg:
# WHAT

Updated Rust in CI to 1.83.0 to fix wasm build issues.


FedericoAmura (approved):
Awesome, needed this. Thanks!

PR #784, opened 2025-01-31
LIT-3748 - Naga price support

MaximusHaximus:
# Description
This PR implements 'complete' price support for Naga.

### Setting the max price for a request
There are two ways SDK consumers can set the max price they want to pay for a request.
1. Existing LitNodeClient behaviour methods (e.g. `decrypt`, `executeJs` and `pkpSign`) have a new optional argument, `userMaxPrice: bigint`. If a `userMaxPrice` is provided when calling those methods, it will be used to calculate how much the `maxPrice` should be for each node.
2. SDK consumers can also configure _default_ values for how much they want to pay for each type of request (SIGN, DECRYPTION, LIT ACTION) on the LitNodeClient by calling `litNodeClient.setDefaultMaxPrice(product: keyof typeof PRODUCT_IDS, price: bigint)`.  If the LitNodeClient is configured with a specific default max price for a particular product, and no `userMaxPrice` is provided when the user calls that product method, the default will be used instead.
3. If no userMaxPrice is provided when calling behaviour methods on LitNodeClietn _and no default has been set for that product_, then the maxPrice for each node will be _un-bound_, and as a result the request cost will be whatever the cheapest subset of nodes that can execute the request charge at that time.
    - Tinny's LitNodeClient can now be configured via environment variable `DEFAULT_MAX_PRICES`, which is an array of strings that will be parsed as bigint.

- If the userMaxPrice provided by the caller or set on the LitNodeClient is too low for current network conditions and your request cannot be fulfilled without spending more than `userMaxPrice`, the LitNodeClient will throw a `MaxPriceTooLow` error with context.

### SessionSigs are no longer consumer-facing
- Note that pricing differs depending on the product being used -- signing can have a different price than running a LIT action, for example.
- As a result of this variability, we needed to create sessionSigs _on the fly_, because sessionSigs must include a maxPrice parameter for _each node we are sending a request to_, and the maxPrice will be different depending on the _type_ of request being sent.
- Because sessionSigs must be created on-the-fly, we have internalized their creation entirely.  Instead of users passing `sessionSigs` to our behaviour methods, users pass an `AuthenticationContext` object, which contains the _parameters to be used to generate session sigs_.  The LitNodeClient then uses the `AuthenticationContext` to create session sigs with the appropriate maxPrice for each node just before it sends the requests to the nodes.
- Due to this change, `getSessionSigs()` is now a private method on the LitNodeClient (`_getSessionSigs()`). It is not expected for SDK consumers to need to create sessionSigs during typical usage from now on.

### Details
- Refactored tracking of node prices to use BigInts instead of numbers
- Pricing data is always fetched for all product IDs -- there are no cases where we ever want pricing indexes to be different from call-to-call since the data returned is positional, not an object
- Pricing data is stored in an array sorted by the DECRYPTION product, containing `{url: string, prices: bigint[] }` objects
- Removed most `SessionSigs` related types, and replaced them with `AuthenticationContext`
- Removed deprecated code from `pkp-base` and `authClient`
  - Replaced (deprecated) `controllerAuthMethods`,`controllerSessionSigs` and `controllerAuthSig` with just `AuthenticationContext`
  - I didn't add `maxUserPrice` to `pkp-base` (and objects that extend from it) behavioural methods.  I am keen for feedback on whether we think being able to use `setDefaultMaxPrice()` on the LitNodeClient instance will be sufficient for usage of these objects. cc: @Ansonhkg @FedericoAmura 
- Added support for loading PriceFeed contract ABIs from Shiva TestNetwork configuration
- Updated error logging for in `tinny-operations` so that we log VErrors appropriately when tests fail, both in-line _where the test failed_, and at the end when showing test failure results, and include VError.info() in printed output when it exists on the error that we catch.
- Updated all local-tests to use the new `AuthenticationContext` pattern instead of using `getSessionSigs()`.
- I added explicit eslint-ignore pragmas to a few files where we have a lot of `any` datatypes.  Files like `interfaces`, `types`, `contracts-sdk` are now "clean", so if you see errors in your IDE gutter in those files you can be confident your changes have introduced them :)
  - I removed `timestamp2Date()`, which had a dynamica `require()` of the `date-and-time` package.  I couldn't find anywhere we use the function.
- @FedericoAmura You are a saint -- nice work on killing the autogen-injected contracts-sdk code.  I've enabled our `import/order` rule in that file now that it's not being dynamically built.  This should help a lot with conflicts while merging/rebasing code in that file from here on out
- I removed the interface for PricingInfo; nothing external needs to consume it, so keeping it out of our advertised interfaces makes it easier to change as we iterate without worrying about it being an API level breaking change.
- The set of Nodes that requests are sent to (and the `nodeSet` provided to those nodes) is now dynamic; we use the required # of nodes for consensus that are also the _cheapest priced nodes_.
- lit-auth-client's AuthProviders had some funky types/interfaces and inheritance that I unravelled, centered around the `authenticate()` abstract method. I removed the `extends AuthenticateOptions` which was a pass-through for an empty interface.
- I added basic unit tests for `get-max-prices-for-node-product`
- I removed the already-deprecated `runOnTargetedNodes()` method that we haven't used in quite some time; this allowed us to also remove `getIpfsId()` from LitNodeClient - I couldn't find anywhere other than `runOnTargetedNodes()` that it was actually used, and I would think that ipfsIds should be able to be computed client-side without running a LIT action anyway (right?)

### Notes:
1. Currently wrapped-keys tests, builds, and publishing is disabled. Wrapped-keys is a bit of an interesting case because we use sessionSigs to authenticate, and the backend doesn't yet support any Naga featureset networks.
2. I removed the `LitClientSessionManager` interface; it was very sticky with SessionSigs, and I think it will be replaced entirely with our AuthManager pattern

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- [X] Ran all local-tests against lit-assets @ 60318791258d273df8209b912b386680d25d0df3)

# Checklist:

- [] I have made corresponding changes to the documentation


Ansonhkg (approved):
LFG

PR #788, opened 2025-02-04
add methods and struct for e2ee

mikelodder7:
# Description

This adds the methods that will be necessary for End-to-End Encryption with lit-node coming shortly after Naga release.
This only uses tweet-nacl and thus no new dependencies are added.

## Type of change

# How Has This Been Tested?

I'm not familiar with how to test in the SDK but would love help in adding unit tests for this.


packages/crypto/src/lib/crypto.ts
```diff
@@ -372,6 +373,103 @@ async function getAmdCert(url: string): Promise<Uint8Array> {
...
 }
 
+export const walletEncrypt = async(
+  myWalletSecretKey: Uint8Array, 
+  theirWalletPublicKey: Uint8Array,
+  sessionSig: Uint8Array,
+  message: Uint8Array
+): Promise<WalletEncryptedPayload> => {
+  const random = new Uint8Array(16);
+  window.crypto.getRandomValues(random);
```
FedericoAmura, line 383: Using `window` won't work on node side, neither deno if we want to put the SDK in LAs. However the method exists on nodes and denos crypto module

packages/crypto/src/lib/crypto.ts
```diff
@@ -372,6 +373,103 @@ async function getAmdCert(url: string): Promise<Uint8Array> {
...
+  const dateNow = Date.now();
+  const createdAt = Math.floor(dateNow / 1000);
+  const timestamp = Buffer.alloc(8);
+  timestamp.writeBigUInt64BE(BigInt(createdAt), 0);
+
+  const myWalletPublicKey = new Uint8Array(32);
+  nacl.crypto_scalarmult_base(myWalletPublicKey, myWalletSecretKey);
```
FedericoAmura, line 390: nit: if we can use @noble curves and hashes way better so we move towards lib unification (and drop tweetnacl)

Ansonhkg (changes requested):
need to handle node environment and unit test is required

---
Update:
- Fixed them here https://github.com/LIT-Protocol/js-sdk/pull/815, review needed @DashKash54 @mikelodder7

PR #660, opened 2024-09-27
WIP: feat: start using zod for schema validation and type definitions

FedericoAmura:
# Description

This PR incorporates `zod` as the schema validation tool and derives types from those

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #801, opened 2025-02-19
feat: add new schemas package using zod, bring acc validation package…

FedericoAmura:
# Description

This PR replaces #660 which never got reviewed due to its size. It adds a new package, @lit-protocol/schemas which contains zod schemas we can use for input/output validation and as a connection between our constants and types

The packages dependency graph is as follows:
Screenshot from 2025-02-20 14-25-38

- `@lit-procotol/constants` as the most base package and containing our constants values defined `as const`. Optionally, they can have [types](https://github.com/LIT-Protocol/js-sdk/pull/801/files#diff-098538339ad726195cc4b9d0dba8fabce31a511a673e8450f02e233192d85e6a) to help unify multiple values but those are not needed to be exposed
- `@lit-protocol/schemas` as a child of constants. Its purpose is to define zod schemas, based on constants if needed, to provide complex runtime validations. This package is not intended for public use as it is internal to this SDK
- `@lit-protocol/types` as a child of schemas. Will derive types from the schemas it needs to derive complex typescript compile time validations
- Rest of packages will most likely depend from the three packages as they all contain useful resources

In this PR there is also a migration of https://github.com/LIT-Protocol/accs-schemas into `@lit-protocol/access-control-conditions-schemas` new package. Original package uses `ajv` to validate ACCs schemas, that was later imported in the SDK, now with zod in the SDK we can unify both and expose them into a public. This brings the benefit of having to update only one source of truth when adding support for a network in ACCs and a slight bundle reduction

## Type of change

# How Has This Been Tested?

- [x] CI still passing
- [x] Type system is still consistent with schema derived types
- [x] Unit tests over some schemas
- [x] ACCs validator and its tests migrated to new ACCs schemas pkg

# Checklist:


local-tests/package.json
```diff
@@ -20,7 +20,6 @@
     "@cypress/code-coverage": "^3.10.0",
     "@cypress/react": "^6.2.0",
     "@cypress/webpack-dev-server": "^2.3.0",
-    "@lit-protocol/accs-schemas": "0.0.7",
```
Ansonhkg, line 23: NICE!

Ansonhkg (approved):
LGTM! - a sentence or two guide on how do we add new chain support would be great!
packages/access-control-conditions-schemas/src/lib/common.ts

PR #802, opened 2025-02-26
Feature/lit 4238 clean unnecessary packages

FedericoAmura:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


local-tests/setup/session-sigs/get-pkp-session-sigs.ts
```diff
@@ -43,7 +42,7 @@ export const getPkpAuthContext = (
     }),
   });
 
-  log('[getPkpAuthContext]: ', authContext);
+  console.log('[getPkpAuthContext]: ', authContext);
```
FedericoAmura, line 45: Testing logs better using `console` than `pino`

package.json
```diff
@@ -112,7 +113,8 @@
     "eslint-plugin-jsx-a11y": "6.9.0",
     "inquirer": "^9.2.21",
     "ipfs-unixfs-importer": "12.0.1",
-    "jest": "27.5.1",
+    "jest": "^29.2.2",
+    "jest-environment-jsdom": "^29.7.0",
```
FedericoAmura, line 117: ts-jest was complaining about this old version

packages/auth/src/lib/authenticators/metamask/eth.ts
```diff
@@ -408,11 +378,8 @@ export const disconnectWeb3 = (): void => {
 
   const storage = LOCAL_STORAGE_KEYS;
 
-  localStorage.removeItem(storage.AUTH_SIGNATURE);
-  localStorage.removeItem(storage.AUTH_SOL_SIGNATURE);
-  localStorage.removeItem(storage.AUTH_COSMOS_SIGNATURE);
-  localStorage.removeItem(storage.WEB3_PROVIDER);
```
FedericoAmura, line 414: None of these were used

packages/auth/src/lib/authenticators/utils.ts
```diff
@@ -18,17 +17,6 @@ import { WebAuthnAuthenticator } from './WebAuthnAuthenticator';
...
 export const LIT_LOGIN_GATEWAY = 'https://login.litgateway.com';
 
-/**
- * Check if OAuth provider is supported
- *
- * @param provider {string} - Auth provider name
- *
- * @returns {boolean} - True if provider is supported
- */
-export function isSocialLoginSupported(provider: string): boolean {
```
FedericoAmura, line 28: Not used anywhere

packages/constants/project.json
```diff
@@ -15,12 +15,6 @@
         "updateBuildableProjectDepsInPackageJson": true
       }
     },
-    "copyJSONFilesToDist": {
```
FedericoAmura, line 18: Not used

packages/constants/src/lib/constants/constants.spec.ts
```diff
@@ -1,91 +1,13 @@
...
-    'baseGoerli',
-  ];
-
-  const MUST_HAVE_SOL_CHAINS = ['solana', 'solanaDevnet', 'solanaTestnet'];
-  const MUST_HAVE_COSMOS_CHAINS = [
-    'cosmos',
-    'kyve',
-    'evmosCosmos',
-    'evmosCosmosTestnet',
-  ];
```
FedericoAmura, line 44: Types enforce this now

Ansonhkg (approved):

Ansonhkg:
nit: should we use standardise using `#` instead of `_` for truly private fields for real runtime encapsulation in our codebase?

FedericoAmura:
> nit: should we use standardise using `#` instead of `_` for truly private fields for real runtime encapsulation in our codebase?

I think we should. It has been supported since 2021 (node v16 and chrome 91) so I don't see any reason why not use it. However I would keep it in a followup to avoid making this bigger

Opinions @MaximusHaximus ?

MaximusHaximus:
Just a reminder that tslib / typescript at multiple had issues with the # markup, so we had intentionally avoided using it in https://github.com/LIT-Protocol/js-sdk/pull/526 -- are we sure it's safe at this point? :D

FedericoAmura:
> Just a reminder that tslib / typescript at multiple had issues with the # markup, so we had intentionally avoided using it in #526 -- are we sure it's safe at this point? :D

A rapid replace and build succeeded
Screenshot from 2025-03-11 22-01-02

As per gpt
```
Older versions of tslib (pre‑2.0) didn’t include the helper functions (like __classPrivateFieldGet and __classPrivateFieldSet) that TypeScript emits when down‑leveling the new ECMAScript # private fields and methods. This meant that if you were using # notation and targeting an older ECMAScript version with tslib 1.x, your compiled code would break at runtime.

The fix is to update tslib to version 2.0 or later. Alternatively, you can let TypeScript inline the helper code instead of relying on tslib, or target a modern ECMAScript version that supports private fields natively.
```
But we are using tslib 2.7 and 2.3. We can let consumers know that tslib requires v2.0.0 [(which was released 5 years ago)](https://www.npmjs.com/package/tslib/v/2.0.0)

MaximusHaximus (changes requested):
One thing that is missing w/ the new logger changes is the ability to configure the SDK logger, which is a pretty fundamental function for the logger.  Since we create instances of `pino` in multiple places _in module scope_, and don't expose those instances, they are all using defaults for transports, formatting, log levels, etc.

I think we still need a (much simpler than before!) SDK-wide logger package, where sdk consumers can do things like set log level and transports for logs being generated by the SDK.

For example, even though we (and our SDK consumers), might want to see `trace` level logs sometimes, for every-day purposes many (most) consumers will not want to see that degree of granularity in the output. Right now, since the loggers are not exposed in any way, there is no way for their `level` to be set.

This also has further repercussions -- one of the big benefits of using `pino` for logging is that its _transports_, _serializers_, and _formatters_ can be configured for various purposes -- e.g. to ship logs to an aggregation service, pretty-print them, or output them as structured JSON.  Right now there is no way for consumers to _configure those transports_.

The simplest way to handle this is, rather than creating new pino instances directly inside of module scopes, we want the logger package to expose two basic methods:
(1) `setLoggerOptions()`, which can mutate the options that are provided to an sdk-wide singleton pino logger instance.
(2)`getLogger()`, which returns the singleton instance of our SDK pino logger. Anyt time `getLogger()` is called, the logger singleton instance will be constructed using the options that are set at that time if it does not already exist, otherwise the existing instance will be returned.

The idea is that, at the top of my app that uses the LIT SDK, before I call any SDK methods, I can configure the SDK logger completely using any supported pino `options`:
```
import { setLoggerOptions } from '@lit-protocol/logger';

setLoggerOptions({level: 10, transports: [...] }); // Here's where I would set up e.g. pino-pretty or other log formatting/destination options.
```

We can also allow mutation of logging configuration _even after it has been used_.  Any time `setLoggerOptions()` is run, we could reset any existing pino singleton, so that the next time code runs that calls `getLogger()`, a new one is created with the newly set options.

With this approach, when a consumer of the SDK wants to control the log verbosity, transports, serializers, etc, they can call our method `setLoggerOptions()` to mutate the options that we use when constructing the logger singleton, and then use the SDK as-normal. 

The critical change in structure inside our code is that everywhere we are currently constructing a logger in module scope, then using it directly inside of our function calls, we must modify it to call `getLogger()` and use the instance returned by that call to call `[logger.info](http://logger.info/)`, `logger.error`, etc. — on the sdk-wide logger instance.

If we want to provide helpers for consumers of the SDK, we could provide some examples of options combinations for e.g. pretty-printing logs, outputting them to files, or exporting them to an external log sink — all of these would be references to the existing `Pino` transport and options APIs!

FedericoAmura:
> One thing that is missing w/ the new logger changes is the ability to configure the SDK logger, which is a pretty fundamental function for the logger. Since we create instances of `pino` in multiple places _in module scope_, and don't expose those instances, they are all using defaults for transports, formatting, log levels, etc.
> 
> I think we still need a (much simpler than before!) SDK-wide logger package, where sdk consumers can do things like set log level and transports for logs being generated by the SDK.
> 
> For example, even though we (and our SDK consumers), might want to see `trace` level logs sometimes, for every-day purposes many (most) consumers will not want to see that degree of granularity in the output. Right now, since the loggers are not exposed in any way, there is no way for their `level` to be set.
> 
> This also has further repercussions -- one of the big benefits of using `pino` for logging is that its _transports_, _serializers_, and _formatters_ can be configured for various purposes -- e.g. to ship logs to an aggregation service, pretty-print them, or output them as structured JSON. Right now there is no way for consumers to _configure those transports_.
> 
> The simplest way to handle this is, rather than creating new pino instances directly inside of module scopes, we want the logger package to expose two basic methods: (1) `setLoggerOptions()`, which can mutate the options that are provided to an sdk-wide singleton pino logger instance. (2)`getLogger()`, which returns the singleton instance of our SDK pino logger. Anyt time `getLogger()` is called, the logger singleton instance will be constructed using the options that are set at that time if it does not already exist, otherwise the existing instance will be returned.
> 
> The idea is that, at the top of my app that uses the LIT SDK, before I call any SDK methods, I can configure the SDK logger completely using any supported pino `options`:
> 
> ```
> import { setLoggerOptions } from '@lit-protocol/logger';
> 
> setLoggerOptions({level: 10, transports: [...] }); // Here's where I would set up e.g. pino-pretty or other log formatting/destination options.
> ```
> 
> We can also allow mutation of logging configuration _even after it has been used_. Any time `setLoggerOptions()` is run, we could reset any existing pino singleton, so that the next time code runs that calls `getLogger()`, a new one is created with the newly set options.
> 
> With this approach, when a consumer of the SDK wants to control the log verbosity, transports, serializers, etc, they can call our method `setLoggerOptions()` to mutate the options that we use when constructing the logger singleton, and then use the SDK as-normal.
> 
> The critical change in structure inside our code is that everywhere we are currently constructing a logger in module scope, then using it directly inside of our function calls, we must modify it to call `getLogger()` and use the instance returned by that call to call `[logger.info](http://logger.info/)`, `logger.error`, etc. — on the sdk-wide logger instance.
> 
> If we want to provide helpers for consumers of the SDK, we could provide some examples of options combinations for e.g. pretty-printing logs, outputting them to files, or exporting them to an external log sink — all of these would be references to the existing `Pino` transport and options APIs!

Great feedback. Considering that I readded the logger pkg to be the centralization piece of loggers. I tested this locally with a script I have and turned out to be a great config place, for example to use pino-pretty when testing locally

Screenshot from 2025-03-12 19-20-00

Ansonhkg:
Merging this to avoid further divergence from the base PRs. Let’s create follow-up PRs if anything needs to be addressed.

PR #812, opened 2025-03-28
add module support

glitch003:
Supports the boolean "module" true or false so we can use module imports, supported here: https://github.com/LIT-Protocol/lit-assets/pull/1749

### DIFF
--- a/packages/types/src/lib/interfaces.ts
+++ b/packages/types/src/lib/interfaces.ts
@@ -482,6 +482,11 @@ export interface JsonExecutionSdkParams
    * auth methods to resolve
    */
   authMethods?: AuthMethod[];
+
+  /**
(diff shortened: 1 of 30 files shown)


packages/types/src/lib/interfaces.ts
```diff
@@ -482,6 +482,11 @@ export interface JsonExecutionSdkParams
    * auth methods to resolve
    */
   authMethods?: AuthMethod[];
+
+  /**
+   * Whether or not the the action is a js "module" with import statements
+   */
+  module?: boolean;
```
Ansonhkg, line 489: I think the boolean type is less informative and explicit, maybe we can use a string literal type instead? eg. 
```ts
module?: 'esm' | 'cjs'
```

PR #827, opened 2025-05-27
feat: add EIP1559 tx type support

Ansonhkg:
# WHAT

add EIP 1559 support

- If gasPrice, maxFeePerGas, and maxPriorityFeePerGas are all undefined -> fetch fee data
- If EIP-1559 supported, then populate maxFeePerGas, maxPriorityFeePerGas, and set type = 2
- if not supported, then fall back to getPrice
- always use manual values if set


glitch003:
works great in my local tests with walletconnect + PKPEthers

PR #837, opened 2025-07-15
Feature/node 4589 naga add missing maxprice for signsessionkey endpoint

Ansonhkg:
# WHAT

This PR fixes two issues:

1. signSessionKey now requires a `maxPrice` field, now the request is including it.
2. `zod-validator-error` pkg is dependent on a specific version of `zod` package - Prefixed all `"zod-validation-error"` packages to `3.4.0` and removed all `bun.lock` that might have cached different versions.


Ansonhkg:
Closing this in favour of this https://github.com/LIT-Protocol/js-sdk/pull/838

PR #845, opened 2025-07-28
feat(artillery): add init-wallets script and update package.json for …

Ansonhkg:
# WHAT

pregen and fund a set number of wallets, then store them into the master private key pool file so that the artillery test could pick an available master key from the pool to run a test to prevent nonce issue.

### DIFF
--- a/e2e/artillery/init-wallets.ts
+++ b/e2e/artillery/init-wallets.ts
@@ -1,0 +1,65 @@
+import { getBalance } from "viem/actions";
+import * as MasterAccountManager from "../src/helper/MasterAccountManager";
+import { createPublicClient, formatEther, http } from "viem";
+import { nagaDev } from "@lit-protocol/networks";
+import { generatePrivateKey, privateKeyToAccount } from "viem/accounts";
+import { fundAccount } from "../src/helper/fundAccount";
+
+const FUNDING_AMOUNT = 0.1;
+const TOPUP_IF_LESS_THAN = 0.05;
+const NUMBER_OF_WALLETS = 500;
+
+(async () => {
+  console.log("🛜  Network:", process.env['NETWORK']);
+  console.log("🔗 RPC URL:", nagaDev.getRpcUrl());
+  const masterAccount = MasterAccountManager.getMasterAccount();
+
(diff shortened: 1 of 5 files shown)


e2e/artillery/init-wallets.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  // --- Fund Wallets ---
+  const masterAccountsFromPool = MasterAccountManager.getMasterAccountsFromPool();
+  console.log("🔑 Total master accounts in pool:", masterAccountsFromPool.length);
+
+  // Check and top up wallets from pool
+  await MasterAccountManager.checkAndTopupWalletsFromPool({
+    publicClient,
+    masterAccount,
```
hwrdtm, line 42: so, we're using the first available master account from L15 to fund all other master accounts from the file that need topping up?

e2e/src/helper/MasterAccountManager.ts
```diff
@@ -0,0 +1,405 @@
...
+    masterAccountPool = masterAccountPool.filter(account => account.address && account.balance && account.privateKey);
+
+    // write the updated masterAccountPool to the file
+    fs.writeFileSync('master-account-pool.json', JSON.stringify(masterAccountPool, null, 2));
+
+    return masterAccountPool;
+  }
+}
+
+export const checkAndTopupWalletsFromPool = async ({
```
hwrdtm, line 107: do we have a check here that the master account we are using to disperse funds has enough funds to cover all the other master accounts that need to be topped up.

e2e/src/helper/MasterAccountManager.ts
```diff
@@ -0,0 +1,405 @@
...
+        console.log("🔑 Updated balance after funding:", address, balance);
+      }
+    });
+  }
+
+  // Write updated pool back to file
+  fs.writeFileSync('master-account-pool.json', JSON.stringify(masterAccountPool, null, 2));
+};
+
+export const generateAndFundWalletsFromPool = async ({
```
hwrdtm, line 264: similar to above, are we missing a check here that the master account has enough funds to fund the newly generated master account wallets? or were you thinking of doing it optimistically / as much as possible.

PR #853, opened 2025-08-08
feat(networks): add fresh price feed fetcher to network modules

Ansonhkg:
# WHAT

- Added `getFreshPriceFeedInfo()` to vNaga modules (`naga-staging`, `naga-dev`, `naga-test`, `naga-local`) to fetch the latest sorted node prices directly from the PriceFeed contract (RPC), without reinitialising state/handshake.
- Exposed the above via `litClient.getContext().getFreshPriceFeedInfo()` in `createLitClient.ts`.
- No behavioural change to existing calls: SDK still uses cached `latestConnectionInfo` by default; cheapest‑N selection remains unchanged. This new method enables callers (e.g. Artillery) to fetch fresh prices just before requests when needed.


MaximusHaximus:
Not sure what happened here, it looks like this was 1 commit to add some code, then a follow-up commit that _reverted_ the first commit -- but then what was merged appears to actually still have the changed code in it 😕

PR #875, opened 2025-08-28
fix(env): add 'naga-staging' to NETWORK enum and update network modul…

Ansonhkg:
# WHAT

Add `naga-test` and `naga-staging` networks to auth services.

### DIFF
--- a/packages/auth-services/src/_setup/initSystemContext.ts
+++ b/packages/auth-services/src/_setup/initSystemContext.ts
@@ -22,11 +22,14 @@ export async function initSystemContext({ appName }: { appName: string }) {
 
   // TODO: Add more supports for other networks
   if (env.NETWORK === 'naga-dev') {
-    networkModule = await import('@lit-protocol/networks');
+    const { nagaDev } = await import('@lit-protocol/networks');
+    networkModule = nagaDev;
   } else if (env.NETWORK === 'naga-test') {
-    networkModule = await import('@lit-protocol/networks');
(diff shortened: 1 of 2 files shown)


packages/auth-services/src/_setup/initSystemContext.ts
```diff
@@ -21,13 +21,16 @@ export async function initSystemContext({ appName }: { appName: string }) {
...
-  if (env.NETWORK !== 'naga-dev') {
+  if (env.NETWORK === 'naga-dev') {
+    networkModule = await import('@lit-protocol/networks');
+  } else if (env.NETWORK === 'naga-test') {
+    networkModule = await import('@lit-protocol/networks');
+  } else if (env.NETWORK === 'naga-staging') {
+    networkModule = await import('@lit-protocol/networks');
+  } else {
     throw new Error(`Unsupported network: ${env.NETWORK}`);
   }
```
Ansonhkg, line 32:
```suggestion
  if (env.NETWORK === 'naga-dev') {
    const { nagaDev } = await import('@lit-protocol/networks');
    networkModule = nagaDev;
  } else if (env.NETWORK === 'naga-test') {
    const { nagaTest } = await import('@lit-protocol/networks');
    networkModule = nagaTest;
  } else if (env.NETWORK === 'naga-staging') {
    const { nagaStaging } = await import('@lit-protocol/networks');
    networkModule = nagaStaging;
  } else {
    throw new Error(`Unsupported network: ${env.NETWORK}`);
  }
```

PR #881, opened 2025-09-05
[Bug] Fix False Positive Custom Auth Tests

Ansonhkg:
# WHAT

Custom auth E2E tests were producing false-positive results. It was working because it was using the cached Alice's delegation auth sig instead of using a fresh PKP.

# CHANGES

- Fixed the custom auth test to use Eve's PKP instead of Alice.
- Updated the custom auth validation ipfs cid [QmTdTemgWBYS76ACdZPttsve6edukyXjCdsNYNK1QDeXKY](https://explorer.litprotocol.com/ipfs/QmTdTemgWBYS76ACdZPttsve6edukyXjCdsNYNK1QDeXKY)
- Disabled some custom auth tests for the time being, will need to have follow up PRs for that.

# TEST

This PR depends on the hardcoded keysets fixes [here](https://github.com/LIT-Protocol/js-sdk/pull/880)

```bash
bun test e2e/src/tickets/custom-auth.spec.ts --timeout 500000
```

### DIFF
--- a/e2e/src/e2e.spec.ts
+++ b/e2e/src/e2e.spec.ts
@@ -25,7 +25,7 @@ describe('all', () => {
 
   // Auth contexts for testing
   let alicePkpAuthContext: any;
-  let aliceCustomAuthContext: any;
+  let eveCustomAuthContext: any;
 
   beforeAll(async () => {
     try {
@@ -33,7 +33,7 @@ describe('all', () => {
 
       // Create PKP and custom auth contexts using helper functions
       // alicePkpAuthContext = await createPkpAuthContext(ctx);
-      aliceCustomAuthContext = await createCustomAuthContext(ctx);
+      eveCustomAuthContext = await createCustomAuthContext(ctx);
     } catch (e) {
       console.error(e);
       process.exit(1);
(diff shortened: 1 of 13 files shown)


packages/networks/src/networks/vNaga/shared/factories/BaseModuleFactory.ts
```diff
@@ -871,6 +872,9 @@ export function createBaseModule<T, M>(config: BaseModuleConfig<T, M>) {
               litActionCode: requestBody.litActionCode,
               litActionIpfsId: requestBody.litActionIpfsId,
               jsParams: requestBody.jsParams,
+              maxPrice: getUserMaxPrice({
```
Ansonhkg, line 875: This was missing in the payload

e2e/src/helper/tests/execute-js.ts
```diff
@@ -3,7 +3,8 @@ import { assert } from '../assertions';
 
 export const createExecuteJsTest = (
   ctx: Awaited<ReturnType<typeof init>>,
-  getAuthContext: () => any
+  getAuthContext: () => any,
+  pubkey?: string
```
Ansonhkg, line 7: optional param for custom auth to pass in pub key instead of using alice's

e2e/src/e2e.spec.ts
```diff
@@ -124,34 +124,51 @@ describe('all', () => {
...
-        });
+        // Disable for now because it requires a different flow
+        // it('pkpPermissionsManagerFlow', () =>
+        //   createPkpPermissionsManagerFlowTest(
+        //     ctx,
+        //     () => eveCustomAuthContext, ctx.eveViemAccountPkp.pubkey
+        //   )());
       });
+
+      // describe('integrations', () => {
```
glitch003, line 162: we don't need these tests anymore?  or plan is to fix and re-enable after this PR?

e2e/src/tickets/custom-auth.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    // Step 3: Mint PKPs for Users - Mint PKPs for your users using the custom auth method
+    // type and validation CID.
+    // Each user gets their own unique PKP tied to your dApp's authentication system.
+    const litClient = await createLitClient({ network: nagaDev });
+    const mockDatabase = new MockDatabase();
+
+    for (const userId of [
+      'alice',
+      // , 'bob'
+    ]) {
```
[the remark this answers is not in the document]
glitch003: but you only loop over 1 item?  is the plan to enable "bob" again in a future PR?

e2e/src/e2e.spec.ts
Ansonhkg: we still need them, but plan is to fix and re-enable after this PR.

PR #882, opened 2025-09-05
[Bug] fix(webauthn): to include scopes in the API

Ansonhkg:
# WHAT 

- Fixed Webauthn COSE issue due to missing pubkey
- Fixed PKP doesn't have scope issue due to missing API param

image

### DIFF
--- a/packages/auth-services/src/auth-server/src/schemas/AuthServiceMintRequestSchema.ts
+++ b/packages/auth-services/src/auth-server/src/schemas/AuthServiceMintRequestSchema.ts
@@ -28,6 +28,7 @@ export type AuthServiceMintRequestTransformed = z.infer<
 export const tAuthServiceMintRequestSchema = t.Object({
   authMethodType: t.String(),
   authMethodId: t.String(),
+  pubkey: t.Optional(t.String()),
   pubkey: t.Optional(t.String({ default: '0x' })),
   scopes: t.Optional(
     t.Array(
--- a/packages/auth-services/src/queue-manager/src/handlers/pkpMint/pkpMint.handler.ts
+++ b/packages/auth-services/src/queue-manager/src/handlers/pkpMint/pkpMint.handler.ts
@@ -14,7 +14,18 @@ export async function handlePkpMintTask(jobData: {
     pubkey: Hex;
     scopes?: ('sign-anything' | 'personal-sign' | 'no-permissions')[];
   };
+  reqId?: string;
 }): Promise<any> {
+  if (
+    // AUTH_METHOD_TYPE.WebAuthn = 3 (without importing the constants package)
+    Number(jobData.requestBody.authMethodType) === 3 &&
+    (!jobData.requestBody.pubkey || jobData.requestBody.pubkey === '0x')
+  ) {
+    throw new Error(
+      `[PKP Mint][HANDLER] WebAuthn requires a non-empty COSE pubkey; got '${jobData.requestBody.pubkey}'. reqId=${jobData.reqId}`
+    );
+  }
+
   const userAuthData: Optional<AuthData, 'accessToken'> = {
     authMethodId: jobData.requestBody.authMethodId,
     authMethodType: Number(jobData.requestBody.authMethodType),
--- a/packages/auth/src/lib/authenticators/native/WebAuthnAuthenticator.ts
+++ b/packages/auth/src/lib/authenticators/native/WebAuthnAuthenticator.ts
@@ -149,6 +149,7 @@ export class WebAuthnAuthenticator {
   public static async registerAndMintPKP(params: {
     username?: string;
     authServiceBaseUrl: string;
+    scopes?: ('sign-anything' | 'personal-sign' | 'no-permissions')[];
   }): Promise<{
     pkpInfo: PKPData;
 
@@ -183,6 +184,7 @@ export class WebAuthnAuthenticator {
       authMethodType: AUTH_METHOD_TYPE.WebAuthn,
       authMethodId: authMethodId,
       pubkey: authMethodPubkey,
+      scopes: params.scopes,
     };
 
     // Immediate mint a new PKP to associate with the auth method
--- a/packages/networks/src/networks/vNaga/shared/managers/LitChainClient/apis/highLevelApis/mintPKP/MintPKPSchema.ts
+++ b/packages/networks/src/networks/vNaga/shared/managers/LitChainClient/apis/highLevelApis/mintPKP/MintPKPSchema.ts
@@ -17,8 +17,10 @@ export const MintPKPSchema = z
 
     // Determine pubkey based on the (potentially derived) authMethodType
     if (data.authMethodType === AUTH_METHOD_TYPE.WebAuthn) {
-      if (!data.pubkey) {
-        throw new Error('pubkey is required for WebAuthn');
+      if (!data.pubkey || data.pubkey === '0x') {
+        throw new Error(
+          `pubkey is required for WebAuthn and cannot be 0x. Received pubkey: "${data.pubkey}" and authMethodType: ${data.authMethodType}`
+        );
       }
       derivedPubkey = data.pubkey as Hex;
     } else {
--- a/packages/lit-client/src/lib/LitClient/createLitClient.ts
+++ b/packages/lit-client/src/lib/LitClient/createLitClient.ts
@@ -350,7 +350,8 @@ export const _createNagaLitClient = async (
     return await networkModule.api.signCustomSessionKey.handleResponse(
       result as any,
       params.requestBody.pkpPublicKey,
-      jitContext
+      jitContext,
+      requestId
     );
   }
 
--- a/packages/networks/src/networks/vNaga/shared/factories/BaseModuleFactory.ts
+++ b/packages/networks/src/networks/vNaga/shared/factories/BaseModuleFactory.ts
@@ -898,7 +898,8 @@ export function createBaseModule<T, M>(config: BaseModuleConfig<T, M>) {
         handleResponse: async (
           result: z.infer<typeof GenericEncryptedPayloadSchema>,
           pkpPublicKey: Hex | string,
-          jitContext: NagaJitContext
+          jitContext: NagaJitContext,
+          requestId?: string
         ) => {
           if (!result.success) {
             E2EERequestManager.handleEncryptedError(


packages/auth-services/src/queue-manager/src/handlers/pkpMint/pkpMint.handler.ts
```diff
@@ -14,7 +14,18 @@ export async function handlePkpMintTask(jobData: {
     pubkey: Hex;
     scopes?: ('sign-anything' | 'personal-sign' | 'no-permissions')[];
   };
+  reqId?: string;
 }): Promise<any> {
+  if (
+    // AUTH_METHOD_TYPE.WebAuthn = 3 (without importing the constants package)
+    Number(jobData.requestBody.authMethodType) === 3 &&
```
Ansonhkg, line 21: Using hardcoded 3 cus don't want to include a whole `constants` package just for this

packages/auth-services/src/auth-server/src/schemas/AuthServiceMintRequestSchema.ts
```diff
@@ -28,7 +28,7 @@ export type AuthServiceMintRequestTransformed = z.infer<
 export const tAuthServiceMintRequestSchema = t.Object({
```
MaximusHaximus, line 28: This seems to be a duplication of `AuthServiceMintRequestSchema` - why not just use our ZOD schema which is defined right above this?  That schema slso still has the .default('0x') applied to it 🤔

Does 'Elysia' not let us just use our ZOD schemas to validate our endpoints for some reason? Why use it instead of Express?

Actually, I checked out the codebase and it appears that `tAuthServiceMintRequestSchema` isn't referenced anywhere...

On further inspection, it looks like the ZOD schema is only used as a type hint `AuthServiceMintRequestRaw` on the handler, and the entire `tAuthServiceMintRequestSchema` is actually duplicated in `pkpMintDoc.ts` (https://github.com/LIT-Protocol/js-sdk/blob/eff05d90543a411f0e62e1327162fc98d143fc56/packages/auth-services/src/auth-server/src/routes/pkp/mint.ts#L21) so we have 3 copies of this schema and the one that appears to be used at runtime is still configured to have a `0x` default string on it

packages/networks/src/networks/vNaga/shared/managers/LitChainClient/apis/highLevelApis/mintPKP/MintPKPSchema.ts
```diff
@@ -17,8 +17,10 @@ export const MintPKPSchema = z
 
     // Determine pubkey based on the (potentially derived) authMethodType
     if (data.authMethodType === AUTH_METHOD_TYPE.WebAuthn) {
-      if (!data.pubkey) {
-        throw new Error('pubkey is required for WebAuthn');
+      if (!data.pubkey || data.pubkey === '0x') {
```
MaximusHaximus, line 20: `MintPKPSchema` here seems to be a fourth copy of `AuthServiceMintRequestSchema` -- shouldn't all of 4 these copies actually be a single schema that is exported from our either the schemas package or the lit client package so that callers can just use it directly and be sure that the schema is in sync with what the `mintPKP` method will actually apply when we call it?

Actually, it looks like the `MintPKPSchema` is incorrect, too -- it's showing `authMethodType` as being a union of bigint or number, but it should be constrained by the values of `AUTH_METHOD_TYPE` -- and we're doing a hard test here using `===` on `AUTH_METHOD_TYPE.WebAuthn`, but if someone passes in a bigint version it won't match, but will pass schema validation...

MaximusHaximus (changes requested):
Added a few issues in-line

packages/auth/src/lib/authenticators/native/WebAuthnAuthenticator.ts
```diff
@@ -149,6 +149,7 @@ export class WebAuthnAuthenticator {
   public static async registerAndMintPKP(params: {
     username?: string;
     authServiceBaseUrl: string;
+    scopes?: ('sign-anything' | 'personal-sign' | 'no-permissions')[];
```
MaximusHaximus, line 152: Should be using the constant definitions / types from  `ScopeSchema`?

packages/auth-services/src/auth-server/src/schemas/AuthServiceMintRequestSchema.ts
Ansonhkg: `Elysia` integrates well with swagger doc based on t schema. 

Follow up npm migration PR will migrate Elysia to Express completely.

packages/auth-services/src/queue-manager/src/handlers/pkpMint/pkpMint.handler.ts
Ansonhkg: Yes, you're right we can just let the Zod schema to handle this - throw this in to specifically catch the Webauthn error.

packages/auth-services/src/auth-server/src/schemas/AuthServiceMintRequestSchema.ts
MaximusHaximus: FYI, we have had a [good experience so far](https://github.com/LIT-Protocol/Vincent/blob/main/packages/libs/registry-sdk/src/lib/openApi/app.ts) with using `zod-to-openapi` to define the endpoints for the registry, which also allows us to codegen an API client for the registry-sdk, though considering how simple the auth-server is that may be overkill here.

We are using `https://www.npmjs.com/package/express-openapi-validator` to provide basic input validation on the endpoints

This is what provides [this output / playground on the deployed service](https://staging.registry.heyvincent.ai/openapi#get-/apps)

packages/schemas/src/lib/auth/auth-schemas.ts
```diff
@@ -1,19 +1,19 @@
...
+  AUTH_METHOD_TYPE_VALUES,
+} from '@lit-protocol/constants';
 import { z } from 'zod';
 import {
   AuthMethodSchema,
   HexPrefixedSchema,
   NodeSetSchema,
   SessionKeyUriSchema,
 } from '../schemas';
+import { ScopeStringSchema } from './ScopeSchema';
```
Ansonhkg, line 12:
```suggestion
import { ScopeSchemaRaw } from './ScopeSchema';
```
properly transform scope strings like "sign-anything" to their BigInt values (like 1n).

packages/schemas/src/lib/auth/auth-schemas.ts
```diff
@@ -64,3 +65,34 @@ export const JsonSignCustomSessionKeyRequestForPkpReturnSchema = z
...
+ * Consolidated schema for PKP mint requests.
+ * This replaces the duplicated schemas across the codebase.
+ * Handles both string and number authMethodType inputs.
+ */
+export const MintPKPRequestSchema = z
+  .object({
+    authMethodId: HexPrefixedSchema,
+    authMethodType: z.coerce.number().pipe(z.nativeEnum(AUTH_METHOD_TYPE)),
+    pubkey: HexPrefixedSchema.default('0x'),
+    scopes: z.array(ScopeStringSchema).optional().default([]),
```
Ansonhkg, line 79:
```suggestion
    scopes: z.array(ScopeSchemaRaw).optional().default([]),
```
properly transform scope strings like "sign-anything" to their BigInt values (like 1n).

packages/schemas/src/lib/auth/auth-schemas.ts
```diff
@@ -1,19 +1,27 @@
...
   HexPrefixedSchema,
   NodeSetSchema,
   SessionKeyUriSchema,
 } from '../schemas';
+import { ScopeSchemaRaw } from './ScopeSchema';
+
+export const CustomAuthDataSchema = z.object({
+  authMethodId: HexPrefixedSchema,
+  // This will be a very big number, unlike our native auth
+  authMethodType: z.bigint(),
```
MaximusHaximus, line 17: Can this value be bigger than Number supports (e.g. > 2^53 ?) If so, we can't represent this data as number (or coerce it to be one like we are trying to do on L24 below)

packages/schemas/src/lib/auth/auth-schemas.ts
```diff
@@ -1,19 +1,27 @@
...
+export type CustomAuthData = z.infer<typeof CustomAuthDataSchema>;
 
 export const AuthDataSchema = z.object({
   authMethodId: HexPrefixedSchema,
-  authMethodType: z.union([
-    AuthMethodSchema.shape.authMethodType,
-    z.number(),
-    z.bigint(),
-  ]),
+  authMethodType: z.coerce.number().pipe(z.nativeEnum(AUTH_METHOD_TYPE)),
```
MaximusHaximus, line 24: See https://github.com/LIT-Protocol/js-sdk/pull/882#discussion_r2352719635 - which was resolved, but without a response. We never use native TS enums in the SDK

packages/schemas/src/lib/auth/auth-schemas.ts
```diff
@@ -64,3 +73,34 @@ export const JsonSignCustomSessionKeyRequestForPkpReturnSchema = z
...
+ * Handles both string and number authMethodType inputs.
+ */
+export const MintPKPRequestSchema = z
+  .object({
+    authMethodId: HexPrefixedSchema,
+    authMethodType: z.coerce.number().pipe(z.nativeEnum(AUTH_METHOD_TYPE)),
+    pubkey: HexPrefixedSchema.default('0x'),
+    scopes: z.array(ScopeSchemaRaw).optional().default([]),
+  })
+  .refine(
```
MaximusHaximus, line 89: 💖 nice, this makes sense

MaximusHaximus (changes requested):
:+1: Looking pretty close -- just one question about the possible input for the custom schema which we are trying to coerce to a number, even though it appears it could be > 2^53 -- and z.nativeEnum usage that might've been missed in my last review.

packages/schemas/src/lib/auth/auth-schemas.ts
Ansonhkg: I'm trying to avoid writing a complex verbose literal-union schema`AuthMethodTypeSchema` because `z.nativeEnum` does accept enum-like object (not only real TS enums).  Zod’s API treats anything shaped like `{ [k: string]: string | number }` as a “native enum,” so the `as const` map works.

Otherwise i'll have to write something like:

```ts
const AuthMethodTypeSchema = z.union(
  (Object.values(AUTH_METHOD_TYPE) as number[]).map(v => z.literal(v)) as [
    z.ZodLiteral<number>,
    ...z.ZodLiteral<number>[]
  ]
);
```

whereas we could just do:

```ts
z.coerce.number().pipe(z.nativeEnum(AUTH_METHOD_TYPE)),
```

image

PR #924, opened 2025-10-04
Artillery Improvements

hwrdtm:
Aggregate specific errors so Artillery backend wouldn't be overwhelmed and give us a 404 when viewing the test report.


hwrdtm:
Closing in favor of #991

